PR #1, opened 2026-03-04
feat: add zellij and tmux workspace managers

rjwittams:
## Summary

- Implement zellij workspace manager with template-based workspace creation, tab management, and state persistence
- Implement tmux workspace manager mapping workspaces to tmux windows with split-window pane support and state persistence  
- Register both in provider discovery with env-var-based detection (ZELLIJ, TMUX)
- Prefer env-var detection over binary-exists checks for workspace manager selection

## Test Plan
- [x] All existing tests pass (7/7)
- [x] Clippy clean
- [ ] Manual test: run flotilla inside tmux, verify workspace listing/creation/switching
- [ ] Manual test: run flotilla inside zellij, verify workspace listing/creation/switching

### DIFF
--- a/src/providers/workspace/tmux.rs
+++ b/src/providers/workspace/tmux.rs
@@ -109,6 +109,8 @@ impl TmuxWorkspaceManager {
     /// Map split direction names to tmux flags.
     /// tmux: -h = horizontal split (pane appears to the right)
     ///        -v = vertical split (pane appears below)
+    /// Note: tmux doesn't support placing a pane to the left or above directly;
+    /// "left" produces the same result as "right" (-h), "up" same as "down" (-v).
     fn split_flag(direction: &str) -> &'static str {
         match direction {
             "left" | "right" => "-h",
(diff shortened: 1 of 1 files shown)


rjwittams:
Fixed in b71686b:

- **#1 tmux select-pane target**: Fixed — now uses `:.N` syntax to correctly target pane within current window.
- **#7 split_flag "left" limitation**: Added doc comment noting the limitation.

Already addressed in previous rounds:
- **#2 Zellij focus**: Fixed in d6bf4d4 (uses focus-previous-pane now). Fragility is a known limitation, same timing approach as all other multiplexer providers.
- **#3 write-chars implicit focus**: Same implicit assumption used in cmux provider. Documenting as known limitation.

Follow-ups filed:
- **#4 Stale state pruning**: #8
- **#5 created_at format**: Nothing reads it yet.
- **#6 Template duplication**: Two call sites, may diverge.
- **#8 Synchronous check_version**: #5
- **#9 cmux binary fallback**: Pre-existing behavior, not introduced in this PR.
- **#10 tmux binary check**: If $TMUX is set, we're inside tmux — the binary is available.

PR #3, opened 2026-03-04
feat: plain git worktree checkout manager

rjwittams:
## Summary

- **Global workspace template fallback**: `~/.config/flotilla/workspace.yaml` is used when no repo-level template exists
- **Plain git worktree CheckoutManager**: New `GitCheckoutManager` using `git worktree` commands directly, as a fallback when `wt` CLI is not installed (or forced via config)
- **Configurable worktree path templates**: Worktrunk-compatible Jinja syntax via minijinja (`{{ repo_path }}/../{{ repo }}.{{ branch | sanitize }}`)
- **Config-driven provider selection**: `[vcs.git.checkouts]` in `~/.config/flotilla/config.toml` with per-repo overrides in `repos/<slug>.toml`
- **Vcs::resolve_repo_root()**: New trait method resolves worktree paths to main repo root, fixing cwd auto-detect adding worktrees as repos

## Test plan

- [x] `cargo check` passes
- [x] `cargo clippy` clean (only pre-existing warnings)
- [x] `cargo test` — all 7 tests pass
- [x] Manual test: set `provider = "git"` in config, verified worktrees listed correctly in UI

### DIFF
--- a/src/providers/vcs/git.rs
+++ b/src/providers/vcs/git.rs
@@ -43,9 +43,9 @@ impl super::Vcs for GitVcs {
         }
         let git_dir = PathBuf::from(String::from_utf8_lossy(&output.stdout).trim());
         // The repo root is the parent of the .git directory.
-        // Note: for bare repos, git-common-dir returns the repo dir itself,
-        // so parent() gives the containing dir — not an issue for our use case
-        // since we only call this from cwd auto-detect.
+        // Note: for bare repos git-common-dir IS the repo dir (e.g. foo.git),
+        // so parent() gives the containing dir, not the repo root. Not an issue
+        // since we only call this from cwd auto-detect on non-bare repos.
         git_dir.parent().map(|p| p.to_path_buf())
     }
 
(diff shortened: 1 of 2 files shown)


rjwittams:
Addressed second review round in e3a2c89:

**1. Config dir inconsistency** — Fixed. The outlier was `executor.rs` using `dirs::config_dir()` — changed to `dirs::home_dir().join(".config/flotilla/...")` to match the rest of the codebase.

**2. default_branch fallback** — Fixed. Now checks for local `main`, `master`, `trunk` branches when `origin/HEAD` is not set.

**3. Bare repo resolve_repo_root** — Already commented in prior round. Follow-up material.

**4. Duplicate parse_working_tree** — Acknowledged, but deferring extraction. Both are ~20 lines in different providers. Not worth the coupling for this PR.

**5. Silent TOML parse errors** — Fixed. Now logs `tracing::warn!` on parse failure.

**6. sanitize filter** — Deliberately matches worktrunk's behaviour for compatibility. No change.

**7. dead_code on path** — Minor, leaving as-is.

rjwittams:
Third review addressed in 17865db. Follow-ups filed:

- #9 — Extract shared git status parsing
- #10 — Handle bare repo in resolve_repo_root  
- #11 — Allow specifying base branch for new worktrees

Other items from review: `render_str` recompilation is acceptable (called rarely, and `add_template_owned` isn't available in minijinja v2). `-D` vs `-d` was deliberately chosen in round 1. `load_config` disk re-reads are fine at current call frequency.

PR #19, opened 2026-03-05
feat: GitHub API ETag caching to prevent rate limiting

rjwittams:
## Summary

- Adds `GhApiClient` that wraps `gh api --include` with ETag-based conditional request caching
- Switches `GitHubIssueTracker` and `GitHubCodeReview` from `gh issue/pr list` to REST API endpoints via `GhApiClient`
- On subsequent refreshes, sends `If-None-Match` headers — GitHub returns 304 Not Modified when data hasn't changed, which **doesn't count against rate limits**
- Guards provider construction on repo slug availability (logs warning and skips if unavailable)

**Rate limit impact:** ~1,080 API calls/hr/repo → ~3 initial calls then free (304s don't count). Verified with live test.

## Changes

| File | What |
|------|------|
| `src/providers/github_api.rs` | New `GhApiClient` with ETag cache + response parser |
| `src/providers/issue_tracker/github.rs` | Uses REST `/repos/{owner}/{repo}/issues`, filters out PRs |
| `src/providers/code_review/github.rs` | Uses REST `/repos/{owner}/{repo}/pulls`, handles merged via `merged_at` |
| `src/providers/discovery.rs` | Creates shared `GhApiClient`, extracts repo slug, guards on availability |
| `src/providers/mod.rs` | Adds `pub mod github_api` |

## Test plan

- [x] 23 unit tests pass (`cargo test`)
- [x] Clippy clean (only benign `status` field warning)
- [x] Live tested: ran against real repo for 25s, confirmed ETag caching works (3 initial API calls, subsequent cycles free)
- [x] Verified 304 responses don't increment rate limit counter

### DIFF
--- a/src/providers/code_review/github.rs
+++ b/src/providers/code_review/github.rs
@@ -2,7 +2,7 @@ use std::path::Path;
 use std::sync::Arc;
 use async_trait::async_trait;
 use crate::providers::types::*;
-use crate::providers::github_api::GhApiClient;
+use crate::providers::github_api::{GhApiClient, clamp_per_page};
 
 pub struct GitHubCodeReview {
     provider_name: String,
@@ -114,9 +114,10 @@ impl super::CodeReview for GitHubCodeReview {
         repo_root: &Path,
         limit: usize,
     ) -> Result<Vec<ChangeRequest>, String> {
+        let per_page = clamp_per_page(limit);
         let endpoint = format!(
             "repos/{}/pulls?state=open&per_page={}",
-            self.repo_slug, limit
+            self.repo_slug, per_page
         );
         let body = self.api.get(&endpoint, repo_root).await?;
         let items: Vec<serde_json::Value> =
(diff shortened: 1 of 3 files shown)


rjwittams:
Addressed items 1-5 in c2a69fa:

1. **ETag header case**: GitHub actually sends `Etag:` (not `ETag:`) via `gh api`, but made parsing fully case-insensitive anyway — more robust. Added a test covering all casings.

2. **304 detection**: Now parses the HTTP status line from stdout (always present with `--include`) instead of matching stderr strings. The `status` field on `GhApiResponse` is now used directly.

3. **Redundant git calls**: Refactored discovery to call `first_remote_url` once, passing the result to both host detection and slug extraction.

4. **`status` field**: Now used in the 304 detection path (consequence of fix #2).

5. **Mutex poisoning**: All `.lock().unwrap()` calls replaced with `.unwrap_or_else(|p| p.into_inner())`.

6. **Round-trip test**: Skipping — would require mocking the `gh` CLI subprocess, which needs test infrastructure that doesn't exist yet. The case-insensitive ETag test does catch the header parsing concern raised.

PR #136, opened 2026-03-08
feat: client-side materialization and delta replay (PRs 4-5)

rjwittams:
## Summary

- **SocketDaemon local state**: maintains per-repo snapshots locally, applies incoming deltas, detects seq gaps and recovers by requesting full snapshots from server
- **`replay_since` RPC**: enables delta log replay on reconnect — server returns just the missing deltas instead of full snapshots when possible
- **`apply_snapshot_delta`**: core utility for mutating snapshots in place from delta changes

Completes Tasks 16-19 of the delta snapshots implementation plan. Tasks 18-19 (TUI delta handling) were already implemented in PR #129.

## Test plan

- [ ] `cargo test --workspace` — 356 tests pass
- [ ] `cargo clippy --all-targets --locked -- -D warnings` — clean
- [ ] 3 new in-process tests for `replay_since` (unknown seq, new repo, up-to-date)
- [ ] 2 new socket roundtrip assertions (up-to-date returns empty, bogus seq returns full snapshot)
- [ ] Verify seq gap recovery: SocketDaemon detects `prev_seq` mismatch, requests full snapshot, emits `SnapshotFull` to TUI

### DIFF
--- a/crates/flotilla-tui/src/socket.rs
+++ b/crates/flotilla-tui/src/socket.rs
@@ -1,4 +1,4 @@
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::path::{Path, PathBuf};
 use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::Arc;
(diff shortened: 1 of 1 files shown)


rjwittams:
**Re: codex findings**

**Finding 1 (event ordering):** Addressed. With the switch to `std::sync::RwLock`, SnapshotFull and happy-path SnapshotDelta are now handled synchronously in the reader loop — seq is updated and event dispatched atomically, no spawn. Only gap recovery (where we need async RPC) is spawned, and recovery replaces state entirely, so ordering relative to the triggering event is irrelevant.

**Finding 2 (startup replay not seeding local_seqs):** Addressed. `SocketDaemon::replay_since()` now seeds `local_seqs` from the returned events before returning them to the caller. The first live delta will find the correct seq in `local_seqs`.

PR #175, opened 2026-03-09
feat: persistent portable terminal sessions (#32)

### LINKED ISSUE #32: Persistent and portable terminal sessions
## Summary

Make workspaces resumable and portable across machines and workspace managers. Flotilla (or something it controls) owns the terminal processes on each host, enabling reconstitution elsewhere.

## Concept

Currently a workspace is tied to a specific terminal multiplexer session on a specific machine. If that session ends, the workspace is gone. The goal is to decouple the logical workspace from its physical terminal host.

### Approach

- Interpose a session persistence layer (e.g. zmx or similar) so terminal state survives disconnection
- Flotilla tracks which panes belong to which workspace
- A workspace can be "rehomed" — its panes reconstituted on a different machine or workspace manager
- Each pane's process and scrollback are migrated or restarted as appropriate

### References

- cmux PR: https://github.com/manaflow-ai/cmux/pull/510
- zmx: https://github.com/neurosnap/zmx
- shpool: https://github.com/shell-pool/shpool <- rust, has a lib 

## Open questions

- How much terminal state needs to be portable vs. recreatable (scrollback, env vars, running processes)?
- How to handle users creating panes outside flotilla's control (new tab, manual split)?
- What's the right persistence layer — zmx, custom, or delegate to workspace managers that support it?

## Dependencies

Foundation for multi-host coordination and agent session migration.

rjwittams:
## Summary

Adds terminal session persistence via shpool integration, enabling terminal sessions that survive tab/window closures and can be reattached.

- **TerminalPool provider trait** with `ShpoolTerminalPool` (shpool integration) and `PassthroughTerminalPool` (fallback) implementations
- **ManagedTerminal protocol types** for tracking terminal session state (running/disconnected)
- **Template format unification**: removed V1/V2 split, unified on content+layout format (~300 lines removed)
- **Full integration**: provider discovery, refresh cycle, correlation with checkouts, workspace manager routing, UI health tracking
- **Robust shpool command handling**: wraps commands in interactive login shell (`$SHELL -lic`) for full user environment (PATH, aliases, node, direnv)

Closes #32

## Test plan
- [x] All 432 tests pass (`cargo test`)
- [x] Clippy clean (`cargo clippy --all-targets --locked -- -D warnings`)
- [x] Formatting clean (`cargo fmt --check`)
- [x] End-to-end verified: create workspace → kill zellij tab → sessions persist as Disconnected → reattach brings back same processes
- [ ] Review shpool integration on clean machine

### DIFF
--- a/crates/flotilla-core/src/providers/terminal/shpool.rs
+++ b/crates/flotilla-core/src/providers/terminal/shpool.rs
@@ -40,12 +40,15 @@ impl ShpoolTerminalPool {
                 continue;
             };
 
-            // Parse "checkout/role/index"
-            let parts: Vec<&str> = rest.splitn(3, '/').collect();
-            if parts.len() != 3 {
+            // Parse "checkout/role/index" from the right — checkout may contain
+            // slashes (e.g. "feature/foo"), but role and index never do.
+            let Some((before_index, index_str)) = rest.rsplit_once('/') else {
                 continue;
-            }
-            let index: u32 = parts[2].parse().unwrap_or(0);
+            };
+            let Some((checkout, role)) = before_index.rsplit_once('/') else {
+                continue;
+            };
+            let index: u32 = index_str.parse().unwrap_or(0);
 
             let status_str = session["status"]
                 .as_str()
@@ -59,11 +62,11 @@ impl ShpoolTerminalPool {
 
             terminals.push(ManagedTerminal {
                 id: ManagedTerminalId {
-                    checkout: parts[0].into(),
-                    role: parts[1].into(),
+                    checkout: checkout.into(),
+                    role: role.into(),
                     index,
                 },
-                role: parts[1].into(),
+                role: role.into(),
                 command: String::new(), // shpool doesn't report the original command
                 working_directory: PathBuf::new(), // populated separately if needed
                 status,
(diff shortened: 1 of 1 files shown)


rjwittams:
Thanks for the thorough review! Addressing each finding:

**BUG: Double-escaping in attach_command** — Pushback: the quoting is correct. There are intentionally two layers serving two different parsers:

1. **Inner escape** (`escaped_cmd`): quotes the raw command for embedding inside `-lic '...'`, consumed by `shell-words` inside shpool
2. **Outer `sq()`**: quotes the entire `--cmd` value for the user's outer shell (the attach command is typed into a terminal by the workspace manager)

Traced example for command `echo 'hello'`:
- `escaped_cmd` = `echo '\''hello'\''`
- format = `/bin/zsh -lic 'echo '\''hello'\'''`
- `sq()` wraps for outer shell → user's shell strips it → shpool receives `/bin/zsh -lic 'echo '\''hello'\'''`
- shell-words parses → `["/bin/zsh", "-lic", "echo 'hello'"]` ✓

Each `'` replacement serves a different parser boundary. This is standard POSIX nested shell quoting.

**DESIGN GAP: overflow/gap fields** — Correct, these are parsed but not yet implemented. They're scaffolding for planned multi-instance overflow behavior. Will add a comment noting they're deferred.

**MINOR: Duplicate role** — Agreed, `ManagedTerminal.role` duplicates `id.role`. Minor cleanup, will address in a follow-up.

**NIT: /bin/sh -lic** — Fair edge case. In practice `$SHELL` is always set (it's set by login and inherited), so `/bin/sh` is a fallback that should never fire. Worth a comment though.

PR #192, opened 2026-03-10
feat: Checkouts section title + Path column (#185, #184)

### LINKED ISSUE #184: feat: add Path column to work item table
## Problem

The work item table currently has no Path column. When using worktrees, the checkout path is only visible in the preview panel on the right. For quick scanning across many work items, having the path inline is valuable.

## Proposal

Add a **Path** column between the icon and Description columns. The path should be shortened intelligently based on worktree naming conventions:

- Strip common prefixes (repo root, `.worktrees/` directory)
- Show only the distinguishing suffix — e.g. for `/Users/robert/dev/flotilla/.worktrees/feat-auth`, show `feat-auth`
- For the main worktree, show `.` or the repo name
- Truncate to column width with ellipsis if needed

The column width should be modest (e.g. `Length(15-20)`) since branch names often duplicate the useful part of the path.

## Current columns

`icon | Description | Branch | WT | WS | PR/CR | Sessions | Issues | Git`

## Proposed columns

`icon | Path | Description | Branch | WT | WS | PR/CR | Sessions | Git`

## Files involved

- `crates/flotilla-tui/src/ui.rs` — table header, `build_item_row()`, column widths
- Path data already available via `item.checkout_key()`

### LINKED ISSUE #185: fix: use "Checkouts" as section title instead of provider-derived label
## Problem

The section title for checkouts is derived from the first checkout manager provider via `section_label()`. Both `WtCheckoutManager` and `GitCheckoutManager` return "Worktrees", but this is a provider implementation detail leaking into the UI.

The concept in flotilla is **checkouts** — a checkout could be a worktree, a plain git branch checkout, or potentially other VCS mechanisms. The section title should reflect the domain concept, not the provider.

## Proposal

Always use "Checkouts" as the section title for this category, regardless of which checkout manager provider is active. This is already the default in `SectionLabels::default()` and the `CheckoutManager` trait default for `section_label()`.

Options:
1. **Hardcode "Checkouts" in `SectionLabels`** — simplest, skip provider derivation for this field
2. **Remove the `section_label()` override** in `wt.rs` and `git_worktree.rs` — let them use the trait default "Checkouts"

Option 2 is cleaner since it keeps the provider-label system intact for other categories (e.g. "Pull Requests" vs "Change Requests" is a useful provider distinction).

## Files involved

- `crates/flotilla-core/src/providers/vcs/wt.rs:120` — overrides to "Worktrees"
- `crates/flotilla-core/src/providers/vcs/git_worktree.rs:237` — overrides to "Worktrees"
- Tests referencing "Worktrees" as section label

rjwittams:
## Summary

- **#185**: Remove `section_label()` overrides in `WtCheckoutManager` and `GitCheckoutManager` that returned "Worktrees", letting the `CheckoutManager` trait default "Checkouts" apply. Updates test stubs and assertions.
- **#184**: Add a Path column between icon and Description in the work item table. Paths are shortened relative to the repo root (strips `.worktrees/` prefix, shows `.` for main checkout). Includes `shorten_path()` helper with unit tests.

## Test plan

- [x] All 592 workspace tests pass
- [x] Snapshot tests updated for new column layout and "Checkouts" header
- [x] `cargo clippy` and `cargo fmt` clean
- [x] Unit tests for `shorten_path` cover: main checkout, worktree path, relative path, path outside root

Closes #185
Closes #184

### DIFF
--- a/crates/flotilla-tui/src/ui_helpers.rs
+++ b/crates/flotilla-tui/src/ui_helpers.rs
@@ -118,7 +118,7 @@ pub fn checkout_indicator(is_main: bool, has_checkout: bool) -> &'static str {
 /// Shorten a checkout path relative to the repo root for display in the table.
 ///
 /// - Main checkout (path == repo_root) → "."
-/// - Under `.worktrees/` → just the directory name (e.g. "feat-auth")
+/// - Under `.worktrees/` → path relative to `.worktrees/` (e.g. "feat-auth" or "group/feat-auth")
 /// - Otherwise → relative path from repo root
 pub fn shorten_path(path: &Path, repo_root: &Path) -> String {
     if path == repo_root {


rjwittams:
Addressed items 1, 3, and 4:

1. **Column width**: Reduced Path from `Length(16)` to `Length(14)`. The "Issues" header still shows as "Iss" at 120-char width — that's the tradeoff at narrow terminals; at 160+ chars it renders fully.
2. **`.worktrees/` prefix**: Agreed, fine for now. No action.
3. **Variable shadowing**: Eliminated the double-bind. Now uses `rel` with `match` + `into_owned()`.
4. **Nested worktree test**: Added test for `.worktrees/group/feat-auth` → `group/feat-auth`.

PR #208, opened 2026-03-10
Fix async pending-request leak in socket client request path

rjwittams:
### Motivation

- Prevent stale entries from accumulating in the `pending` map when `send_request` returns early due to serialization, write/flush, timeout, or cancelled oneshot error paths.

### Description

- Harden `send_request` in `crates/flotilla-client/src/lib.rs` to remove the pending entry on serialization failure, socket write/flush failure, request timeout, or cancelled response. 
- Convert request serialization to a `match` so the entry is removed on `serde_json::to_string` errors. 
- Ensure write errors remove the pending entry before returning, and explicitly remove the pending entry on timeout or cancelled oneshot. 
- Add a regression test `send_request_cleans_pending_on_cancelled_response` to verify the `pending` map is empty after the cancelled-response path.

### Testing

- Ran `cargo fmt --all` and `cargo test -p flotilla-client --locked` and all tests in the `flotilla-client` crate passed (7 passed, 0 failed).

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_69b0857bcd70832f819803161e437cf6)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,7 @@ members = [
 resolver = "2"
 
 [workspace.dependencies]
-tokio = { version = "1", features = ["full"] }
+tokio = { version = "1", features = ["full", "test-util"] }
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 async-trait = "0.1"
--- a/crates/flotilla-client/src/lib.rs
+++ b/crates/flotilla-client/src/lib.rs
@@ -385,7 +385,7 @@ async fn send_request(
         }
     };
 
-    let write_result: Result<(), String> = {
+    let write_result = async {
         let mut w = writer.lock().await;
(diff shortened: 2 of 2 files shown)


rjwittams:
Fixed. The write path now uses an async block so write errors stay in the local Result and the pending entry is removed before returning. Added regression coverage for closed-peer write failure and timeout cleanup, and simplified the redundant remove/drop pattern in the cancelled-response tests. Verified with cargo fmt --check, cargo clippy --all-targets --locked --features flotilla-daemon/skip-no-sandbox-tests -- -D warnings, and TMPDIR="/Users/robert/dev/flotilla.codex-review-for-concurrency-issues/.codex-tmp" cargo test --workspace --locked --features flotilla-daemon/skip-no-sandbox-tests.

PR #248, opened 2026-03-12
fix: discover cmux workspaces across windows

rjwittams:
## Summary
- enumerate cmux windows and aggregate per-window workspaces in the cmux workspace manager
- skip individual window listing/parsing failures with warnings and dedupe duplicate workspace refs
- add focused cmux unit coverage plus the design and implementation plan docs

## Test Plan
- `cargo test --locked -p flotilla-core providers::workspace::cmux::tests`
- `cargo test --locked -p flotilla-tui --test snapshots delete_confirm_with_` (fails in two pre-existing snapshot cases unrelated to this change)
- `mkdir -p .codex-tmp && TMPDIR="$PWD/.codex-tmp" cargo test --workspace --locked --features flotilla-daemon/skip-no-sandbox-tests` (same two pre-existing snapshot failures)

## Notes
The TUI snapshot failures are in `delete_confirm_with_uncommitted_files` and `delete_confirm_with_many_uncommitted_files`, showing a `Path` vs `Source` table-header diff outside the cmux adapter changes.

### DIFF
--- a/crates/flotilla-core/src/providers/workspace/cmux.rs
+++ b/crates/flotilla-core/src/providers/workspace/cmux.rs
@@ -43,10 +43,15 @@ impl CmuxWorkspaceManager {
         let windows = parsed["windows"]
             .as_array()
             .ok_or("cmux list-windows: response missing 'windows' array")?;
-        Ok(windows
-            .iter()
-            .filter_map(|window| window["ref"].as_str().map(ToString::to_string))
-            .collect())
+        let mut refs = Vec::new();
+        for window in windows {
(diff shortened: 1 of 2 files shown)


rjwittams:
Fixed the missing-ref diagnostic in parse_window_refs and added an implemented status note to the plan doc. Left the dedupe loop structure unchanged; the later clarification is correct that the single-lookup form is better for the common non-duplicate path.

rjwittams:
Fixed both minor items: parse_window_refs now warns when cmux returns a window entry without a ref, and the plan doc now includes an implemented status note so it no longer reads like live work.

PR #249, opened 2026-03-12
feat: shpool lifecycle fixes (#237, #243, #244)

### LINKED ISSUE #237: Clean up managed terminals when checkout is removed
## Problem

`kill_terminal` is implemented on `ShpoolTerminalPool` but has zero call sites. When a checkout is removed via `RemoveCheckout` in `executor.rs`, only the checkout manager is called — the associated shpool sessions (`flotilla/<branch>/shell/0` etc.) persist indefinitely.

## Proposed fix

In `executor.rs` `RemoveCheckout` handler, enumerate managed terminals matching the checkout's branch name and call `kill_terminal` for each.

## Context

From PR #234 review. Part of broader terminal lifecycle work (#181).

### LINKED ISSUE #243: Explicitly start shpool daemon with flotilla config
## Problem

Flotilla relies on shpool's auto-daemonization when the first `attach` runs. The `-c` config flag is passed to all shpool commands, but it only affects the daemon if shpool auto-starts one. If a daemon is already running on the socket (from a previous session or manual start), it keeps its old config.

This means `prompt_prefix = ""` (daemon-side setting) doesn't take effect until the daemon is restarted, while `forward_env` (client-side setting) works immediately.

## Proposed fix

Explicitly start the shpool daemon during discovery/startup:

```
shpool --socket <path> -c <config> daemon
```

This ensures the daemon always runs with flotilla's managed config. Need to handle the case where a daemon is already running on the socket (either kill+restart, or check first and skip).

## Context

Discovered while testing PR #234. The TERMINFO forwarding worked (client-side) but prompt prefix persisted (daemon-side) because an existing daemon was already running.

### LINKED ISSUE #244: Work around shpool stale socket detection failure on macOS
## Problem

When the shpool daemon dies, its Unix socket file remains on disk. On macOS, `connect()` to a stale Unix socket **succeeds** (unlike Linux where it returns `ConnectionRefused`). This means shpool's auto-daemonize logic (`libshpool/src/daemonize.rs:36-41`) thinks a daemon is already running and returns `Ok(())`, but the attach then fails with:

```
Error: daemonizing: launched daemon, but control socket never came up
```

shpool has a pid file (`daemonized-shpool.pid`) but doesn't use it for liveness checks. This is ultimately an upstream shpool bug, but flotilla should work around it.

## Proposed workaround

Before the first shpool command, check the pid file. If the pid is dead, remove the stale socket so shpool's auto-daemonize works correctly. This could live in `ShpoolTerminalPool::new()` or in a health check before `list_terminals()`.

## Upstream

Should also be reported to https://github.com/shell-pool/shpool — the fix belongs in `maybe_fork_daemon()`: after a successful `connect()`, verify the daemon is alive via the pid file or a protocol-level ping.

## Context

Discovered while testing PR #234. The `connect()` behavior difference between macOS and Linux is the root cause.

rjwittams:
## Summary

- **#244 (macOS stale socket):** Clean up stale shpool sockets on macOS where `connect()` to dead Unix sockets succeeds. Uses `libc::kill(pid, 0)` to probe daemon liveness before removing orphaned socket/pid files.
- **#243 (explicit daemon start):** Spawn `shpool daemon` with flotilla-managed config during provider discovery instead of relying on shpool's auto-daemonize (which uses default config). Kills existing daemon first to pick up config changes.
- **#237 (terminal cleanup on delete):** Thread `terminal_keys` from correlation engine through `WorkItem` → `DeleteConfirm` UI mode → `RemoveCheckout` command → executor, which calls `kill_terminal()` (best-effort) for each correlated terminal session after successful checkout removal.

## Changes

| Area | What changed |
|------|-------------|
| `flotilla-protocol` | Added `terminal_keys: Vec<ManagedTerminalId>` to `WorkItem` and `Command::RemoveCheckout` with `#[serde(default)]` |
| `flotilla-core/data.rs` | Added `terminal_ids` to `CorrelatedWorkItem`, populated from `ManagedTerminal` correlation items |
| `flotilla-core/convert.rs` | Maps `terminal_ids` → `terminal_keys` in protocol conversion |
| `flotilla-core/shpool.rs` | `clean_stale_socket()` with `#[cfg(unix)]` libc pid check; async `create()` factory with `start_daemon()`; sync `new()` kept for tests |
| `flotilla-core/executor.rs` | `RemoveCheckout` handler calls `kill_terminal()` per terminal_key after successful removal |
| `flotilla-tui` | `DeleteConfirm` carries `terminal_keys` through loading → confirm → command dispatch |

## Test plan

- [x] 3 unit tests for stale socket cleanup (dead pid, orphan socket, no-op)
- [x] 1 test for async `create()` factory method
- [x] 1 test for `terminal_ids` population from correlation groups
- [x] 1 test for executor terminal kill via `MockTerminalPool`
- [x] 1 serde backward-compat test for `WorkItem.terminal_keys`
- [x] All 759 workspace tests pass
- [x] `cargo fmt`, `cargo clippy --all-targets --locked -- -D warnings` clean

Closes #237, closes #243, closes #244

### DIFF
--- a/crates/flotilla-core/src/providers/terminal/shpool.rs
+++ b/crates/flotilla-core/src/providers/terminal/shpool.rs
@@ -54,6 +54,18 @@ impl ShpoolTerminalPool {
         }
     }
 
+    /// Check if a process is alive. Returns true for both "alive and ours"
+    /// (kill returns 0) and "alive but not ours" (EPERM).
+    #[cfg(unix)]
+    fn is_process_alive(pid: i32) -> bool {
+        let rc = unsafe { libc::kill(pid, 0) };
+        if rc == 0 {
+            return true;
+        }
+        // EPERM means the process exists but we can't signal it
+        std::io::Error::last_os_error().raw_os_error() == Some(libc::EPERM)
+    }
+
     /// Remove stale shpool socket and pid files when the daemon is dead.
     ///
     /// On macOS, `connect()` to a stale Unix socket succeeds (unlike Linux
@@ -70,9 +82,7 @@ impl ShpoolTerminalPool {
         match std::fs::read_to_string(&pid_path) {
             Ok(contents) => {
                 if let Some(pid) = contents.trim().parse::<i32>().ok().filter(|&p| p > 0) {
-                    // Signal 0 checks process existence without sending a signal
-                    let alive = unsafe { libc::kill(pid, 0) } == 0;
-                    if alive {
+                    if Self::is_process_alive(pid) {
                         tracing::debug!(%pid, "shpool daemon is alive, keeping socket");
                         return;
                     }


rjwittams:
Thanks for the thorough review!

**PID validation** — Fixed in 22d73c9. Both `clean_stale_socket` and `start_daemon` now use `.ok().filter(|&p| p > 0)` to reject non-positive PIDs.

**Child handle drop** — Added a clarifying comment in the same commit explaining the intentional drop (tokio doesn't kill on drop).

**Always kills daemon** — Filed as #251 for follow-up. Config-hash comparison to skip restart when config hasn't changed is a good idea but out of scope for this PR.

**Sequential kill / test exercises real binary** — Agreed these are fine for now. Noted for future reference.

PR #252, opened 2026-03-12
feat: add dynamic layout controls with RepoViewLayout enum

rjwittams:
## Summary
- Unify `PreviewPositionMode` + `visible: bool` into a single `RepoViewLayout` enum with variants: `Auto`, `Zoom`, `Right`, `Below`
- Single `l` key cycles through all layout modes (replacing `v` for position and `P` for visibility)
- Status bar shows `Layout(l): {variant}` combining the hint and current state
- Config simplified to a single `layout` field (no backwards compat needed)
- Auto mode dynamically picks side-by-side or stacked based on terminal dimensions

## Test plan
- [x] Unit tests for cycle order (Auto → Zoom → Right → Below → Auto)
- [x] Unit tests for layout resolution (auto, explicit, zoom returns None)
- [x] Config round-trip tests (load/save layout)
- [x] Key handler test for `l` binding
- [x] Snapshot tests updated (13 snapshots, 2 orphaned deleted)
- [x] clippy clean, full workspace tests pass (786 tests)

### DIFF
--- a/crates/flotilla-core/src/config.rs
+++ b/crates/flotilla-core/src/config.rs
@@ -55,34 +55,20 @@ pub struct UiConfig {
     pub preview: PreviewConfig,
 }
 
-#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
+#[derive(Debug, Clone, Default, Deserialize, Serialize, PartialEq, Eq)]
 pub struct PreviewConfig {
     #[serde(default)]
-    pub position_mode: PreviewPositionModeConfig,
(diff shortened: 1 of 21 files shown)


rjwittams:
Thanks for the thorough review. Filed follow-ups for the two actionable items:

- #260 — Add test for `(false, false)` fallback in `resolve_auto_preview_position`
- #261 — Deduplicate `TestDaemon` and `StubDaemon`

The other two items (redundant `if let Some` guard, test name precision) are cosmetic and not worth a separate change.

PR #311, opened 2026-03-13
feat: redesign tui status bar

rjwittams:
## Summary
- redesign the normal-mode status bar into status, key-ribbon, and task sections
- add clickable Zellij-style key ribbons, dismissable status items, and an uppercase `K` toggle for the key section
- add status-bar model tests, interaction tests, and updated snapshot coverage

## Test Plan
- [x] cargo test -p flotilla-tui --locked
- [x] cargo clippy --all-targets --locked -- -D warnings
- [x] rustfmt --edition 2021 --check crates/flotilla-tui/src/lib.rs crates/flotilla-tui/src/status_bar.rs crates/flotilla-tui/src/app/ui_state.rs crates/flotilla-tui/src/app/mod.rs crates/flotilla-tui/src/app/key_handlers.rs crates/flotilla-tui/src/ui.rs crates/flotilla-tui/tests/snapshots.rs

### DIFF
--- a/crates/flotilla-tui/src/app/mod.rs
+++ b/crates/flotilla-tui/src/app/mod.rs
@@ -168,6 +168,36 @@ pub struct VisibleStatusItem {
     pub text: String,
 }
 
+fn error_status_item(message: &str) -> VisibleStatusItem {
+    VisibleStatusItem { id: 0, text: format!("ERROR {}", message) }
+}
+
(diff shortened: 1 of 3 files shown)


rjwittams:
Addressed in d0171c7. Fixed the stale status-message dismissal by routing all status_message updates through a shared setter that clears dismissed id 0 when the message changes, extracted a single visible-status helper used by both App::visible_status_items and the renderer, and made StatusBarUiState::default() set show_keys directly. The spinner animation is intentionally deferred; follow-up filed as #314.

PR #312, opened 2026-03-14
feat: shimmer animation for in-progress status text

rjwittams:
## Summary
- Adds a cosine-wave shimmer effect that sweeps across status bar text during in-flight operations (e.g. "Generating branch name...", "Loading safety info...")
- True color terminals get smooth RGB gradient (dim amber → bright yellow), limited terminals fall back to DIM/BOLD yellow modifiers
- Tick rate lowered from 250ms to 50ms with ticks only triggering redraws when animation is active

## Notes
- Will likely need rebasing once the status bar PR lands — the shimmer integrates at the same render points

## Test plan
- [ ] Verify shimmer animates during branch generation (AI-generated branch name)
- [ ] Verify shimmer animates on any in-flight command in the status bar
- [ ] Verify shimmer animates on "Loading safety info..." in delete confirm dialog
- [ ] Verify no extra CPU usage when idle (ticks discarded when no animation needed)
- [ ] Verify fallback styling on non-true-color terminals

### DIFF
--- a/crates/flotilla-tui/src/shimmer.rs
+++ b/crates/flotilla-tui/src/shimmer.rs
@@ -15,7 +15,8 @@ fn elapsed_since_start() -> Duration {
 }
 
 fn has_true_color() -> bool {
-    std::env::var("COLORTERM").map(|v| v == "truecolor" || v == "24bit").unwrap_or(false)
+    static TRUE_COLOR: OnceLock<bool> = OnceLock::new();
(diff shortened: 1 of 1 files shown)


rjwittams:
Fixed both items:
- `has_true_color()` now uses `OnceLock` to cache the env var lookup
- `shimmer_spans` uses `Vec::with_capacity(chars.len())`
- Also removed the redundant `t.clamp(0.0, 1.0)` since the cosine formula already bounds `t` to `[0, 1]`

PR #316, opened 2026-03-14
Add routed control commands for CLI and TUI

rjwittams:
## Summary
- add a routed command envelope and peer-forwarded command lifecycle so control actions can target hosts through the shared protocol
- unify daemon, TUI, and CLI control execution on `execute()` with streamed progress events and structured results for refresh, repo add/remove, and checkout removal
- expose the new control command surface in the root CLI, including `host <name> ...` routing and shared event-driven command handling

## Test Plan
- [x] cargo fmt --check
- [x] git diff --check
- [x] cargo test --workspace --locked


rjwittams:
Tightened the regression test in crates/flotilla-core/tests/in_process_daemon.rs to subscribe to daemon events, wait for CommandFinished for the issued command_id, and assert CommandResult::CheckoutStatus(_). That now verifies successful context_repo resolution rather than only checking that a command ID was allocated.

PR #333, opened 2026-03-14
fix: attach pending action context to RemoveCheckout confirm handler

rjwittams:
## Summary

Fixes the pending action shimmer not appearing when removing a checkout.

- `handle_delete_confirm_key` dispatched `RemoveCheckout` via plain `push()` without pending context, so the row never showed a shimmer indicator after confirming deletion
- `DeleteConfirm` now carries `identity: WorkItemIdentity` (same pattern as `CloseConfirm`), and the confirm handler uses `push_with_context` to attach it
- Also preserves the identity through the `CheckoutStatus` → `DeleteConfirm` transition in `executor.rs`

## Test Plan

- [x] `cargo test --locked` — all pass
- [x] `cargo clippy --all-targets --locked -- -D warnings` — clean
- [ ] Manual: remove a checkout, verify shimmer appears on the row after confirming

### DIFF
--- a/crates/flotilla-tui/src/app/executor.rs
+++ b/crates/flotilla-tui/src/app/executor.rs
@@ -96,7 +96,10 @@ pub fn handle_result(result: CommandResult, app: &mut App) {
         CommandResult::CheckoutStatus(info) => {
             let (terminal_keys, identity) = match &app.ui.mode {
                 UiMode::DeleteConfirm { terminal_keys, identity, .. } => (terminal_keys.clone(), identity.clone()),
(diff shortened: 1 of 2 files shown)


rjwittams:
Both addressed in 71b3de0:

1. **Fallback identity sentinel** — added `tracing::warn!` when `CheckoutStatus` arrives outside `DeleteConfirm` mode.
2. **Missing test** — added `delete_confirm_attaches_pending_context` that verifies `push_with_context` is called with the correct identity when confirming deletion.

PR #342, opened 2026-03-14
Refactor repo state and replay to use RepoIdentity

rjwittams:
## Summary
- re-key protocol replay, client seq tracking, daemon state, and TUI tab state around RepoIdentity
- preserve multiple tracked local roots for a shared identity in InProcessDaemon, with deterministic preferred-path behavior and explicit multi-root regression coverage
- keep routed terminal/workspace follow-up and socket replay handling attached to repo identity instead of host-local paths

## Test Plan
- [x] cargo test --workspace --locked

### DIFF
--- a/crates/flotilla-tui/src/app/mod.rs
+++ b/crates/flotilla-tui/src/app/mod.rs
@@ -383,7 +383,7 @@ impl App {
                 tracing::info!(%command_id, %description, "command started");
                 self.in_flight.insert(command_id, InFlightCommand { repo_identity, repo, description });
             }
-            DaemonEvent::CommandFinished { command_id, host, repo, result, .. } => {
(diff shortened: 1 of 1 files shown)


rjwittams:
Follow-up landed in 111009d and 3ee9a0d. This PR now promotes a newly added local root over a synthetic remote-only root, emits an immediate snapshot when the preferred root changes, documents the merged-after-first-broadcast checkout view and the local-first invariant, documents that last_snapshot is the merged per-identity view, and moves the add_repo path->identity insert into the repos write section. I’m leaving the inject_issues/inject_issues_from_entries unification and the repeated slug-to-identity helper extraction as follow-up cleanup rather than expanding this refactor further.

PR #348, opened 2026-03-14
feat: glyph showcase example

rjwittams:
## Summary
- Adds a scrollable ratatui example (`cargo run --example glyph_showcase`) that displays interesting Unicode glyphs for terminal UIs
- Covers box drawing, block elements, Braille patterns, arrows, geometric shapes, math symbols, status indicators, Nerd Font codepoints, Unicode separators, charts (bar, sparkline, braille, heatmap, dot plot, gauge), tree views, and practical TUI patterns
- Useful as a reference when designing TUI layouts and for comparing glyph support across terminals and fonts

## Test plan
- [x] `cargo build --example glyph_showcase` compiles cleanly
- [ ] `cargo run --example glyph_showcase` renders and scrolls with j/k/arrows/PgUp/PgDn/Home/End/q

### DIFF
--- a/examples/glyph_showcase.rs
+++ b/examples/glyph_showcase.rs
@@ -6,15 +6,16 @@ use std::io;
 
 use ratatui::{
     crossterm::event::{self, Event, KeyCode, KeyEventKind},
-    layout::{Constraint, Layout, Rect},
+    layout::{Constraint, Layout},
     style::{Color, Modifier, Style},
     text::{Line, Span},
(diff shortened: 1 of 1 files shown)


rjwittams:
Thanks for the review! All three items addressed in c9ce15b:

1. **Wrap/scroll mismatch** — dropped `Wrap`, using horizontal clipping instead. Line count now matches exactly.
2. **Cached `total_lines`** — computed once in `App::new()` and stored as a field.
3. **First-frame viewport** — viewport height is now captured from inside the `draw` closure rather than from `get_frame()` before drawing.

PR #370, opened 2026-03-15
fix: replay initial state on CLI watch connect

### LINKED ISSUE #356: investigate: validate and stabilize CLI watch before scoped filters
## Summary

Investigate the current CLI `watch` command behavior before adding scoped variants such as:

- `flotilla host <host> watch`
- `flotilla repo <path_or_slug> watch`
- `flotilla checkout <path> watch`

## Problem

`#284` originally included scoped watch syntax, but current behavior needs validation first. Adding filters on top of a stream path we do not trust risks baking in the wrong semantics.

## Goals

- Verify the current end-to-end behavior of `flotilla watch` for local and peer events
- Identify correctness gaps in subscription, replay, disconnect handling, or event attribution
- Fix any issues found, or document the invariant surface if behavior is already correct
- Leave a clear foundation for follow-up scoped watch filtering work

## Non-goals

- Implement scoped watch filters yet
- Extend topology rendering

## Related

- Split from #284
- Future follow-up: scoped watch command syntax and filtering

rjwittams:
## Summary

- `flotilla watch` now calls `replay_since` on connect so users see current repo state immediately, matching the TUI bootstrap pattern
- Warns on stderr if initial replay fails instead of silently showing nothing
- Deduplicates events between replay response and broadcast buffer using per-repo seq tracking
- Fixes `SocketDaemon::replay_since` seq seeding to use monotonic update, preventing spurious gap recovery when live events race ahead of the replay response

Closes #356

## Test plan

- [x] `cargo test --locked` — all pass
- [x] `cargo clippy --all-targets --locked -- -D warnings` — clean
- [x] Manual: `flotilla --config-dir /tmp/test watch` shows `[snapshot]` lines immediately on connect
- [x] Manual: `flotilla --config-dir /tmp/test watch --json` outputs full snapshot JSON on connect
- [ ] Verify no duplicate events when daemon is under load (subscribe-replay race window)

### DIFF
--- a/crates/flotilla-tui/src/cli.rs
+++ b/crates/flotilla-tui/src/cli.rs
@@ -450,14 +450,8 @@ pub async fn run_watch(socket_path: &Path, format: OutputFormat) -> Result<(), S
     match daemon.replay_since(&HashMap::new()).await {
         Ok(events) => {
             for event in &events {
-                match event {
-                    DaemonEvent::SnapshotFull(snap) => {
-                        replay_seqs.insert(snap.repo_identity.clone(), snap.seq);
-                    }
-                    DaemonEvent::SnapshotDelta(delta) => {
-                        replay_seqs.insert(delta.repo_identity.clone(), delta.seq);
-                    }
-                    _ => {}
+                if let Some((identity, seq)) = event_seq(event) {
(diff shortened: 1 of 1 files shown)


rjwittams:
All three fixed in 1e2dd3e: replay_seqs building now reuses event_seq(), uses monotonic max, and the dedup check destructures to (identity, seq).

PR #404, opened 2026-03-17
feat: symbolic step plans with per-host targeting and auto-workspace

rjwittams:
## Summary

- **Fix auto-workspace-after-checkout**: Workspace creation is now a step in the checkout plan itself, eliminating the race condition where the TUI auto-chained `CreateWorkspaceForCheckout` before provider data refreshed
- **Symbolic step actions**: Steps can now be closures or symbolic `StepAction` variants resolved at execution time by an `ExecutorStepResolver`, with access to fresh registry/config state
- **Per-step host targeting**: `StepHost::Local` / `StepHost::Remote(HostName)` declares where each step executes — foundation for multi-host step plans
- **Step outcome environment**: Steps receive prior outcomes as input instead of sharing mutable slots; later step failures preserve earlier meaningful results (e.g. `CheckoutCreated` survives workspace failure)
- **Shpool daemon stability**: Robust daemon state detection with pid files, three-way probe results (`Healthy`/`Inconclusive`/`Stale`), and conservative handling of legacy daemons without pid files

## Test plan

- [ ] Manual: create a local checkout from an issue — workspace should auto-create
- [ ] Manual: verify workspace creation failure shows step failed event but checkout still appears

### DIFF
--- a/crates/flotilla-core/src/executor.rs
+++ b/crates/flotilla-core/src/executor.rs
@@ -68,6 +68,10 @@ pub async fn build_plan(
     config_base: PathBuf,
     attachable_store: SharedAttachableStore,
     local_host: HostName,
+    // TODO(multi-host): When a command is forwarded from another host, this carries
+    // the requester's hostname so plan builders can stamp `StepHost::Remote(originator)`
+    // on steps that need to run back on the presentation host (e.g. workspace creation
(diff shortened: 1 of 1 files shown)


rjwittams:
Filed as #405 for follow-up. Thanks for the thorough reviews!

PR #414, opened 2026-03-18
feat: attachable set lifecycle — repo scoping, cascade delete, orphan reaping

### LINKED ISSUE #385: Define AttachableSet lifecycle and deletion policy
## Summary

Define explicit lifecycle rules for `AttachableSet` persistence so Flotilla can distinguish between:

- live terminal/workspace presence
- durable attachable identity
- destructive cleanup

We now treat persisted attachable identity as authoritative and terminal scans as liveness updates. The remaining gap is the lifecycle policy for when an `AttachableSet` should be retained, hidden, or deleted.

## Why this needs its own issue

The current registry model lets attachable sets persist indefinitely unless they are reused. That is acceptable for identity rollout, but it leaves key product and cleanup behavior undefined.

In particular, we need explicit rules for cases like:

1. A user closes a workspace but wants the attachable set to remain reusable.
2. A user wants to say "I never want to attach to this again".
3. A set is clearly owned by a checkout/workspace and should be deleted when that owner is destroyed.
4. A future unowned or cross-cutting set goes stale and should expire automatically.

## Desired outcome

Specify the policy for:

- when an `AttachableSet` remains durable after workspace/terminal disappearance
- when user intent should trigger explicit deletion
- when ownership should cause cascade deletion
- whether unowned sets should have timeout/staleness rules
- what should happen to provider bindings and member attachables when the set lifecycle changes

## Notes

This is separate from scan-time liveness:

- terminal scans should update presence/state
- attachable-set deletion should be an explicit lifecycle decision

A likely model is:

- presence changes mark members disconnected or absent
- attachable identity persists independently
- deletion comes from explicit user action, ownership teardown, or a defined stale-set policy

## Related

- #239 Managed terminal pool with identity and lifecycle management
- #378 tracking: AttachableSet identity rollout
- #360 Remote workspace correlation: associate workspace managers with managed terminal sets

rjwittams:
## Summary

- **Repo-scoped projection**: Sets filtered by checkout path match instead of terminal scan results. Fixes sets appearing in every repo and correlation flapping.
- **Cascade delete**: Removing a checkout deletes its owned attachable sets, members, and bindings, then tears down shpool sessions.
- **Orphan reaper**: Kills shpool sessions with no matching binding (private instance, so unbound = shouldn't exist).
- **Binding-based liveness**: Replaces `missed_scans` tracking. Known terminals emitted as Disconnected when shpool is down instead of vanishing.
- **Host-scoped cascade**: Cascade only affects local-host checkouts, not remote ones with the same branch name.

## Test plan

- [x] Store contract tests: `remove_set` and `sets_for_checkout` (both file-backed and in-memory)
- [x] Parse roundtrip tests for `parse_terminal_session_binding_ref`
- [x] Projection tests: repo-scoped filtering, stable without terminal scan
- [x] Executor tests: cascade delete removes store entries and kills terminals
- [x] Shpool tests: orphan reaping, skip reap when down, disconnected from bindings, emit disconnected when unreachable
- [x] Integration test: full lifecycle via InProcessDaemon (create → verify → delete → verify gone)
- [x] All 1615 tests pass, clippy clean

Closes #385

### DIFF
--- a/crates/flotilla-core/src/refresh.rs
+++ b/crates/flotilla-core/src/refresh.rs
@@ -1090,4 +1090,46 @@ mod tests {
 
         assert_eq!(pd.attachable_sets.len(), 1, "set should appear without terminal scan");
     }
+
+    #[test]
+    fn project_attachable_data_strips_orphaned_local_set_id_from_terminal() {
+        let store = crate::attachable::shared_in_memory_attachable_store();
+        let host = flotilla_protocol::HostName::local();
+        let checkout = flotilla_protocol::HostPath::new(host.clone(), "/repo/wt-deleted");
+
+        let set_id = {
+            let mut s = store.lock().expect("lock");
+            let set_id = s.ensure_terminal_set(Some(host.clone()), Some(checkout.clone()));
+            s.ensure_terminal_attachable(
+                &set_id,
+                "terminal_pool",
+                "shpool",
+                "flotilla/deleted/shell/0",
+                crate::attachable::TerminalPurpose { checkout: "deleted".into(), role: "shell".into(), index: 0 },
+                "bash",
+                std::path::PathBuf::from("/repo/wt-deleted"),
+                flotilla_protocol::TerminalStatus::Disconnected,
+            );
+            set_id
+        };
+
+        // Terminal references the set, but checkout is gone from pd.checkouts
+        let mut pd = ProviderData::default();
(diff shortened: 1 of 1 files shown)


rjwittams:
Added the missing unit test for the strip behaviour in 6c60f0b — `project_attachable_data_strips_orphaned_local_set_id_from_terminal`. Covers the exact scenario described: terminal with a binding to a local set whose checkout is absent from `pd.checkouts`, verifying both that the set is not projected and that the terminal's `attachable_set_id` is cleared.

PR #467, opened 2026-03-22
refactor: extract TerminalManager, replace ManagedTerminalId with AttachableId

rjwittams:
## Summary

- Extract terminal identity management from providers (cleat, shpool) into a new `TerminalManager` that owns the `SharedAttachableStore` for terminal concerns
- Replace `ManagedTerminalId` (checkout/role/index) with `AttachableId` as the sole terminal identity — the AttachableId doubles as the session name passed to the terminal pool
- Simplify `TerminalPool` trait to a pure CLI adapter (no store, no identity management)
- Eliminate terminal bindings (the session name IS the AttachableId)
- Remove `attachable_store` parameter from `Factory::probe`
- Remove unnecessary `Box` from `SharedAttachableStore` type alias

## Test plan

- [ ] All 1862 workspace tests pass
- [ ] `cargo clippy` clean
- [ ] `cargo +nightly-2026-03-12 fmt --check` clean
- [ ] Terminal status refresh works (reconcile_terminal_statuses updates store from pool before snapshot projection)
- [ ] Terminals appear in correlation/work items/TUI (ManagedTerminal restored with AttachableId key)

### DIFF
--- a/crates/flotilla-core/src/attachable/store.rs
+++ b/crates/flotilla-core/src/attachable/store.rs
@@ -365,6 +365,8 @@ impl AttachableStoreState {
 
     fn update_terminal_status(&mut self, id: &AttachableId, status: TerminalStatus) -> bool {
         if let Some(attachable) = self.registry.attachables.get_mut(id) {
+            // Irrefutable while AttachableContent has one variant; will become a
+            // compile error (forcing a decision) if a second variant is added.
             let AttachableContent::Terminal(ref mut terminal) = attachable.content;
             if terminal.status != status {
                 terminal.status = status;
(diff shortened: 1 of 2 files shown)


rjwittams:
All issues addressed:

**Fixed:**
- **cascade_delete save** — now calls `store.save()` after removals (9ac4950)
- **Idempotent allocation** — `allocate_set` returns existing set for checkout; `allocate_terminal` returns existing terminal if purpose matches (9ac4950)
- **refresh filters to terminals only** (9ac4950)
- **Session name inconsistency** — executor fully wired through TerminalManager (ccd8abf), both paths use `attachable_id.to_string()`
- **Irrefutable let** — kept as `let` with comment documenting intent (compile error on second variant is the desired behavior)

**Follow-up issues filed:**
- #468 — end-to-end test for terminal_ids correlation pipeline
- #469 — align ensure_running failure handling

**Acknowledged (no action):**
- TeleportFlow passes `None` for terminal_manager — intentional, teleport workspace creation uses pre-prepared commands from the remote host

PR #504, opened 2026-03-24
fix daemon connection lifecycle and routed peer sends

### LINKED ISSUE #492: bug: SocketDaemon drop does not tear down client connection
## Summary

Dropping `SocketDaemon` does not tear down the client connection. The background reader task owns cloned writer/state Arcs, but `SocketDaemon` stores no `JoinHandle` and has no `Drop` cleanup, so the daemon can keep a phantom client alive and delay idle shutdown after the caller drops the client.

## Details

**Location:** `crates/flotilla-client/src/lib.rs:58-101`

The `SocketDaemon` struct holds Arcs for writer, pending, next_id, and local_seqs but:
- The background reader task spawned at line 101 via `tokio::spawn()` discards the `JoinHandle`
- No `Drop` implementation exists on `SocketDaemon`
- The spawned task captures clones of all 5 Arcs (pending, writer, next_id, local_seqs, recovering)

When the caller drops `SocketDaemon`, the background reader task continues running, holding Arc references that keep the read half of the socket alive. The daemon-side `ClientConnection` only decrements `client_count` when it reads EOF — which may be delayed if the socket isn't fully closed.

## Impact

- Client-side resource leak (bounded — task eventually exits on EOF/error/runtime shutdown)
- Daemon idle shutdown can be delayed by phantom clients that haven't cleanly disconnected
- No way to explicitly close a `SocketDaemon` connection

## Fix Direction

- Store the `JoinHandle` in `SocketDaemon`
- Implement `Drop` (or an explicit `close()` method) that aborts the reader task and drops the write half
- This ensures the daemon sees EOF promptly after the client is done

### LINKED ISSUE #493: bug: PeerManager mutex held across awaited network sends
## Summary

Peer routing holds the `PeerManager` mutex across awaited network sends. A slow or wedged peer write can block unrelated routing, reconnect bookkeeping, and remote command handling, and it creates real starvation risk.

## Details

**Locations:**
- `crates/flotilla-daemon/src/server/remote_commands.rs:89-92, 276, 296, 316, 334`
- `crates/flotilla-daemon/src/peer/manager.rs:612, 699, 734`
- `crates/flotilla-daemon/src/peer/peer_runtime.rs:207-257`

The `tokio::sync::Mutex` is held across `send_to().await` calls in multiple locations:

```rust
// remote_commands.rs:89-92
let send_result = {
    let pm = self.peer_manager.lock().await;
    pm.send_to(&target_host, PeerWireMessage::Routed(routed)).await  // AWAIT INSIDE LOCK
};
```

The same pattern repeats for routed responses (lines 276, 296, 316, 334) and in `peer_runtime.rs` where `handle_inbound().await` plus nested `send_to().await` both happen under the lock.

## Impact

- A slow or wedged peer write blocks **all** unrelated routing, reconnect bookkeeping, and remote command handling
- Starvation scenario: Task A holds lock waiting on slow peer → Task B (reconnect) blocks → Task C (inbound from other peer) blocks
- `execute_forwarded_command()` holds the lock across multiple sequential `send_to()` calls in a loop

## Mitigating Factors

- Some scopes are already tight (lines 69-72 drop lock before send — good pattern)
- Channel-based senders buffer rather than block indefinitely
- `tokio::sync::Mutex` is async-aware so won't panic, but will cause starvation

## Fix Direction

Extract sender references outside the lock scope, then `.await` the send after releasing:

```rust
let sender = {
    let pm = self.peer_manager.lock().await;
    pm.resolve_sender(&target_host)?
};
// Lock released
sender.send(msg).await
```

Apply consistently across all identified locations.

### LINKED ISSUE #496: bug: daemon idle-timeout accounting ignores peer connections
## Summary

Only request clients increment `client_count`; inbound peer sessions do not, so a non-follower node can decide it is idle and shut down while it is still serving the mesh.

## Details

**Locations:**
- `crates/flotilla-daemon/src/server.rs:238-274` (idle timeout logic)
- `crates/flotilla-daemon/src/server/client_connection.rs:39, 107-109` (increment/decrement)
- `crates/flotilla-daemon/src/server/peer_connection.rs:24-31` (no count params)

`ClientConnection` increments/decrements `client_count` on connect/disconnect. `PeerConnection` receives **no** `client_count` or `client_notify` parameters — it doesn't participate in idle tracking at all.

The idle timeout watcher (line 238) fires when `client_count == 0`. The code recognizes this for follower nodes (comment: "disabled for follower-mode daemons which serve peer connections") but **non-follower nodes also accept inbound peer connections**.

## Scenario

1. Non-follower node has `follower: false`
2. Accepts inbound peer connections from mesh nodes
3. No request clients connected → `client_count == 0`
4. Idle timeout fires → node shuts down while actively serving peers

## Impact

- Mesh disruption: other nodes relying on this node for data replication or remote command forwarding will fail
- Silent failure: clean shutdown log with no indication of active peers
- Data loss risk: in-flight peer messages dropped

## Fix Direction

Include peer connections in activity tracking — either reuse `client_count` (renamed to `connection_count`) or add a separate `peer_count` that also suppresses idle shutdown.

rjwittams:
## Summary
- count inbound peer connections in daemon idle-shutdown accounting so active peers keep the daemon alive
- stop holding the peer manager lock across awaited peer sends, and drain queued sends in the channel test harness to match runtime behavior
- tear down `SocketDaemon` reader tasks on drop so client disconnects close promptly instead of leaving phantom connections

## Test Plan
- `mkdir -p .codex-tmp && TMPDIR="$PWD/.codex-tmp" cargo clippy --workspace --all-targets --locked -- -D warnings`
- `mkdir -p .codex-tmp && TMPDIR="$PWD/.codex-tmp" cargo test --workspace --locked`

Closes #492
Closes #493
Closes #496

### DIFF
--- a/crates/flotilla-core/src/agents/store.rs
+++ b/crates/flotilla-core/src/agents/store.rs
@@ -7,7 +7,7 @@ use std::{
 use flotilla_protocol::{AgentHarness, AgentStatus, AttachableId};
 use serde::{Deserialize, Serialize};
 
-use crate::{config::flotilla_config_dir, path_context::DaemonHostPath};
+use crate::path_context::DaemonHostPath;
 
 /// Persisted state for a single agent instance.
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
@@ -118,10 +118,6 @@ pub struct AgentStateStore {
 }
 
 impl AgentStateStore {
-    pub fn new() -> Self {
-        Self::with_path(flotilla_config_dir().join("agents").join("state.json"))
-    }
-
     pub fn with_base(base: &DaemonHostPath) -> Self {
         Self::with_path(base.join("agents").join("state.json"))
     }
@@ -144,12 +140,6 @@ impl AgentStateStore {
             Err(e) => return Err(format!("failed to read agent state: {e}")),
         };
         serde_json::from_str(&contents).map_err(|e| format!("failed to parse agent state: {e}"))
-    }
-}
-
-impl Default for AgentStateStore {
-    fn default() -> Self {
-        Self::new()
     }
 }
 
--- a/crates/flotilla-core/src/attachable/store.rs
+++ b/crates/flotilla-core/src/attachable/store.rs
@@ -13,10 +13,7 @@ use super::types::{
     Attachable, AttachableContent, AttachableId, AttachableSet, AttachableSetId, BindingObjectKind, ProviderBinding, TerminalAttachable,
     TerminalPurpose,
 };
-use crate::{
-    config::flotilla_config_dir,
-    path_context::{DaemonHostPath, ExecutionEnvironmentPath},
-};
+use crate::path_context::{DaemonHostPath, ExecutionEnvironmentPath};
 
 type BindingKey = (String, String, BindingObjectKind, String);
 
@@ -386,10 +383,6 @@ pub struct AttachableStore {
 }
 
 impl AttachableStore {
-    pub fn new() -> Self {
-        Self::with_path(flotilla_config_dir().join("attachables").join("registry.json"))
-    }
-
     pub fn with_base(base: &DaemonHostPath) -> Self {
         Self::with_path(base.join("attachables").join("registry.json"))
     }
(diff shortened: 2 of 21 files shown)


rjwittams:
Implemented in e2fbc57. Queued peer sends now carry the target host, dispatch_pending_sends logs peer and msg_kind on send failure, and queue_send_to now warns when resolve_sender fails instead of silently dropping the message. Verified with cargo +nightly-2026-03-12 fmt --check and cargo test -p flotilla-daemon --locked.

PR #512, opened 2026-03-25
fix: harden tui repo-scoped actions

### LINKED ISSUE #497: bug: background issue commands fail silently, can wedge UI
## Summary

`SetIssueViewport`/`FetchMoreIssues`/`SearchIssues`/`ClearIssueSearch` are spawned as fire-and-forget background tasks with errors discarded. The UI marks issue fetch/search as pending and only clears that state on later daemon events. If the command fails before any event, the UI can remain wedged.

## Details

**Locations:**
- `crates/flotilla-tui/src/app/executor.rs:30-35`
- `crates/flotilla-tui/src/app/key_handlers.rs:180-190`
- `crates/flotilla-tui/src/app/mod.rs:707, 785`

The error is literally discarded:
```rust
tokio::spawn(async move {
    let _ = daemon.execute(cmd).await;  // Error ignored
});
```

The most impactful case is `FetchMoreIssues`:
1. UI sets `issue_fetch_pending = true` before dispatch (key_handlers.rs:186)
2. Command is spawned as background
3. `issue_fetch_pending` is only cleared on daemon snapshot/delta events (mod.rs:707, 785)
4. If command fails before generating an event, the flag stays `true` permanently
5. The `!issue_fetch_pending` guard (key_handlers.rs:180) blocks further infinite scroll
6. No timeout, no error display — requires app restart to recover

## Impact

- `FetchMoreIssues` failure permanently blocks infinite scroll (most severe)
- `SearchIssues` failure is less visible but still loses the user's action
- No error feedback to the user in any case

## Fix Direction

- Capture errors from `daemon.execute()` and surface them (status bar message or clear pending state)
- Add a timeout fallback to clear `issue_fetch_pending` if no daemon event arrives
- Consider removing the background-command special case and treating all commands through the pending state system

### LINKED ISSUE #498: bug: overview tab can panic on repo-scoped palette/search actions
## Summary

The overview tab exposes repo-scoped palette/search actions that assume an active repo exists. Several handlers index `ctx.repo_order[ctx.active_repo]` without guards, so opening the palette or issue search from overview/no-repo state can panic with index-out-of-bounds.

## Details

**Locations:**
- `crates/flotilla-tui/src/widgets/command_palette.rs:63, 105, 142`
- `crates/flotilla-tui/src/widgets/issue_search.rs:41, 54`
- `crates/flotilla-tui/src/widgets/overview_page.rs:56`

Unchecked access:
```rust
let repo_identity = ctx.repo_order[ctx.active_repo].clone();  // panics if empty
```

The overview page explicitly exposes `Action::OpenCommandPalette` (overview_page.rs:56). The `active()` method in the model also does unguarded indexing: `&self.repos[&self.repo_order[self.active_repo]]`.

Many other code paths already guard against `repo_order.is_empty()` (app/mod.rs:496, widgets/tabs.rs:268, app/navigation.rs:22/37/67, app/key_handlers.rs:154).

## Impact

- Panic (index out of bounds) if command palette or issue search is triggered with no repos configured
- Requires the relatively rare state of empty `repo_order` at runtime

## Fix Direction

Add the same `repo_order.is_empty()` guards used elsewhere, returning early or showing a "no repos" message.

rjwittams:
## Summary
- guard repo-scoped command palette and issue search actions so overview and empty-repo states show `No active repo` instead of panicking
- route immediate background issue-command dispatch failures back into app state so `issue_fetch_pending` and related status stay recoverable
- add focused regressions for both bug paths and save the short implementation plan

Fixes #497
Fixes #498

## Test Plan
- [x] `cargo test -p flotilla-tui --locked`

### DIFF
--- a/crates/flotilla-core/src/config.rs
+++ b/crates/flotilla-core/src/config.rs
@@ -313,7 +313,11 @@ pub fn path_to_slug(path: &Path) -> String {
 }
 
 fn repo_file_key(path: &Path) -> String {
-    urlencoding::encode(&path.to_string_lossy()).into_owned()
+    let key = urlencoding::encode(&path.to_string_lossy()).into_owned();
+    if key.len() > 200 {
+        tracing::warn!(key_len = key.len(), path = %path.display(), "repo config filename key is close to filesystem limit");
+    }
+    key
 }
 
 /// Owns daemon-side paths and caches the global `FlotillaConfig`.
--- a/crates/flotilla-tui/src/app/mod.rs
+++ b/crates/flotilla-tui/src/app/mod.rs
@@ -488,8 +488,12 @@ impl App {
                                 }
                             }
                         }
+                        // These commands do not set per-repo pending flags, so
+                        // there is nothing to unwind beyond surfacing the error.
                         CommandAction::SearchIssues { .. } | CommandAction::ClearIssueSearch { .. } => {}
-                        _ => {}
+                        other => {
+                            tracing::warn!(action = ?other, "unexpected background issue command failure");
+                        }
                     }
                     self.set_status_message(Some(error));
                 }


rjwittams:
Fixed in 1318ead. repo_file_key now warns when the encoded filename key approaches filesystem limits, and drain_background_updates now documents the no-cleanup search paths while warning on any unexpected background issue command failure.

PR #553, opened 2026-03-28
fix: scope Codex session discovery by repo

### LINKED ISSUE #495: bug: Codex session discovery is not repo-scoped (cache + fallback)
## Summary

`CodexCodingAgent` session discovery is not repo-scoped in two ways: the environment-id cache is global instead of keyed by repo slug, and the fallback path matches by bare repo name rather than full owner/repo.

## Details

**Location:** `crates/flotilla-core/src/providers/coding_agent/codex.rs`

### Problem 1: Global environment cache (lines 212-215, 383-405)

`EnvCache` stores a flat `Vec<String>` with no repo key:

```rust
struct EnvCache {
    environment_ids: Vec<String>,
    loaded_at: Option<Instant>,
}
```

The cache is populated for the first `repo_slug` that calls `list_sessions()`, then reused for all repos for the 10-minute TTL. A single `CodexCodingAgent` instance serves all repos in the daemon.

### Problem 2: Bare-name fallback matching (lines 324-347)

`fetch_tasks_by_label()` extracts only the bare repo name:

```rust
let repo_name = repo_slug.rsplit_once('/').map(|(_, name)| name).unwrap_or(repo_slug);
```

Then filters by `label.eq_ignore_ascii_case(repo_name)`. Repos `orgA/flotilla` and `orgB/flotilla` both match tasks labeled `"flotilla"`.

The primary API path (line 268-283) correctly uses full `owner/repo` — the problem is only in the fallback triggered on network error or empty env list.

## Impact

- Cross-repo data leakage: unrelated repos can surface each other's Codex tasks
- Cache contamination persists for 10 minutes per TTL cycle
- Same-named repos across different orgs will cross-contaminate in fallback path

## Fix Direction

1. Key the cache by `repo_slug`: `HashMap<String, (Vec<String>, Instant)>`
2. Use full `owner/repo` slug (or at minimum validate both components) in fallback matching

rjwittams:
## Summary

- Environment ID cache was global (shared across all repos) — now keyed by repo slug via `HashMap<String, EnvCacheEntry>`
- Fallback path matched tasks by arbitrary environment label (bare repo name), causing cross-contamination between same-named repos in different orgs — now uses `GET /wham/environments` and matches by `repo_map.repository_full_name`
- Deduplicates env IDs when multiple `repo_map` refs point at the same repo

## Details

Codex "environments" are user-created entities with arbitrary labels — they're not inherently named after repos. The old fallback compared `task_status_display.environment_label` against the bare repo name (e.g. "flotilla"), which would match across orgs. The new fallback fetches all environments and checks each environment's `repo_map[*].repository_full_name` for the target `owner/repo` slug, verified against the live Codex API.

The fallback fixture (`codex_label_fallback.yaml`) is synthetic since it tests error-recovery logic requiring a 500 from the by-repo endpoint. The cache isolation fixture (`codex_cache_isolation.yaml`) is also synthetic but covers the behavioral contract of calling `list_sessions()` for two repos on one agent instance.

Closes #495

## Test plan

- [x] `fallback_uses_repo_map_not_label` — verifies fallback scopes by `repo_map` not label, caches discovered env IDs
- [x] `env_cache_isolates_repos` — behavioral test: two repos on one agent get independent env ID caches
- [x] Existing codex tests pass (29 total)
- [x] Full workspace test suite passes
- [x] `cargo clippy`, `cargo +nightly-2026-03-12 fmt --check` clean

### DIFF
--- a/crates/flotilla-core/src/providers/coding_agent/codex.rs
+++ b/crates/flotilla-core/src/providers/coding_agent/codex.rs
@@ -86,7 +86,6 @@ struct FullEnvironmentInfo {
 
 #[derive(Debug, Deserialize)]
 struct EnvironmentRepo {
-    #[serde(default)]
     repository_full_name: Option<String>,
 }
 
@@ -503,7 +502,7 @@ impl super::CloudAgentService for CodexCodingAgent {
                         }
                     }
                     Err(e) => {
-                        debug!(provider = "codex", error = %e, "environment lookup failed, falling back to label match");
+                        debug!(provider = "codex", error = %e, "by-repo lookup failed, falling back to all-environments repo_map");
                         return self.fallback_via_all_environments(repo_slug, &auth).await;
                     }
                 }


rjwittams:
Thanks for the thorough review.

**Synthetic fixtures:** Added a comment to `codex_cache_isolation.yaml` explaining why it's synthetic (requires two repos each with active Codex tasks — hard to maintain as stable test state). The `codex_label_fallback.yaml` already had this treatment. Both have TODOs for when recording infrastructure supports these cases.

**No negative cache:** Valid observation. Filed as #555 — a short-TTL negative cache would reduce repeated API round trips for repos with no Codex environments.

**Auth retry gap in fallback:** Also valid. Filed as #556 — the fallback path should mirror the primary path's retry logic. Matches existing behaviour of the old `fetch_tasks_by_label` so not a regression.

**Partial move correctness and cache trigger paths:** Confirmed, thanks for verifying.

PR #576, opened 2026-04-13
[codex] fix issue-row actions and branch naming fallbacks

rjwittams:
## Summary
- make branch-name generation keep the originally requested issue IDs and fetch missing issue details from the preferred tracker before asking AI for a name
- preserve checkout host/source information for environment-qualified paths so remote checkout rows render against the right host context
- treat native TUI issue rows like selectable actionable items so multi-select, action menus, and branch-name generation work consistently across issue sections

## Why
These fixes address two related gaps in repo-scoped issue workflows. Branch-name generation could lose issue context when the current provider snapshot did not already contain every requested issue, and native issue rows in the TUI did not participate in the same action and multi-select paths as work-item rows. Remote checkout path rendering also lacked a reliable host fallback in some environment-qualified cases.

## Impact
Issue-driven workflows stay usable even when the issue metadata has to be fetched lazily, and users can generate branches directly from native issue rows or issue-row multi-select without falling back to work-item-backed sections. Remote checkout rows also show paths relative to the correct host root more reliably.

## Root Cause
The executor only consulted in-memory issue data when building branch-name context and derived fallback issue/provider pairs from resolved issues rather than the user-requested IDs. In the TUI, several selection and intent paths were hard-wired to `WorkItem` rows and ignored native `IssueRow` identities.

## Validation
- `mkdir -p .codex-tmp && TMPDIR="$PWD/.codex-tmp" cargo clippy --workspace --all-targets --locked -- -D warnings`
- `mkdir -p .codex-tmp && TMPDIR="$PWD/.codex-tmp" cargo test --workspace --locked` *(fails in this sandbox on existing machine-identity-dependent tests: `host_identity::tests::resolve_local_node_id_uses_machine_scoped_identity_dir`, `in_process::tests::get_repo_providers_uses_preferred_root_environment_host_discovery_for_non_local_direct_repo`)*
- `mkdir -p .codex-tmp && TMPDIR="$PWD/.codex-tmp" cargo test -p flotilla-core --locked generate_branch_name`
- `mkdir -p .codex-tmp && TMPDIR="$PWD/.codex-tmp" cargo test -p flotilla-core --locked checkout_source_falls_back_to_provider_host_for_environment_qualified_paths`
- `mkdir -p .codex-tmp && TMPDIR="$PWD/.codex-tmp" cargo test -p flotilla-tui --locked app::key_handlers::tests`
- `mkdir -p .codex-tmp && TMPDIR="$PWD/.codex-tmp" cargo test -p flotilla-tui --locked widgets::repo_page::tests`
- `mkdir -p .codex-tmp && TMPDIR="$PWD/.codex-tmp" cargo test -p flotilla-tui --locked widgets::split_table::tests`

### DIFF
--- a/crates/flotilla-core/src/data.rs
+++ b/crates/flotilla-core/src/data.rs
@@ -393,9 +393,11 @@ fn group_to_work_item(providers: &ProviderData, group: &CorrelatedGroup, group_i
     let description = pr_title.or(session_title).or(agent_title).or_else(|| branch.clone()).or(set_description).unwrap_or_default();
 
     let source = match &anchor {
-        CorrelatedAnchor::Checkout(co) => {
-            co.host_path().map(|path| path.host.to_string()).or_else(|| co.key.host_name().map(ToString::to_string))
-        }
+        CorrelatedAnchor::Checkout(co) => co
+            .host_path()
+            .map(|path| path.host.to_string())
+            .or_else(|| co.key.host_name().map(ToString::to_string))
+            .or_else(|| host.as_ref().map(ToString::to_string)),
         CorrelatedAnchor::AttachableSet(id) => providers.attachable_sets.get(id).and_then(|set| {
             set.checkout
                 .as_ref()
--- a/crates/flotilla-core/src/data/tests.rs
+++ b/crates/flotilla-core/src/data/tests.rs
@@ -226,6 +226,32 @@ fn host_falls_back_to_checkout_host_provenance_when_key_has_host_id() {
 }
 
 #[test]
+fn checkout_source_falls_back_to_provider_host_for_environment_qualified_paths() {
+    let checkout_key = flotilla_protocol::qualified_path::QualifiedPath::environment(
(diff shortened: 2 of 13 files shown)


rjwittams:
Pushed a follow-up fix in f4b77742. I updated GenerateBranchName so unresolved issue IDs still enter the AI path with placeholder context like issue 42 instead of short-circuiting straight to the fallback; that restores the cancellable in-flight provider behavior and fixes the CI regression. I also switched the fetched-issue merge from or_insert to insert, and made ArchiveSession use the same effective execution context as GenerateBranchName for symmetry. I left the synthetic issue-row node id as-is: this helper work item is only used locally for intent availability and action-menu construction, while command routing stays repo-scoped rather than depending on that synthetic node id.

rjwittams:
Addressed in b0272e1c. I added a short comment on the synthetic issue-row WorkItem explaining that my_node_id() should already be known for this local repo view and that the issue-row fallback node id is only a safety net.

PR #579, opened 2026-04-14
feat: prototype flotilla resources and workflow templates

rjwittams:
## Summary
- add the stage-1 `flotilla-resources` crate with typed resource APIs, in-memory semantics, and raw Kubernetes HTTP support
- add the stage-2 `WorkflowTemplate` resource with validation, CRD schema, sample YAML, local validation example, and backend coverage
- tighten the workflow template design docs and capture the implementation plan used for this branch

## Test Plan
- [x] `cargo test -p flotilla-resources --locked`
- [x] `cargo build -p flotilla-resources --locked --example validate_workflow --example k8s_crud`
- [x] `cargo run -p flotilla-resources --example validate_workflow -- crates/flotilla-resources/examples/review-and-fix.yaml`
- [ ] `FLOTILLA_RUN_K8S_TESTS=1 cargo test -p flotilla-resources --locked -- --ignored`


rjwittams:
Fixed in 5e10937a.

- re-exported `WatchStream` from `lib.rs`
- switched the workflow-template stale-update assertion to `expect_err` to satisfy Clippy
- removed the duplicate `Visiting` path from DAG traversal so cycle detection has one clear route
- strip trailing `
` when decoding watch events and added CRLF coverage in `http_wire`
- made the kubeconfig auth limitation explicit for token auth and missing cert/key cases
- documented the in-memory event-log growth tradeoff and `ensure_crd` replace semantics

I did not change the workspace-dependency nit because those crates are not centralized in `[workspace.dependencies]` in this repo today, and I left `ResourceError::decode` as-is for the prototype.

PR #581, opened 2026-04-14
[codex] implement convoy resource stage 3

rjwittams:
## Summary
- add typed `StatusPatch` support to `flotilla-resources` and implement the Stage 3 `Convoy` resource, reconcile logic, CRD, controller example, and end-to-end tests
- extend the resource test surface with convoy reconcile/status-patch coverage plus Kubernetes integration coverage updates
- seed deterministic `machine_id` values in daemon-backed test harnesses so the sandbox-safe workspace test command no longer depends on ambient host identity

## Why
- Stage 3 needs a concrete convoy resource/controller implementation on top of the Stage 1 resource client surface
- the workspace test failure was not sandbox-specific; several daemon harnesses were constructing local configs without `machine_id`, which broke on hosts without ambient identity

## Validation
- `/bin/zsh -lc 'mkdir -p .codex-tmp && TMPDIR="$PWD/.codex-tmp" cargo clippy --workspace --all-targets --locked -- -D warnings'`
- `/bin/zsh -lc 'mkdir -p .codex-tmp && TMPDIR="$PWD/.codex-tmp" cargo test --workspace --locked --features flotilla-daemon/skip-no-sandbox-tests'`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -421,9 +421,9 @@ dependencies = [
 
 [[package]]
 name = "cc"
-version = "1.2.57"
+version = "1.2.58"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7a0dd1ca384932ff3641c8718a02769f1698e7563dc6974ffd03346116310423"
+checksum = "e1e928d4b69e3077709075a938a05ffbedfa53a84c8f766efbf8220bb1ff60e1"
 dependencies = [
  "find-msvc-tools",
  "jobserver",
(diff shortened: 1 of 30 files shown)


rjwittams:
Fixed. The controller now treats WorkflowTemplate not-found as reconcile input instead of returning early, so missing templates transition convoys to Failed through the existing FailInit path. I added an in-memory controller regression test for that case, and moved tracing-subscriber to dev-dependencies because it is only used by the example binary.

PR #582, opened 2026-04-14
feat: add task provisioning runtime

rjwittams:
## Summary
- add the generic provisioning resource surface, controller framework, and the new `flotilla-controllers` crate for environment, clone, checkout, terminal session, and task workspace reconciliation
- extend convoy provisioning to create and observe `TaskWorkspace` resources, add the `convoy <name> task <task> complete` command path, and wire daemon startup to host the provisioning runtime and heartbeat loop
- split the production daemon into `flotillad`, keep the root `flotilla` binary client-only, and update the spec plus command-palette snapshots for the new convoy command surface

## Test Plan
- [x] `cargo test -p flotilla-daemon runtime::tests::in_memory_stage4a_flow_reaches_running_and_completes_convoy -- --nocapture`
- [x] `cargo test -p flotilla-daemon --locked`
- [x] `cargo test -p flotilla-client --locked`
- [x] `cargo test --locked src::tests`
- [x] `/bin/zsh -lc 'mkdir -p .codex-tmp && TMPDIR="$PWD/.codex-tmp" cargo clippy --workspace --all-targets --locked -- -D warnings'`
- [x] `/bin/zsh -lc 'mkdir -p .codex-tmp && TMPDIR="$PWD/.codex-tmp" cargo test --workspace --locked --features flotilla-daemon/skip-no-sandbox-tests'`


rjwittams:
### Code review

Found 1 issue:

1. Finalizer is attached lazily after resource creation, leaving a race window where a resource can be deleted before the controller has processed it — `run_finalizer` will not run because the deletion-path branch is gated on the finalizer already being present. `Actuation::Create*` calls `create_if_missing`, which creates the object with `finalizers: Vec::new()`, and the finalizer is only appended later when the primary-watch event is pulled off the queue. If a deletion happens (or arrives first via the watch) before that attachment, the cleanup hook is silently skipped and on-disk state (worktrees, clones, containers, sessions) leaks.

Creation path (empty finalizers):
https://github.com/flotilla-org/flotilla/blob/8fe9b78a0b06a8b8fca7e2d428c9fb54c99908c3/crates/flotilla-resources/src/controller/mod.rs#L247-L265

Attachment is gated on `deletion_timestamp.is_none()`, and the deletion-path branch requires the finalizer to already be present:
https://github.com/flotilla-org/flotilla/blob/8fe9b78a0b06a8b8fca7e2d428c9fb54c99908c3/crates/flotilla-resources/src/controller/mod.rs#L368-L401

Consider either having the actuator include the finalizer in the initial `InputMeta`, or treating "deletion_timestamp set + finalizer absent + reconciler has a finalizer name" as a state that still triggers `run_finalizer`.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rjwittams:
Addressed the actionable review items in `b9f59069`.

In this PR:
- `TerminalSessionReconciler` now uses the injected `now` value when marking failures, with a focused regression test.
- Removed the no-op `resource_backend_for_test()` wrapper and switched the test to the existing public `resource_backend()` accessor.
- Cleaned up the duplicated per-test `meta()` helpers by reusing the shared helper.
- Removed the dead suppression line from `DockerEnvironmentActuator` by dropping the unused stored fields.
- Documented the intended `fetch_dependencies()` contract in the controller framework so the idempotent prepare/inspect usage in clone/checkout is explicit rather than implicit.

Already fixed earlier on this branch:
- watch restart backoff and restart supervision inside `ControllerLoop`
- finalizer attachment / teardown behavior, including the lazy-finalizer race called out later in the thread

Filed follow-ups for the two valid wider-scope concerns that I did not fold into this PR:
- #583 supervise provisioning controllers after reconcile-path errors
- #584 thread provisioning namespace into convoy completion handling

PR #585, opened 2026-04-15
refactor: reduce resource client prototype code size

rjwittams:
## Summary
- add `bon` and `rstest` as the normal builder/parameterized-test tools for the resources prototype work
- extract shared controller/resource test harnesses and a reusable backend contract suite
- move provisioning happy-path coverage outward into controller-loop / observable-behavior tests while trimming duplicated unit-test setup
- factor daemon runtime git test setup into a dedicated helper and finish the controller-loop harness cleanup
- include the phase-plan docs that drove this cleanup pass

## Testing

## Notes
- `crates/flotilla-resources/tests/http_wire.rs` was inspected but left alone because it is exercising HTTP wire semantics rather than the generic backend contract shape, so converting it was not a trivial cleanup in this pass.
- Representative line-count reductions from the cleanup:
  - `crates/flotilla-daemon/src/runtime.rs`: `1141 -> 1106`
  - `crates/flotilla-controllers/tests/task_workspace_reconciler.rs`: `441 -> 300`
  - `crates/flotilla-resources/tests/in_memory.rs`: `239 -> 114`
  - `crates/flotilla-resources/tests/workflow_template_in_memory.rs`: `116 -> 56`


rjwittams:
Fixed in 993cc37d. I extracted the shared wait_until helper for TestLoopHarness, replaced the too_many_arguments test helpers with bon builder-backed fixture structs, relaxed the shared contract assertions so they no longer assume monotone integer resourceVersions, and added comments explaining the single-case rstest scaffolding for future backend fixtures. I also ran cargo +nightly-2026-03-12 fmt --check and cargo test -p flotilla-resources -p flotilla-controllers --locked locally.

rjwittams:
Fixed in ef090889. I dropped InputMeta::with_finalizers since it had no callers; the controller loop only uses with_added_finalizer/without_finalizer. Re-ran cargo +nightly-2026-03-12 fmt --check and cargo test -p flotilla-resources --locked locally.

Issue #589, opened 2026-04-22
TUI: arbitrary tab model


rjwittams:
Tabs are currently overview or repo and all repos are opened 

Make them arbitrary widgets
* convoys 
* single convoy/maybe task or vessel
* approvals/events from hooks
* project (multiple repo + subdir)
* workflow
* persistent agent 
* welcome / onboarding 
* etc 

Each tab has an address for deep link opening just that tab in another TUI (-> shove to presentation manager). Eventually same tab can open in the web

rjwittams:
Note from PR #590 work — flagging a related structural concern for whenever the arbitrary-tab model is designed:

**`tab_order` vs `repo_order`:** `TuiModel.repo_order: Vec<RepoIdentity>` currently conflates "the list of repo-type tabs" with "tab-bar order." When we added the Convoys tab, `{` / `}` (move-tab) became awkward — the action dispatches on `model.active_repo` regardless of which tab type is active, so pressing `{` on Convoys silently reorders the invisible repo list. PR #590 fixed the symptom by moving those bindings from `TabPage` back to `Normal`, but the underlying shape is still "repo list happens to equal the reorderable section of the tab bar."

In an arbitrary-tab world this probably wants:
- `tab_order: Vec<TabId>` as the canonical tab-bar order (with positional constraints like Flotilla-always-first, Add-always-last expressed as policy).
- `reorderable` (and perhaps `addressable`, `closable`) as per-tab-kind metadata.
- `MoveTabLeft`/`Right` action handlers consult the metadata rather than implicitly assuming a repo context.
- Only one place persists tab order (today the daemon has its own `repo_order` field in `in_process.rs:453` that isn't synced to the TUI; shared `tab-order.json` via `ConfigStore` is how both currently overlap).

Also: there's a latent binding-vs-action scope mismatch that the arbitrary-tab design should formalise — "app-global bindings" (like `/`, `q`, `?`) and "tab-kind-specific actions" (like `MoveTab`, which only makes sense for reorderable kinds) are different things, but the current model exposes them at the same `BindingModeId` layer.

Not asking to solve this here — just capturing the observations so the tab model design has them.

rjwittams:
[...]

PR #590, opened 2026-04-23
feat: stage 6 PR 1 — read-only convoy TUI view

rjwittams:
## Summary

First PR for stage 6 of the convoy implementation plan. Read-only convoy view end-to-end:

- New namespace-scoped stream (`StreamKey::Namespace`) carrying convoy summaries from daemon to TUI.
- `ConvoyProjection` in the daemon watches `Convoy` + `Presentation` resources and emits `NamespaceSnapshot` / `NamespaceDelta` events.
- New `Convoys` tab in the TUI rendering list + detail (task DAG via `tui-tree-widget`) with status glyphs.
- `flotilla-client` + CLI `watch` replay / gap-recovery extended for the new stream type.
- Selection / filter / auto-select behaviour implemented; initial snapshot state populates the tab.

Spec: `docs/superpowers/specs/2026-04-21-tui-convoy-view-design.md` (PR 1 slice). Per-task Presentation addendum (`docs/superpowers/specs/2026-04-22-per-task-presentation-design.md`) is a prerequisite for the PR 3 (task attach) follow-up, not for this PR.

## Test plan

- [x] `cargo test --workspace --locked --features flotilla-daemon/skip-no-sandbox-tests`
- [ ] Verify Convoys tab renders in a live TUI session with a convoy created via the backend.


rjwittams:
## Review round addressed (commits `33c71941`, `6e840b2f`, `d4594620`)

All four main findings and all five minor items from the review rounds are fixed:

**Main:**
1. Filter duplication — convoy filter now lives in one place (`filter_convoys_by_str` free function in `app/mod.rs`), consumed by `RenderContext`. Free function rather than `App::visible_convoys(&self)` method because the draw closure needs `&mut app.ui` alongside the filter — borrow splitting via the free function avoids the conflict. `screen.rs` no longer re-implements the filter.
2. `ConvoyScope` enum + `scope` field removed from `ConvoysPage`. Per-scope filtering is applied at the `App` layer; a scope field can come back when scope-specific rendering is actually needed.
3. `#[allow(dead_code)]` removed from `wire_convoy_phase`, `wire_task_phase`, `summarize_task`, `summarize_convoy`, and where possible on `NamespaceView` / `ConvoyProjection`. Compiler stays quiet after removal.
4. `namespace_state` correctness: delta-arm now skips when no prior snapshot exists (logs a warning, awaits the next snapshot) instead of synthesizing `seq: 0`. On `RecvError::Lagged`, the mirror is cleared to avoid serving known-stale state to reconnecting clients. Proper structural fix (have `replay_since` consult `ConvoyProjection` directly) filed as #591.

**Minor:**
- `NamespaceModel` uses `#[derive(Default)]`.
- Redundant `fn presentation(...)` test helper removed in favour of `presentation_obj`.
- Stale `// TODO: namespace stream support` comments removed from `cli.rs`.
- `TreeState::default()`-per-render in `detail.rs` has a `TODO(task-attach)` pointing at where it should be lifted into `ConvoysUiState`.
- Hardcoded `"flotilla"` namespace now has a single pointer to #589 (arbitrary tabs) noting where multi-namespace scoping will be added.

Tests: full workspace green. No snapshot changes.

Thanks for the repeated rounds — persistence paid off.

PR #592, opened 2026-04-27
feat(tui): convoy task completion via command palette

rjwittams:
## Summary

Stage 6 PR 2: wire `x` on the Convoys tab to complete the selected task by opening the command palette pre-filled with the existing `convoy <id> task <name> complete` noun-verb command. The palette is the action surface — no bespoke confirm widget.

- Convoys tab gains nested focus. `j/k/up/down` move within the focused pane; `enter/l/right` drill from the convoy list into the task tree; `esc/left` returns. Bindings match the rest of the app (vim + arrow keys both work).
- `x` in the task tree opens the palette pre-filled with `convoy <id> task <task> complete `; pressing Enter dispatches the existing `ConvoyTaskComplete` command.
- Palette gains native `convoy` subject completions (active convoys from the namespace state) and task-name completions after `convoy <id> task <Tab>`. Typing the command directly works on any tab.
- Border style indicates focused pane; selected task highlighted in the tree.
- Plumbs `selected_task`, `focus`, and `namespaces` through `RenderContext` / `WidgetContext` so the convoys page widgets and the palette completion path can read them.

## Test plan
- [x] 884 unit tests pass (`cargo test -p flotilla-tui --lib --locked`)
- [x] 2 E2E tests pass (`cargo test -p flotilla-tui --test convoy_view_e2e --locked`) — the new test drives a real `InProcessDaemon` end-to-end: bootstraps a convoy with a task, types `l → x → Enter`, and asserts `TaskPhase::Completed` propagates back to the TUI's namespace state via `NamespaceDelta`
- [x] Workspace tests pass (`cargo test --workspace --locked`)

Branched off `main` after PR #590 merged. Stage 6 PR 3 (task attach) depends on the per-task Presentation addendum (`docs/superpowers/specs/2026-04-22-per-task-presentation-design.md`) and is not blocked by this PR.

### DIFF
--- a/crates/flotilla-core/src/config.rs
+++ b/crates/flotilla-core/src/config.rs
@@ -131,6 +131,8 @@ pub struct KeysConfig {
     #[serde(default)]
     pub convoys: HashMap<String, String>,
     #[serde(default)]
+    pub convoy_tasks: HashMap<String, String>,
+    #[serde(default)]
     pub action_menu: HashMap<String, String>,
     #[serde(default)]
     pub delete_confirm: HashMap<String, String>,
--- a/crates/flotilla-tui/src/app/key_handlers.rs
+++ b/crates/flotilla-tui/src/app/key_handlers.rs
@@ -405,11 +405,17 @@ impl App {
     }
 
     /// Open the command palette pre-filled with `convoy <id> task <task> complete `.
-    /// No-op when no convoy or task is selected.
+    /// No-op when no convoy or task is selected. Convoy ids and task names are
+    /// arbitrary strings (no validation), so they are routed through the
+    /// palette's quoting helper to round-trip whitespace and special characters.
     pub(super) fn open_complete_convoy_task_palette(&mut self) {
         let Some(convoy_id) = self.convoys_ui.selected.as_ref() else { return };
         let Some(task) = self.convoys_ui.selected_task.as_ref() else { return };
-        let prefill = format!("convoy {} task {} complete ", convoy_id.name(), task);
+        let prefill = format!(
+            "convoy {} task {} complete ",
+            crate::palette::quote_palette_token(convoy_id.name()),
+            crate::palette::quote_palette_token(task),
+        );
         let widget = crate::widgets::command_palette::CommandPaletteWidget::with_prefill(prefill, None);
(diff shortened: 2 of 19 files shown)


rjwittams:
Fixed in cb9fde6e.

1. **Task names with spaces** — verified there's no validation on task names in `WorkflowTemplate`, so this was a real footgun. Added a `quote_palette_token` helper next to the tokenizer that wraps in double quotes and escapes embedded `"`/`\` when needed, used it in the prefill for both the convoy id and task name. Round-trip property test exercises plain identifiers, whitespace, embedded quotes, and backslashes. Plus an integration test confirming `x` on a task named `"fix my bug"` produces `convoy fix-bug-123 task "fix my bug" complete `.

2. **Focus stuck on Tasks when clamped** — agreed, that visual limbo wasn't intentional. `clamp_selected_task` now also resets `focus` to `List` when it nulls out `selected_task`. Updated `delta_removing_selected_task_clamps_to_none` (renamed `_and_drops_focus`) to assert the new behavior.

3. **Dead `partial` branch** — replaced with the two early-returns you suggested.

`{:?}` for phase descriptions noted; will switch to `Display` when/if the enums grow one.

rjwittams:
Added the single-quote case (`it's`) to the round-trip test in 62b19849.

On the snapshot test concern — the `convoy_list_snapshot_three_phases` snapshot file on disk hasn't been touched by this PR (verified with `git diff origin/main -- crates/flotilla-tui/src/widgets/convoys_page/snapshots/`). Insta uses `ratatui::TestBackend` which renders character-only — bold border style isn't captured in the snapshot text, so the `focused: true` change doesn't show up there. The new `convoy_detail_with_selected_task_renders` test instead asserts the rendered text directly without insta to cover the style-changed paths, since insta wouldn't see the difference.

Linear scan in `convoy_task_completions` and `{:?}` for phase descriptions both noted as deferred, agreed they're not worth doing now.

rjwittams:
Added the limitation comment in f108ed7d. Verified the Enter-dispatch path uses `tokenize_palette_input` (quote-aware) so the command dispatches correctly even when Tab completion would mis-tokenize a hypothetical whitespace-bearing convoy id.

PR #596, opened 2026-05-13
refactor(daemon): thread provisioning namespace into convoy completion

### LINKED ISSUE #584: Thread provisioning namespace into convoy completion handling
From PR #582 review: https://github.com/flotilla-org/flotilla/pull/582#issuecomment-4248236077

InProcessDaemon handles MarkConvoyTaskComplete by reading Convoy resources from a hardcoded flotilla namespace, while provisioning startup/runtime uses RuntimeOptions::namespace.

Today this is not user-visible because Stage 4a runs in the default namespace, but it leaves convoy completion as the one provisioning path that bypasses namespace configuration.

Scope:
- choose a single source of truth for the provisioning/resource namespace
- thread that through InProcessDaemon convoy completion handling
- update tests so convoy completion works against non-default namespaces too

rjwittams:
## Summary
- Replace the hardcoded `"flotilla"` namespace in `InProcessDaemon`'s `ConvoyTaskComplete` handler with a configurable per-daemon namespace, set from `RuntimeOptions::namespace` at daemon-runtime startup.
- Make `flotilla_core::in_process::DEFAULT_PROVISIONING_NAMESPACE` the single source of truth for the default; `RuntimeOptions::default()` now pulls from it instead of duplicating the literal.

## Why
From PR #582 review: convoy completion was the one provisioning path that ignored `RuntimeOptions::namespace`. Not user-visible today because Stage 4a runs in the default namespace, but it left a configuration hole that would bite as soon as a non-default namespace is used.

## Test plan
- [x] New test `convoy_completion_command_targets_configured_provisioning_namespace` verifies completion targets the configured namespace and does not fall back to `"flotilla"`.

Closes #584.

### DIFF
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -495,7 +495,16 @@ pub struct InProcessDaemon {
     /// `NamespaceDelta` events arrive on the broadcast channel.  Consulted by
     /// `replay_since` so reconnecting clients can receive missed namespace state.
     namespace_state: Arc<RwLock<HashMap<String, flotilla_protocol::namespace::NamespaceSnapshot>>>,
+    /// Provisioning namespace used by daemon-side resource operations (e.g.
+    /// looking up the Convoy whose task is being marked complete). Set by the
+    /// daemon runtime at startup; defaults to [`DEFAULT_PROVISIONING_NAMESPACE`].
+    provisioning_namespace: RwLock<String>,
 }
+
+/// Default provisioning namespace used until [`InProcessDaemon::set_provisioning_namespace`]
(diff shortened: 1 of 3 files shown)


rjwittams:
Thanks for the careful read. On the getter doc — leaving it as-is.

The setter's doc carries WHY information (caller + timing), but a getter doc here would be pure WHAT ("returns the namespace"), which CLAUDE.md explicitly steers away from ("Don't explain WHAT the code does, since well-named identifiers already do that"). The existing `daemon_socket_path` setter/getter pair I'm mirroring also has no doc on either side, so the asymmetry is intentional and consistent with repo precedent.

PR #597, opened 2026-05-13
refactor(daemon): replay_since reads namespace state from projection directly

### LINKED ISSUE #591: refactor: replay_since should read namespace state from ConvoyProjection directly
From PR #590 review: the namespace_state mirror in InProcessDaemon (crates/flotilla-core/src/in_process.rs) is a broadcast-subscriber-driven cache of convoy state, consulted by DaemonHandle::replay_since. This has correctness seams: on broadcast lag, the mirror can hold stale data; if a delta arrives without a prior snapshot (shouldn't happen but can under lag/race), synthesizing baseline risks client regression.

PR #590 patched the symptoms (clear on lag; skip delta-before-snapshot). The proper fix is structural: replay_since should consult ConvoyProjection's own in-memory view directly rather than relying on a mirror. Needs the projection to expose a borrow-compatible read-only accessor, or restructure so the projection owns the authoritative state for both live broadcast *and* replay.

Scope: affects only namespace-scoped replay; repo/host streams have their own authority structure.

Mentioned in the claude[bot] review at https://github.com/flotilla-org/flotilla/pull/590#issuecomment-4313877641 (and earlier rounds).

rjwittams:
## Summary

Previously, `InProcessDaemon` maintained a `namespace_state` mirror populated by a background subscriber on the broadcast channel, and `replay_since` consulted that cache. The mirror had two correctness seams:

- On broadcast lag we cleared the entire cache (no namespace state in replay until the next snapshot arrives).
- A delta arriving before any snapshot for a namespace had to be discarded to avoid synthesising a seq-0 baseline that would regress reconnecting clients.

This PR makes `ConvoyProjection` the sole owner of authoritative namespace state. A small new type — `NamespaceProjectionState` in `flotilla-core` — wraps an `Arc<RwLock<HashMap<String, NamespaceView>>>`, is cloned cheaply, and is shared between the projection (single writer) and `InProcessDaemon::replay_since` (reader). The broadcast-driven mirror, its subscriber task, and both workarounds disappear with it.

Wiring follows the existing `set_daemon_socket_path` / `set_provisioning_namespace` precedent: the daemon runtime constructs the shared state in `start_with_options` and installs it on the daemon before the projection is spawned.

Closes #591.

## Changes

- `flotilla-core/src/namespace_projection.rs` (new) — `NamespaceView` and `NamespaceProjectionState`.
- `InProcessDaemon`:
  - Drop `namespace_state: Arc<RwLock<HashMap<…, NamespaceSnapshot>>>` and the spawned broadcast subscriber.
  - Add `set_namespace_projection_state` / `namespace_projection_state` accessors.
  - `replay_since` reads directly from the projection state.
- `ConvoyProjection`:
  - Take `NamespaceProjectionState` at construction, read/write through it.
  - `apply_convoy_event` and `apply_presentation_event` lock the shared state instead of a private `HashMap`.
- `DaemonRuntime::start_with_options` creates the shared state once and installs it on the daemon before spawning the projection.

## Test plan

- [x] Replace `namespace_delta_before_snapshot_is_skipped` and `namespace_state_cleared_on_broadcast_lag` (which probed the mirror's failure modes) with two tests against the new direct-read path:
  - `replay_since_reads_namespace_snapshots_from_projection_state` — writes go into the shared state, `replay_since` synthesises a `NamespaceSnapshot` with the expected `seq` and convoys.
  - `replay_since_skips_namespace_when_last_seen_matches_seq` — client at the current `seq` does not receive a redundant snapshot.

### DIFF
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -41,6 +41,7 @@ use crate::{
     },
     host_registry::HostCounts,
     model::{provider_names_from_registry, repo_name, RepoModel},
+    namespace_projection::NamespaceProjectionState,
     path_context::{DaemonHostPath, ExecutionEnvironmentPath},
     providers::{
         discovery::{discover_providers, run_host_detectors, DiscoveryResult, DiscoveryRuntime, EnvironmentAssertion, EnvironmentBag},
@@ -490,11 +491,11 @@ pub struct InProcessDaemon {
     /// Used to inject FLOTILLA_DAEMON_SOCKET into managed terminal sessions.
     daemon_socket_path: RwLock<Option<PathBuf>>,
     resource_backend: ResourceBackend,
-    /// Retained namespace snapshots, keyed by namespace name.
-    /// Updated by a background subscriber whenever `NamespaceSnapshot` or
-    /// `NamespaceDelta` events arrive on the broadcast channel.  Consulted by
-    /// `replay_since` so reconnecting clients can receive missed namespace state.
-    namespace_state: Arc<RwLock<HashMap<String, flotilla_protocol::namespace::NamespaceSnapshot>>>,
+    /// Shared with `ConvoyProjection` (in flotilla-daemon): the projection is the
+    /// sole writer, the daemon reads here for `replay_since`.  Replaces an earlier
+    /// broadcast-driven mirror that had correctness seams under lag and
+    /// delta-before-snapshot races.  Set by the daemon runtime at startup.
+    namespace_projection_state: RwLock<NamespaceProjectionState>,
     /// Provisioning namespace used by daemon-side resource operations (e.g.
     /// looking up the Convoy whose task is being marked complete). Set by the
     /// daemon runtime at startup; defaults to [`DEFAULT_PROVISIONING_NAMESPACE`].
@@ -626,7 +627,7 @@ impl InProcessDaemon {
             agent_state_store,
             daemon_socket_path: RwLock::new(None),
             resource_backend,
-            namespace_state: Arc::new(RwLock::new(HashMap::new())),
+            namespace_projection_state: RwLock::new(NamespaceProjectionState::new()),
             provisioning_namespace: RwLock::new(DEFAULT_PROVISIONING_NAMESPACE.to_string()),
         });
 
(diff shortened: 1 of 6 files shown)


rjwittams:
Thanks for the careful review.

**1. `emitted_initial_snapshot` leak (accepted).** Good layering point. Moved to a private `HashMap<String, bool>` on `ConvoyProjection`; `NamespaceView` in the core crate now only holds `convoys` + `seq` — what the shared contract actually needs. 8ae2a96c.

**2. `or_default()` (accepted).** Done in the same commit.

**3. Double `RwLock` on `namespace_projection_state` (kept).** Matches the existing `set_daemon_socket_path` / `set_provisioning_namespace` precedent on `InProcessDaemon`. Agreed `OnceCell` would model "set-once at startup" more precisely — happy to revisit those accessors together as substrate work later.

rjwittams:
Thanks. Three of the four taken:

**Two-step flag write (accepted).** Single read + single write now. d74f5d8f.

**`==` not `>=` (accepted).** Added a comment explaining the rationale — a client whose `last_seen` is ahead of the daemon (e.g. after a restart that resets in-memory seq) still receives a fresh snapshot.

**Test coverage (accepted).** Added:
- `replay_since_emits_only_stale_namespace_when_one_is_current` — multi-namespace replay with one current, one stale.
- `replay_since_resends_snapshot_when_client_seq_is_ahead` — regression guard for the `==` behaviour.

**Clone-then-lock comment (declined).** The pattern is `read().await.clone()` to release the outer lock before acquiring the inner; with names like `projection_state` and `all_snapshots()` carrying the intent, a comment would be explaining WHAT rather than a non-obvious WHY. CLAUDE.md's comment policy. Happy to revisit if the next reader trips over it.

rjwittams:
Thanks for the thorough pass.

**`HashSet<String>` over `HashMap<String, bool>` (accepted).** Done in 079c1cf3 — `contains()` reads cleaner.

**Brief comment on split read/write across the lock boundary (declined).** The safety here flows from `&mut self` on `apply_convoy_event`: no other caller can touch `self.emitted_initial_snapshot` (or anything else on the projection) while this method runs, even across the `await` on `self.state.write()`. Rust's borrow checker enforces this at the signature; a comment would be re-explaining the language. Happy to revisit if `apply_convoy_event` ever stops taking `&mut self`.

**Other observations (read accessor asymmetry, double-lock, Arc round-trip in `spawn_controller_loops`, RwLock<Option>, test setup helper):** all flagged as not requiring change. Agreed — leaving as-is for consistency with the existing `set_daemon_socket_path` / `set_provisioning_namespace` precedent. Worth revisiting together as a small substrate pass later.

rjwittams:
Thanks. Two of the four taken:

**1. Dead local in `start_with_options` (accepted).** Threaded `NamespaceProjectionState` through `spawn_controller_loops` as a parameter; the projection task no longer re-fetches it via the daemon getter, and the daemon `Arc` is no longer cloned just to reach the state. 5b36d8e7.

**2. Unreachable `else { return }` in `apply_presentation_event` (accepted).** Converted to `expect("convoy present; write lock held continuously from the cloned read above")` — same pattern as the neighbouring `get_mut`. Same commit.

**3 + 4 (no change).** State-without-writer when `start_controllers = false` is exactly the old behaviour and not a bug. Test setup helper is a fair quality note but, as you noted, consistent with the surrounding file — would be a follow-up cleanup rather than this PR.

rjwittams:
Added `#[derive(Debug)]` on `NamespaceProjectionState` in 2d340fd2. Enqueueing for merge.

PR #598, opened 2026-05-14
feat(daemon): bounded restart supervision for provisioning controllers

### LINKED ISSUE #583: Supervise provisioning controllers after reconcile-path errors
From PR #582 review: https://github.com/flotilla-org/flotilla/pull/582#issuecomment-4248236077

spawn_controller_loops now survives watch-stream disconnects inside ControllerLoop, but the outer daemon task still exits permanently if ControllerLoop::run() returns Err from reconcile-path work such as fetch_dependencies, apply_actuation, or status patching.

We should wrap each spawned provisioning controller in bounded restart supervision with backoff so transient backend/runtime failures do not silently disable provisioning until daemon restart.

Scope:
- add outer-loop restart/backoff around each controller task in crates/flotilla-daemon/src/runtime.rs
- fail loudly or surface metrics/logging when the retry budget is exhausted
- add a focused runtime test covering reconcile-path error recovery

rjwittams:
## Summary

- Wrap each `ControllerLoop` spawn in a new `supervise()` helper with bounded restart + exponential backoff
- Add `ControllerSupervision` config exposed via `RuntimeOptions::controller_supervision` (defaults: 10 consecutive failures, 100ms → 30s backoff, 60s reset window)
- Log at `error` level when the restart budget is exhausted

## Why

PR #582 made `ControllerLoop`'s internal watch loop survive watch-stream disconnects, but reconcile-path errors (`fetch_dependencies`, `apply_actuation`, status patches) still bubble out of `run()` and end the daemon's `tokio::spawn` task. Result: a single transient backend failure silently disabled provisioning until daemon restart.

The supervisor wraps each of the 7 controller spawn sites (Environment, Clone, Checkout, TerminalSession, TaskWorkspace, Presentation, Convoy). The 8th spawn (`ConvoyProjection`) has a different return type and is out of scope.

A run that survives `success_reset_after` resets the consecutive-failure counter — a controller that runs cleanly for a minute and then trips on a new error gets the full retry budget again, instead of accumulating toward exhaustion across the controller's lifetime.

## Test plan

- [x] 4 new unit tests covering supervise contract: clean-Ok, retry-until-success, budget exhaustion, reset-after-long-run
- [x] Existing `in_memory_stage4a_flow_reaches_running_and_completes_convoy` still passes (verifies wiring doesn't break the happy path)

Closes #583.

### DIFF
--- a/crates/flotilla-daemon/src/lib.rs
+++ b/crates/flotilla-daemon/src/lib.rs
@@ -5,3 +5,4 @@ pub mod cli;
 pub mod peer;
 pub mod runtime;
 pub mod server;
+pub mod supervisor;
--- a/crates/flotilla-daemon/src/runtime.rs
+++ b/crates/flotilla-daemon/src/runtime.rs
@@ -34,9 +34,12 @@ use flotilla_resources::{
 };
 use serde_json::json;
 use tokio::{sync::Mutex, task::JoinHandle};
-use tracing::{error, warn};
+use tracing::warn;
 
-use crate::ConvoyProjection;
+use crate::{
+    supervisor::{supervise, ControllerSupervision},
+    ConvoyProjection,
+};
 
 const DEFAULT_DOCKER_IMAGE: &str = "ubuntu:24.04";
 const DEFAULT_REPO_DIR_SUFFIX: &str = "dev/flotilla-repos";
@@ -46,6 +49,7 @@ pub struct RuntimeOptions {
     pub namespace: String,
     pub heartbeat_interval: Duration,
     pub controller_resync_interval: Duration,
+    pub controller_supervision: ControllerSupervision,
     pub start_controllers: bool,
 }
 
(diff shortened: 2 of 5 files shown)


rjwittams:
Addressed in 824b1a5e:

**1. Off-by-one** — fixed by changing `>` to `>=`. With default `max_consecutive_failures = 10`, exhaustion now happens on the 10th consecutive failure, matching the field name. Tests updated.

**2. Module doc block** — trimmed to a single line capturing only the non-obvious WHY (reconcile-path errors bubble out and kill the task), per CLAUDE.md's "no multi-paragraph docstrings" rule.

**3. `bon::Builder` derive** — declining. The premise "all other structs in the codebase with >3 fields use bon::Builder" doesn't hold: sibling `RuntimeOptions` in the same `runtime.rs` module has 4 fields and uses a plain struct literal too. Adding Builder to `ControllerSupervision` alone would split the local convention; aligning both is scope creep for a supervision PR. Both are flat config structs with a `Default` impl and no optional fields — the CLAUDE.md examples (`InputMeta`, `ControllerObjectMeta`, deep spec types) all describe heavier types where Builder pulls real weight.

PR #599, opened 2026-05-17
feat(cli): convoy create + workflow-template apply

rjwittams:
## Summary

First user-facing way to launch a convoy.

- `flotilla convoy <name> create --template <ref> [--input k=v]... [--repo URL] [--ref BRANCH]`
- `flotilla workflow-template <name> apply --file <path.yaml>` (create-or-update semantics)
- `scratch` WorkflowTemplate seeded at daemon startup: 1 task, 1 tool process that opens a shell labelled with the convoy name + the `topic` input. Demonstrates the full loop without needing the user to first build a template.

## Why

Until now, convoys could only be created programmatically — tests construct them via the resource backend directly. The TUI's convoys page literally renders "Create one via 'flotilla convoy create ...' (coming soon)". This PR makes that real for the limited tool-only path.

## Wire layout

- **Protocol** (`flotilla-protocol`): new `CommandAction::ConvoyCreate` and `CommandAction::WorkflowTemplateApply`, matching `CommandValue::{ConvoyCreated, WorkflowTemplateApplied}`; existing roundtrip tests extended.
- **CLI parsing** (`flotilla-commands`): `ConvoyVerb::Create` parses `--input k=v` via custom value parser; new `WorkflowTemplateNoun` reads the YAML file at resolve time.
- **Daemon** (`InProcessDaemon::execute`): both actions handled inline against `ResourceBackend`. `WorkflowTemplateApply` parses YAML (`serde_yml`), runs `workflow_template::validate`, then does `get → update` or falls back to `create`. `ConvoyCreate` builds `ConvoySpec` from the protocol fields.
- **Runtime**: `register_startup_resources` now seeds `default_workflow_templates()` idempotently.

## Snapshot deltas

Only column-width shifts in `command_palette_open` / `command_palette_selection` — the new `workflow-template` noun widens the longest-name column. The visible list itself is unchanged in content.

## Not in this cut

- `--wait` to stream convoy phase transitions
- Multi-template seeding (one is enough for first cut)
- Agent processes (still gated at Stage 4b in the reconciler)
- cwd auto-detection of repo (intentionally explicit until Project resource exists)

## Test plan

- [x] `cargo test --workspace --locked` — including 4 new integration tests in `crates/flotilla-daemon/tests/convoy_create_cli.rs`:
  - seeded `scratch` template appears after daemon startup
  - `ConvoyCreate` action creates a Convoy resource
  - `WorkflowTemplateApply` creates, then updates a custom template
  - invalid YAML produces a `CommandValue::Error`
- [ ] Smoke test the CLI against a real daemon

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1221,6 +1221,7 @@ version = "0.1.0"
 dependencies = [
  "clap",
  "flotilla-protocol",
+ "tempfile",
 ]
 
 [[package]]
--- a/crates/flotilla-commands/Cargo.toml
+++ b/crates/flotilla-commands/Cargo.toml
@@ -7,3 +7,6 @@ license.workspace = true
 [dependencies]
 flotilla-protocol = { path = "../flotilla-protocol" }
 clap = { version = "4", features = ["derive", "string"] }
+
+[dev-dependencies]
+tempfile = "3"
--- a/crates/flotilla-commands/src/commands/convoy.rs
+++ b/crates/flotilla-commands/src/commands/convoy.rs
@@ -20,6 +20,29 @@ pub struct ConvoyNoun {
 pub enum ConvoyVerb {
     /// Manage convoy tasks
     Task(ConvoyTaskNoun),
+    /// Create a convoy from a workflow template
+    Create {
+        /// Workflow template to instantiate
+        #[arg(long)]
+        template: String,
+        /// Input value (repeatable): --input key=value
+        #[arg(long = "input", value_parser = parse_input_kv)]
+        inputs: Vec<(String, String)>,
+        /// Repository URL the workflow operates on
+        #[arg(long = "repo")]
+        repository_url: Option<String>,
+        /// Git ref (branch/tag/commit) within the repository
+        #[arg(long = "ref")]
+        r#ref: Option<String>,
+    },
+}
+
+fn parse_input_kv(raw: &str) -> Result<(String, String), String> {
+    let (key, value) = raw.split_once('=').ok_or_else(|| format!("input must be key=value: {raw}"))?;
+    if key.is_empty() {
+        return Err(format!("input key cannot be empty: {raw}"));
+    }
+    Ok((key.to_string(), value.to_string()))
 }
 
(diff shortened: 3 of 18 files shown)


rjwittams:
Addressed in 35af439e:

**1. Display impls don't shell-quote** — fixed. New `flotilla_commands::quote_value` helper (matches the palette tokenizer's quoting rules). Applied to `--template`, `--input`, `--repo`, `--ref`, and `--file`. New tests assert quoting fires for whitespace-bearing values. Left the existing `AgentNoun::Display` etc. alone — same issue applies to most nouns and is worth a sweep, but out of scope for this PR.

**2. `.is_ok()` swallows non-NotFound errors** — fixed in `ensure_default_workflow_templates`, plus the same pattern in sibling helpers `ensure_host_exists` and `ensure_host_direct_environment_exists` (which had the same footgun pre-existing). Now uses explicit `Err(ResourceError::NotFound { .. })` matching with a targeted "check X: {err}" error for anything else.

**3. `{e:?}` vs `{e}`** — `ValidationError` had no `Display` impl. Added one (plus for `InterpolationLocation`, `InterpolationField`) and switched to `{e}`. Errors now read like "duplicate task name \`work\`" instead of `DuplicateTaskName { name: "work" }`.

**4. `empty_identity` duplication** — extracted `empty_repo_identity()`; replaced 4 inline literals.

**5. TOCTOU** — acknowledged. Fine for interactive CLI use; if the apply path ever becomes a hot path for concurrent automation we'd want a CAS-style update or proper apply semantics on the backend.

**6. No `workflow_ref` existence check** — filed as #600. Agree it'd be friendlier; the reconciler does eventually surface this but a command-time check is a quick win.

**7. Missing round-trip test for `workflow-template apply`** — added `round_trip_apply` and `apply_display_quotes_path_with_whitespace`.

rjwittams:
Filed the round-trip-via-tokenizer note as #601 — needs moving the palette tokenizer into `flotilla-commands` to be shared with `assert_round_trip`, so out of scope here.

PR #603, opened 2026-05-22
feat(resources,cli): add Project resource + project_ref on ConvoySpec

rjwittams:
## Summary

New \`Project\` resource for grouping convoys around shared identity — one or more repositories with optional monorepo subpath and default branch. Sits alongside the existing TrackedRepo concept (collapsing them is deferred until Yeoman/Governor surface a clearer driver).

\`ConvoySpec\` gains a \`project_ref: Option<String>\` field. **Metadata only in v1** — the reconciler doesn't consult it. Future work: substitute the project's repository/ref when the convoy doesn't supply one.

## Resource shape

\`\`\`rust
pub struct ProjectSpec {
    pub display_name: Option<String>,
    pub repositories: Vec<ProjectRepositorySpec>,
}

pub struct ProjectRepositorySpec {
    pub repo: String,                  // URL
    pub subpath: Option<String>,       // monorepo slice
    pub default_branch: Option<String>,
}
\`\`\`

\`Vec<repo, subpath>\` from the start, per the design discussion — supports cross-project (multiple repos as one logical unit) and monorepo-slice (one repo split into multiple projects).

## CLI

- \`flotilla project <name> create [--display-name "…"] [--repo URL [--subpath …] [--ref …]]\` — single-repo convenience.
- \`flotilla project <name> apply --file <yaml>\` — full multi-repo form.
- \`flotilla convoy <name> create … --project <project>\` — tags the convoy with a project grouping.

## Wire layout

- **Protocol**: new \`CommandAction::ProjectCreate\` / \`ProjectApply\` + matching \`CommandValue\` shapes; \`ConvoyCreate\` gains \`project_ref\`. Roundtrip cases extended.
- **Daemon**: \`InProcessDaemon\` handles both actions inline against \`ResourceBackend\`. \`ProjectApply\` parses YAML (\`serde_yml\`) and does \`get → update\` or \`create\`.
- **CLI**: new \`ProjectNoun\` in \`flotilla-commands\`; wired into \`NounCommand\` AND \`SubCommand\` in \`src/main.rs\` (avoiding #602's omission).

## \`Message\` enum allow

\`#[allow(clippy::large_enum_variant)]\` on \`Message\` — same allow already on \`Response\`. Adding two more action variants to \`CommandAction\` pushed the Request variant past clippy's threshold; boxing every variant clutters the wire shape for no real benefit.

## Test plan

- [x] \`cargo +nightly-2026-03-12 fmt --check\`
- [x] \`cargo clippy --workspace --all-targets --locked -- -D warnings\`
- [x] \`cargo test --workspace --locked\` — including 5 new integration tests in \`crates/flotilla-daemon/tests/project_cli.rs\`:
  - \`project_create_command_creates_project_resource\` — single-repo create
  - \`project_apply_supports_multi_repo\` — YAML with two entries (one with subpath)
  - \`project_apply_updates_existing\` — apply-as-update semantics
  - \`convoy_create_carries_project_ref\` — \`--project\` flows through to ConvoySpec
  - \`project_apply_rejects_invalid_yaml\` — clear error path
- [x] Top-level CLI regression test \`cli_parses_project_noun\` in \`src/main.rs\` (mirrors #602's fix)
- [ ] Smoke against a real daemon

## Not in this cut

- Reconciler doesn't read \`project_ref\` yet
- No TUI surface
- No auto-creation of Projects from existing TrackedRepos
- No multi-repo-aware convoy expansion

### DIFF
--- a/crates/flotilla-commands/src/commands/convoy.rs
+++ b/crates/flotilla-commands/src/commands/convoy.rs
@@ -35,6 +35,9 @@ pub enum ConvoyVerb {
         /// Git ref (branch/tag/commit) within the repository
         #[arg(long = "ref")]
         r#ref: Option<String>,
+        /// Project this convoy belongs to (metadata grouping)
+        #[arg(long = "project")]
+        project_ref: Option<String>,
     },
 }
 
@@ -80,12 +83,19 @@ impl ConvoyNoun {
                     host: HostResolution::Local,
                 }),
             },
-            ConvoyVerb::Create { template, inputs, repository_url, r#ref } => Ok(Resolved::NeedsContext {
+            ConvoyVerb::Create { template, inputs, repository_url, r#ref, project_ref } => Ok(Resolved::NeedsContext {
                 command: Command {
                     node_id: None,
                     provisioning_target: None,
                     context_repo: None,
-                    action: CommandAction::ConvoyCreate { name: self.subject, workflow_ref: template, inputs, repository_url, r#ref },
+                    action: CommandAction::ConvoyCreate {
+                        name: self.subject,
+                        workflow_ref: template,
+                        inputs,
+                        repository_url,
+                        r#ref,
+                        project_ref,
+                    },
                 },
                 repo: RepoContext::None,
                 host: HostResolution::Local,
@@ -109,7 +119,7 @@ impl std::fmt::Display for ConvoyNoun {
                     }
                 }
             }
-            ConvoyVerb::Create { template, inputs, repository_url, r#ref } => {
+            ConvoyVerb::Create { template, inputs, repository_url, r#ref, project_ref } => {
                 write!(f, " create --template {}", quote_value(template))?;
                 for (k, v) in inputs {
                     write!(f, " --input {}", quote_value(&format!("{k}={v}")))?;
@@ -119,6 +129,9 @@ impl std::fmt::Display for ConvoyNoun {
                 }
                 if let Some(reference) = r#ref {
                     write!(f, " --ref {}", quote_value(reference))?;
+                }
+                if let Some(project) = project_ref {
+                    write!(f, " --project {}", quote_value(project))?;
                 }
             }
         }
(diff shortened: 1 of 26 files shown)


rjwittams:
Addressed in ffc3ac42:

**1. `--subpath` / `--ref` silently dropped without `--repo`** — fixed. clap `requires = "repository_url"` on both args now produces a clear "the following required arguments were not provided: --repo" error at parse time. Two unit tests (`subpath_without_repo_is_rejected`, `ref_without_repo_is_rejected`) assert the rejection.

**2. TOCTOU in `ProjectApply`** — acknowledged. Same `get → update/create` pattern as `WorkflowTemplateApply`; not a regression and fine for interactive CLI use. When concurrent automation or a real persistence backend lands we'll want CAS-style updates or proper backend-side apply semantics — flagged but not a v1 blocker.

**3. Test input for invalid YAML** — added `project_apply_rejects_wrong_shape_yaml` covering the serde deserialisation error path (valid YAML, `repositories` entry missing the required `repo` field). Complements the existing syntax-error case.

**4. `bon::Builder` on `ProjectSpec` / `ProjectRepositorySpec`** — declining. `ProjectSpec` has 2 fields and `ProjectRepositorySpec` has 3; CLAUDE.md sets the bar at "more than three fields, deep nesting, or many optional fields." Both are under the threshold and shallow. If they grow we can add Builder then.

rjwittams:
Picked up (1) and (2) in 0a669f71:

- Added a one-line comment on `round_trip_apply` clarifying that the path doesn't need to exist (only the parse↔Display cycle is exercised).
- Doc comment on `ProjectRepositorySpec::default_branch` notes it surfaces as `--ref` on the CLI.

(3) Boxing as a future option noted — agree it's a separate concern.

(4) `await_command_result` duplication acknowledged. Two-file duplication of a ~12-line helper isn't worth a shared utility right now; we can extract if a third integration test wants it or the timeout/loop shape has to change. Same call as the `start_daemon` helper which is also duplicated.

Issue #613, opened 2026-06-21
Dogfood a real convoy end-to-end


rjwittams:
## Parent
#604

## What to build
Prove the launched-work path on a real workflow: seed a WorkflowTemplate, then create → provision → attach → complete a Convoy, dogfooded by a human. Capture the gaps surfaced (provisioning, presentation, completion UX) as follow-up issues.

## Acceptance criteria
- [ ] At least one seeded WorkflowTemplate usable for real work
- [ ] Full lifecycle exercised by hand: create → provision → attach → complete
- [ ] Gaps captured as follow-up issues
- [ ] Works against the embedded (sqlite-backed) control plane

## Blocked by
- #612

Issue #612, opened 2026-06-21
Embed the control plane in the flotilla daemon (sqlite-backed)


rjwittams:
## Parent
#604

## What to build
Run the resource store + controller runtime in-process inside the flotilla daemon, backed by the embedded store, so Convoys and their reconcilers work with no external Kubernetes. End-to-end: daemon boots → control plane up → a Convoy can be created and reconciled locally.

## Acceptance criteria
- [ ] Daemon starts the embedded control plane as part of its lifecycle
- [ ] Convoy create → reconcile works against the embedded store (no k8s)
- [ ] Controller runtime supervised within the daemon
- [ ] Existing convoy projection / TUI convoy view works against the embedded store

## Blocked by
- #611

Issue #611, opened 2026-06-21
sqlite/libsql ResourceBackend


rjwittams:
## Parent
#604

## What to build
A durable, embedded `ResourceBackend` (sqlite or libsql/turso) behind the existing `TypedResolver`, so the control plane runs without a live Kubernetes cluster. Stores the canonical k8s object keyed by (group, version, kind, namespace, name) with the in-memory backend's exact version and watch semantics. The in-memory backend remains the contract; this backend must pass the same suite.

## Acceptance criteria
- [ ] New backend selectable behind `TypedResolver`, API-identical to other backends
- [ ] Passes the in-memory contract test suite (CRUD, optimistic concurrency, list, watch-from-version, namespace isolation)
- [ ] Survives process restart (durability verified)
- [ ] Monotonic resourceVersion preserved across restarts

## Blocked by
- #606

Issue #607, opened 2026-06-21
Design: observed-resource contribution + lifecycle authority


rjwittams:
## Parent
#604

## What to build
Design slice (HITL). Define how the **Fleet Observer** contributes *observed resources* into the shared resource store, and how **Lifecycle Authority** (Observed / Adopted / Managed, per ADR 0003) is represented and read.

Decisions to land:
- How providers emit observed resources (same kinds as desired; distinguished by label/field).
- How lifecycle authority is carried on a resource and how adoption promotes along the axis (never a kind change).
- How the **Aggregator** reads observed + desired together to produce views.
- The contract reconcilers must honour (never create/destroy observed/adopted).

## Acceptance criteria
- [ ] Written design (short spec or ADR addendum) covering the above
- [ ] Worked example: the user's hand-managed local clone as a permanently-Adopted resource
- [ ] Shape agreed for how one provider will emit observed resources (feeds the next slice)

## Blocked by
None - can start immediately

rjwittams:
Design resolved via a grilling session. Captured as ADRs 0004-0006 + CONTEXT updates in PR #621:
- ADR 0004: observed resources ephemeral + generational; lifecycle authority selects backing store.
- ADR 0005: main view = surface-agnostic composition of panels (tab=composition, panel=filtered resource view, convoy-row→Task DAG); fixed defaults now.
- ADR 0006: observed things split into resources vs reference data (PRs/issues are link-only reference data, not resources); convoy external results are referenced, not modeled; CloudAgent/Environment may be service-provided; reachability is orthogonal.

Downstream: #614 (tracer: Checkout) unchanged; #622 (panel view-model); #623 (reference-data + Aggregator linking); #616 re-scoped to migrate observed resource providers + delete the pipeline. Parked: #624 (agentic convoy), #625 (host/environment model).

PR #617, opened 2026-06-21
docs: transition roadmap, ADRs, glossary, and Plane-A freeze

rjwittams:
Captures the fleet-observer → control-plane transition worked out in a design session. Docs only — no code changes.

## What's here
- **`CONTEXT.md`** — glossary for the system and the product family (cleat / porthole / uishell).
- **`docs/adr/0001`** — k8s-isomorphic resource model; k8s is *a backend*, not the API contract (always keep a k8s representation; defer a k8s API edge).
- **`docs/adr/0002`** — multi-host is resource-store federation via the **Tender**, not peer-merge; the resource watch stream *is* the replication log.
- **`docs/adr/0003`** — observed / adopted / managed resources share one store; lifecycle authority is per-resource (the hand-managed local clone is permanently *adopted*).
- **`docs/roadmap.md`** — phases 0–6 and the two freezes; the sequencing rule (freeze the doomed plane now, extract generics only after the boundary is proven).
- **`CLAUDE.md`** — the **Plane-A freeze** note (the one enforceable artifact of the freeze: visible to every agent session).

## Tracking
Umbrella: #604. Critical path to a k8s-free dogfoodable convoy: #606 → #611 → #612 → #613; #607 unblocks the observer reshape in parallel.

Refs #604.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -6,6 +6,14 @@ A system for improving multi-agent developer workflows. Consists of a daemon mes
 
 We are in a **no backwards compatibility** phase. Protocol types, snapshot formats, config file formats, and wire formats can all change freely without migration logic or deprecation paths.
 
+### Plane-A freeze (transition in progress)
+
+Flotilla is mid-transition from a fleet *observer* (providers → correlation → WorkItems → snapshots → TUI) to a *control plane* (k8s-isomorphic resources → reconcilers → convoys). See `docs/roadmap.md`, `CONTEXT.md`, and `docs/adr/0001-0003`; the umbrella is flotilla-org/flotilla#604.
+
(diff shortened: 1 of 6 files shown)


rjwittams:
Thanks — all six addressed in the latest commit:

1. **'bridge' on the avoid list** — replaced with **Tender** throughout ADR 0002 (it predated the CONTEXT.md rename). Fixed.
2. **Hardcoded local paths** — dropped the `~/dev/...` paths from CONTEXT.md's External Collaborators; they're just named as sibling repositories now. Fixed.
3. **'Q1' ambiguity** — those were grilling-**question** numbers (Q1/Q6/Q7), not calendar quarters, and meaningless outside the session — so I removed them entirely rather than pinning a year. Fixed.
4. **ADR status/date** — added `Status: Accepted` / `Date: 2026-06-21` headers to all three ADRs. Fixed.
5. **Invalid `docs/adr/0001-0003` path** — now points at `docs/adr/` in CLAUDE.md. Fixed.
6. **Peer LOC undercount** — you're right; `peer/` is ~5,400 LOC including tests (and the `manager/` subdir I'd missed). Corrected in ADR 0002.

rjwittams:
Thanks for the follow-up. Both addressed:

1. **Punctuation** — added a colon after "stay relevant" (it was an artifact of removing the `(Q6)` ref). Fixed.
2. **A/B vs Plane A/B clash** — good catch on the confusion. I kept the A/B labels rather than dropping them, because **B** is referenced externally as "ADR 0001-B" (in `roadmap.md` phase 1 and issue #606), so dropping would dangle those refs. Instead I added a parenthetical noting the A/B labels are local to the ADR and unrelated to Plane A/Plane B. That removes the double-take while keeping the references intact.

Issue #604, opened 2026-06-21
tracking: Plane-A → control-plane transition roadmap


rjwittams:
Umbrella tracking issue for the Flotilla transition out of its two-plane straddle (fleet observer → control plane). Rewritten 2026-08-24; prior era preserved in edit history.

Source of truth in-repo: `/CONTEXT.md` (glossary) · `docs/adr/` (0001–0035) · `docs/roadmap.md` (phase table).

## The freezes (standing)
1. **Plane A is bugfix-only**; peer-merge is deletion-bound, gated below.
2. **TUI is factored, not frozen** (re-measure: #1543).

## Phase state

- **0–2 — complete.** Freeze, embeddable control plane (sqlite backend), convoy as the real launch path — all dogfooded daily.
- **3 — substantially complete.** Observed checkouts, lifecycle authority, Aggregator, observer-pipeline deletion (#1475) landed. Residuals: #616 (surviving ProviderData consumers — needs revalidation against the post-#1475 tree), #623 (reference-data linking — partially superseded by forge-tier providers; revalidate), #633 (M1 dogfood acceptance, HITL).
- **4 — in progress.** Store replication is live (`server/replicator.rs`, definitions-class federation proven by today's WorkflowTemplate work). **Peer-merge deletion is now gated on the tracked-root cull**: #1721 (audit ruled 2026-08-24, ADR 0031 step 2 discharged) → slices **#1767 → #1768 → #1769 → #1770**. After slice 4, the peer deletion contract gets written against a re-homed codebase. Tender extraction: #608, plus a new charter requirement from cleat#198 (session channels as a first-class primitive); design now on wayfinder map #1861 (provisional design `docs/architecture/tender.md`, research draft #1878).
- **5 — pending.** #609 (control-plane core extraction), after 4 proves the boundary.
- **6 — in progress.** #610 (shared view-model / surfaces), #1543. The wheelhouse↔cleat render-output surface (cleat#198) is this phase's native-app front: the provider C API + packet protocol exist; remote routing is the open half.

## Active fronts (crewed or freshly ruled)

- **#1767** — host-tier registry extraction — cull slice 1 of 4; PR #1771 open/green. Slices #1768→#1769→#1770 queued behind it.
- **#1766** — skills provisioning for vessels — PR #1772 open/green; cleat#199 (skill sync) also crewed.
[...]

Issue #518, opened 2026-03-26
Ensure parent directories exist before writing state files


rjwittams:
## Problem

Shpool terminal provider constructs paths like `state_dir/shpool/shpool.socket` and `state_dir/shpool/config.toml` but doesn't ensure the `shpool/` subdirectory exists before use. This causes shpool to fail with:

```
Error: registering config file for watching
Caused by:
    0: best_effort_watch on .local/state/flotilla/shpool/config.toml
    1: adding notify watch for config file
```

## Analysis

Most subsystems handle this correctly — tmux, zellij, agents, SSH peers, and config store all call `create_dir_all()` before writing. The shpool provider is the main gap:

- `write_config()` in shpool.rs correctly creates parent dirs before writing config
- But `start_daemon()` does NOT create parent dirs before:
  - The socket path (`shpool/shpool.socket`)
  - The log file (`shpool/daemonized-shpool.log`)
  - The PID file (`shpool/daemonized-shpool.pid`)

## Fix

Add `create_dir_all` for the shpool subdirectory in `start_daemon()` before any file operations. Since most other providers already follow this pattern, the fix is localised to the shpool provider.

**File:** `crates/flotilla-core/src/providers/terminal/shpool.rs` — `start_daemon()` function.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #621, opened 2026-06-21
docs: observer-reshape design (ADRs 0004-0006) for #607

rjwittams:
Resolves the #607 design (observed-resource contribution + lifecycle authority + what becomes of the main views). Docs only.

## Decisions
- **ADR 0004** — observed resources flow through the store interface but are **ephemeral + generational**; lifecycle authority selects the backing store (Observed/Adopted → ephemeral, Managed → durable). A daemon restart starts a new generation.
- **ADR 0005** — the main view is a **surface-agnostic composition of panels** (tab = composition; panel = filtered resource view at a scope; row = resource; convoy-row drills to its Task DAG). Fixed defaults now, composable later (#518).
- **ADR 0006** — observed things split into **resources** (lifecycle-owned: Checkout, CloudAgent, TerminalSession, Environment, Host — some service-provided) vs **reference data** (link-only external records: ChangeRequest, Issue). A convoy's **external result** (PR, CMS article, video) is referenced, not modeled. Reachability (Tender vs external API) is orthogonal to this cut.

## CONTEXT
Adds Generation, CloudAgent, Reference Data, External Result; refines Environment (container; host/env relationship flagged open) and Aggregator.

## Follow-ups (filed separately)
- Reference-data layer + Aggregator linking; dashboard panel view-model.
- Parked Brainstorms: agentic/dynamic convoy launch; host/environment/location model.

Refs #604, #607.

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -41,6 +41,26 @@ A running program within a **Task**, resolved to a placement by selectors
 (e.g. `capability: code`) rather than by template variable substitution.
 _Avoid_: Pane, terminal (those are presentation concerns).
 
+**CloudAgent**:
+A coding agent whose harness loop runs remotely / under management (Claude Code,
+Codex, Cursor as a managed session), as opposed to an agent running locally in a
+terminal. A **resource** with a lifecycle, typically **service-provided**
+(claude.ai, cursor) and reached via the service's API rather than the **Tender**.
+_Avoid_: Agent (ambiguous — reserve for the abstract notion), bot, session.
+
+**Reference Data**:
+Service-scoped external records that Flotilla links to but does not own the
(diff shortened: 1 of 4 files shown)


rjwittams:
Thanks — all three addressed in b80782eb:
1. Added `**Status:** Accepted` / `**Date:**` headers to ADRs 0004-0006 (confirmed 0001-0003 carry them, so matching the convention).
2. CONTEXT `CloudAgent` entry now notes `ManagedAgent` is used interchangeably, with CloudAgent as canonical and the distinction not yet pinned.
3. ADR 0004's `Observed / Adopted` bullet now forward-references the deferred note so the Adopted treatment isn't confusing on first read.

PR #639, opened 2026-07-06
[codex] Resolve attach through recursive next hop

### LINKED ISSUE #637: Make flotilla attach resolve one recursive next hop
## Summary

Follow up from #631. The first store-backed attach implementation can return a fully wrapped hop-chain command. The desired next shape is for `flotilla attach <session>` to resolve only the next hop locally, then invoke `flotilla attach <session>` on that host, recursively, until the host that owns the terminal session runs the local terminal attach command.

This is narrower than the broader transport-agnostic attach vision in #368.

## Desired behavior

- User runs `flotilla attach <ref>` on any host that can see the session reference.
- The local daemon resolves whether the session is local or remote.
- If local: attach directly to the terminal provider session.
- If remote: run a single next-hop command such as `ssh -t <next-hop> flotilla attach <same-ref>`.
- Each hop repeats the same local resolution, so command construction does not need deeply nested SSH/shell quoting.

## Why

- Keeps resolution local to the host that has the relevant resource/provider view.
- Avoids centralizing all route knowledge at the original caller.
- Avoids complex nested quoting for multi-hop attach chains.
- Gives us a cleaner stepping stone toward #368.

## Acceptance criteria

- Remote attach composes one hop at a time instead of returning a fully flattened hop-chain command.
- Local attach still uses the terminal provider attach args.
- Errors clearly distinguish unknown reference, ambiguous reference, unreachable next hop, and terminal-provider attach failures.
- Tests cover at least local attach, one-hop remote attach, and a multi-hop route shape without asserting brittle nested shell quoting.

rjwittams:
## Summary

Closes #637.

- Change store-backed attach so remote sessions resolve to a single recursive next-hop command instead of a flattened provider attach chain.
- Use topology routes to choose the immediate next-hop host when the target host is reachable transitively.
- Keep local attach using terminal-provider attach args, preserving the existing local Docker environment wrapping path.
- Add tests for local attach, one-hop remote attach, routed multi-hop shape, and unreachable next-hop errors.

## Tests

### DIFF
--- a/crates/flotilla-core/src/hop_chain/remote.rs
+++ b/crates/flotilla-core/src/hop_chain/remote.rs
@@ -81,6 +81,25 @@ impl SshRemoteHopResolver {
         args.push(Arg::Quoted(info.target.clone()));
         args
     }
+
+    /// Build a single SSH hop that runs `command` on `host`.
+    ///
+    /// Unlike `resolve_wrap`, this intentionally does not inspect or wrap an
+    /// already-resolved inner action. Store-backed recursive attach uses this
+    /// to emit one next-hop command and lets the next daemon resolve the
+    /// following hop locally. The command still runs through a login shell so
+    /// user-installed tools such as `flotilla` under `~/.cargo/bin` are found.
+    pub fn one_hop_command_args(&self, host: &HostName, command: Vec<Arg>) -> Result<Vec<Arg>, String> {
+        let info = self.ssh_info(host)?;
+        let mut ssh_args = self.ssh_prefix_args(&info);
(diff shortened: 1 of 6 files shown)


rjwittams:
Thanks. The blocking review items are fixed in fd34f1a7: recursive SSH attach now uses the login-shell wrapper, topology fallbacks are exposed most-recently-learned first using Reverse(learned_epoch), and the defensive host-registry error paths have in_process coverage. I’m leaving the disconnected-primary end-to-end attach case out of this PR since the ordering behavior is covered at the PeerManager unit level and the re-review marks it non-blocking.

PR #656, opened 2026-07-07
fix: resolve attach from fleet replicas

### LINKED ISSUE #654: attach resolution ignores fleet replicas — cross-host attach fails despite ls visibility
Found in T3 (#633). From kiwi, `flotilla ls` shows `day2 … running … feta` (replica-cache path), but `flotilla attach day2` fails with `no attach target matching 'day2'`.

Cause: `resolve_attach_command_internal` (in_process.rs ~2248) lists `TerminalSession` resources from the **local durable store only** — remote sessions exist only in the replica cache, which resolution never consults. The recursive next-hop machinery (#637) can't kick in because resolution never discovers a remote owner.

Fix shape: resolution should consult local store **plus fleet replicas** (same data `ls` renders); a replica match resolves to the next-hop command (`ssh -t <host> flotilla attach <ref>`) per #637. Staleness applies: a stale replica match should still forward (the owning host re-resolves authoritatively and errors honestly if gone).

Minor UX alongside: a not-found currently prints a raw eyre error + backtrace pointer (src/main.rs:400) — should be a plain one-line error.

Workaround meanwhile: `ssh -t feta flotilla attach day2`.

rjwittams:
## Summary

- Teach attach resolution to consider running sessions from the fleet replica cache, not just local `ResourceTerminalSession` rows.
- Resolve replica matches as one recursive next-hop attach command so stale or currently unreachable cached rows still forward to the owning host.
- Keep replica attach visibility consistent with configured `hosts.toml` entries and local bare-role matching behavior.
- Print human attach errors as a plain one-line message instead of a color-eyre report.

## Root Cause

`flotilla ls` could show remote fleet replicas from another host, but `flotilla attach <ref>` only searched the local terminal-session store. That made visible cross-host sessions fail with `no attach target matching ...`.

## Validation

- `cargo test -p flotilla-core --locked attach_query_resolves_fleet_replica_session_as_one_recursive_hop -- --nocapture`
- `cargo test -p flotilla-core --locked attach_query_ignores_fleet_replica_hosts_that_are_not_configured -- --nocapture`
- `cargo test -p flotilla-core --locked attach_query -- --nocapture`
- `cargo test -p flotilla-core --locked`
- `git diff --check`
- `cargo clippy -p flotilla-core --all-targets --locked -- -D warnings`
- CI passed: Format, Clippy, Test, claude-review

Closes #654

### DIFF
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -255,6 +255,27 @@ fn attach_reference_label(session_name: &str, labels: &BTreeMap<String, String>)
     }
 }
 
+fn fleet_row_attach_reference_keys(row: &FleetListRow) -> Vec<String> {
+    let mut refs = vec![row.convoy.clone(), row.vessel.clone(), row.crew.clone()];
+    if row.crew != "-" {
+        refs.push(format!("{}/{}", row.convoy, row.crew));
+        if let Some((_task, role)) = row.crew.rsplit_once('/') {
+            refs.push(role.to_string());
+        }
+    }
+    refs.sort();
+    refs.dedup();
+    refs
+}
+
+fn fleet_row_attach_reference_label(row: &FleetListRow) -> String {
+    if row.crew == "-" {
+        format!("{} ({})", row.convoy, row.host)
+    } else {
+        format!("{}/{} ({})", row.convoy, row.crew, row.host)
+    }
+}
+
 fn session_status_label(phase: Option<ResourceTerminalSessionPhase>) -> String {
     match phase {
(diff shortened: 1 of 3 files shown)


rjwittams:
Fixed the two concrete polish points from this review: attach resolution now scopes replica candidates to hosts currently configured in hosts.toml, matching fleet list visibility, and replica rows now include the bare role key (for example coder from implement/coder) so cross-host attach matches local attach behavior. Added regression coverage for both cases. I left the broader ambiguity/filter coverage as-is since the attach-query suite already covers the generic ambiguity/running-state behavior and these new tests lock the replica-specific deltas.

PR #664, opened 2026-07-09
feat: replace convoy projection with aggregator panels

### LINKED ISSUE #622: Aggregator: generalize ConvoyProjection to emit the panel view-model; convoy view rides panels
## Parent
#604

## What to build
Per ADR 0005 and the Q4 contract decision (2026-07-03, deliberate 'bite the bullet'): generalize `ConvoyProjection` into the daemon-side **Aggregator** now — **one panel stream, no parallel period**. The view-model is forged against the hardest known consumer (the convoy view) from day one.

- Surface-agnostic **panel view-model** wire types in `flotilla-protocol`: tab = composition of panels; panel = (resource kind/query, scope, columns, intents); rows reference resources; hierarchical drill-down (convoy → Task DAG) and per-row intents are expressible. No rendering/styling types.
- Generalize `ConvoyProjection` → **Aggregator**: watches both stores (durable managed + ephemeral observed, honouring generation resets), joins, emits panel snapshots/deltas with reconnect/replay semantics at least equal to today's `replay_since`.
- The existing convoy view (list + task DAG detail + attach via 'a' + complete via palette) is re-expressed as a default panel composition and the TUI renders it **from the panel stream**.
- Delete `NamespaceSnapshot`/`NamespaceDelta` once the convoy view rides panels — they are replaced, not paralleled.

## Acceptance criteria
- [ ] Panel view-model types in flotilla-protocol, free of rendering deps
- [ ] Daemon-side Aggregator watches both stores and emits panel snapshots/deltas; reconnect/replay covered by tests
- [ ] Convoy TUI view rendered from the panel stream at feature parity (list, DAG drill-down, attach, complete)
- [ ] NamespaceSnapshot/NamespaceDelta removed
- [ ] TUI snapshot-test changes individually justified (parity, not regression)

## Blocked by
- #626

rjwittams:
## Summary

- add a rendering-free, generic panel view-model protocol with resource-backed rows, hierarchical DAGs, and vessel/leg intents
- replace the namespace stream with Aggregator panel snapshots and deltas across durable, observed, and complete fleet-replica cache state
- drive the convoy TUI from the panel stream at list, drill-down, attach, and completion parity, including explicit remote-host routing

## Why

The convoy tab used a local-only namespace projection while `flotilla ls` used the fleet replica cache, so the two views could disagree. This change gives surfaces one replica-aware panel stream and removes the parallel namespace wire contract.

## Validation

- all 50 TUI render snapshots pass unchanged
- final standards and #622 spec reviews report no blocking findings

## Human checkpoint

**Robert: convoy-view parity check required before this PR is marked ready or enqueued.** Please verify list, DAG drill-down, attach via `a`, complete via the palette, and fleet-visible remote convoys. This PR has not been enqueued.

Closes #622


rjwittams:
## Human parity check: NOT PASSED — and the parity bar itself was mis-specced

Robert drove the convoy tab on the live kiwi/feta fleet (both daemons on this branch). Verdict in two parts: the **backend is solid and verified**; the **TUI does not pass the gate**, partly through in-PR bugs and partly because the handoff defined parity against the wrong reference. Do not enqueue.

### What's verified good (keep)

- `NamespaceSnapshot`/`Delta` fully replaced, not paralleled; panel wire types rendering-free and speaking vessel/leg.
- Aggregator merges durable + observed (generation-aware) + fleet replica broadcast; panels and `ls` share the replica cache — the divergence complaint this PR set out to fix is fixed at the data layer.
- Live cross-host attach resolution through panel rows works (kiwi → `flotilla attach day4` → feta crew shell, session survives detach).
- Zero snapshot changes, CI green.

### Must-fix in this PR

1. **Failed convoys render as "initializing…".** `aggregator.rs:340` computes `initializing = workflow_snapshot.is_none()`, which stays true forever for a convoy that fails input validation (live specimen: `day1`, failed 'missing input topic'). The row *carries* `phase: failed` + the message, but `detail.rs:40` renders the initializing placeholder instead of them. The new view is less truthful than `ls` here. `initializing` must be false once the convoy is in a terminal phase, and the detail view should never let it mask phase/message.
2. **Focus trap in the task pane.** `ConvoyTasks` mode exits only via left-arrow/ESC (`binding_table.rs:157-158`); `h` is unbound, so the vim grammar that enters (`l`) can't leave. Bind `h` (and check the mode's exits generally match the rest of the TUI).
3. **`a` (AttachConvoyLeg) is a dead key in real environments.** It dispatches `SelectWorkspace` through the presentation-manager stack, and with no presentation manager configured (the normal case today, including the dogfood fleet) it can never attach. Either wire it to the attach path that actually works (the terminal-pool semantics the CLI `flotilla attach` and the repo views use) or remove the binding — don't ship a key that silently does nothing.
4. **Question, not necessarily a change:** the palette-completion e2e flipped `start_controllers: true → false` and hand-creates the `Presentation`. What coverage did that trade away, and why was it needed?

### Scope note: the parity gate stays closed after these fixes

The handoff (my error, not yours) defined parity as "feature parity with the old convoy tab". The old tab was never a real workflow. The meaningful bar is **parity with working with attachable sets in the repo views**: shared table widget, mouse behaviour, information density (vessel/crew/state/host/staleness), `.` action menu, and attach-that-attaches on a crew row. The panel wire model in this PR already carries columns/typed values/row intents — i.e. the data is shaped for exactly that rendering; the TUI here adapts it back into the old bespoke widgets instead. That work will be specced as its own slice against the repo-view workflow (design question pending: better convoys tab vs vessels surfacing in the repo views themselves). This PR is not expected to include it — but the convoy view won't be declared at parity until that slice lands.

Fix 1–3 (and answer 4), then flag for re-review. Enqueue stays held.

rjwittams:
Addressed the parity feedback in `69a849f7`.

1. Failed convoys no longer remain initializing after entering a terminal phase. The Aggregator now emits `initializing: false` for terminal phases, and the detail renderer defensively refuses to let a stale initializing flag mask a terminal phase/message. Added source and render regressions for the live `missing input topic` shape.
2. `h` now exits the task pane alongside left-arrow and Esc, with a navigation regression test.
3. Diagnosis of the reported `a` behavior: it dispatched `SelectWorkspace`, whose no-presentation-manager behavior is intentionally successful but side-effect-free. Since the working attach interaction belongs to the separately scoped crew-row workflow, I removed the advertised `a` binding and bespoke handler rather than replacing it with another partial leg-level attach. The generic panel attach intent remains available for that renderer.
4. The completion e2e disables controllers because the test directly owns and mutates Convoy status to isolate Aggregator → panel → TUI → completion dispatch. Running controllers concurrently would race that fixture; controller reconciliation is covered by the daemon runtime flow tests. The hand-created Presentation was added only to force an attach intent and did not contribute to completion coverage. It has now been removed, along with the unrelated workspace assertions, and the reason for `start_controllers: false` is documented inline.

Validation is green locally: pinned formatter, workspace clippy with `-D warnings`, full workspace tests, both convoy-view e2es, and all 50 render snapshots unchanged. Ready for re-review; still draft and not enqueued.

rjwittams:
Followed up on the automated review’s priority findings in `87c74675`:

- Aggregator watch/list/channel termination now returns `ResourceError`; `spawn_aggregator_task` runs it under bounded supervision. Every restart recreates the durable and observed watches from a fresh list/generation cursor and reloads the complete replica cache. Added a closed-channel regression for the restart contract.
- Removed the production-dead `apply_replica_snapshot`; replacement/removal coverage now exercises the complete-cache path used by runtime.
- Converted the remaining `local_seqs` lock unwraps to `expect("sequence lock poisoned")` for a consistent client failure context.

The remaining automated suggestions (shared panel keys/accessors, additional protocol/integration coverage, vocabulary polish, and presentation-map cleanup) are useful non-blocking follow-up work rather than parity blockers.

Exact local gates remain green: pinned format, workspace clippy with `-D warnings`, full workspace tests, and unchanged render snapshots. PR remains draft and unqueued.

rjwittams:
Amendment to must-fix 3 from Robert: leave the `a` binding in place wired as-is — don't remove it. The real attach semantics get fixed as part of the follow-on view work, not this PR. Must-fixes 1 (initializing masking failure) and 2 (focus trap) stand, plus the question in 4.

rjwittams:
Addressed the latest review follow-up in `3a74e881`:

- Restored the `a` binding exactly as requested in Robert's amendment. It remains wired to the existing attach path; this PR does not change its semantics. The attach/view behavior remains follow-on work.
- Fixed restart relisting so a supervised Aggregator restart clears and rebuilds the local projection from the four authoritative list/watch sources. Bootstrap events are suppressed until every watch is established, then one full snapshot is emitted, preventing rows deleted during an outage from being rebroadcast as ghosts.
- Added `restart_relist_removes_local_rows_missing_from_stores` to cover the stale-row regression.

Local verification:
- `git diff --check`

rjwittams:
Filed the remaining non-blocking panel contract and Aggregator lifecycle hardening as #669. The generic view and functional attach work remain in #666.

PR #673, opened 2026-07-11
fix: preserve qualified checkout paths

rjwittams:
## Summary

- make `CheckoutCreated` carry a `QualifiedPath` instead of a raw `PathBuf`
- preserve existing checkout identities and qualify newly created checkouts from their host or environment execution context
- forward that identity through workspace preparation so attachable sets and presentation bindings correlate to the same checkout
- add host and environment regression coverage and update protocol consumers

## Why

Creating a checkout followed by opening its workspace could produce two attachable sets for the same filesystem path. The terminal-bearing set used a legacy host-name qualifier, while the presentation binding used the stable host-ID qualifier, leaving the original repo-tab attachable set unopenable.

The qualifier was lost because `CheckoutCreated` carried only a `PathBuf`; `PrepareWorkspace` therefore persisted a fallback host-name-qualified key. Carrying `QualifiedPath` through the step outcome makes the identity loss unrepresentable.

## Impact

New host checkouts use the stable host ID, provisioned checkouts use their environment ID, and existing checkout keys remain unchanged. Existing duplicate registry entries are intentionally not migrated.

## Validation

- `CARGO_INCREMENTAL=0 CARGO_PROFILE_DEV_DEBUG=0 cargo check --workspace --all-targets --locked`
- `CARGO_INCREMENTAL=0 CARGO_PROFILE_TEST_DEBUG=0 cargo test -p flotilla-core --locked` — 973 passed
- `CARGO_INCREMENTAL=0 CARGO_PROFILE_TEST_DEBUG=0 cargo test -p flotilla-protocol --locked` — 219 passed

Clippy reaches all changed crates but is currently blocked by the pre-existing duplicated `#[cfg(test)]` attribute in `crates/flotilla-daemon/src/runtime/test_git_repo.rs`.

### DIFF
--- a/crates/flotilla-core/src/executor.rs
+++ b/crates/flotilla-core/src/executor.rs
@@ -14,8 +14,9 @@ use std::{
 };
 
 use flotilla_protocol::{
-    arg::Arg, qualified_path::QualifiedPath, CheckoutTarget, Command, CommandAction, CommandValue, HostName, NodeId, PreparedWorkspace,
-    ResolvedPaneCommand,
+    arg::Arg,
+    qualified_path::{PathQualifier, QualifiedPath},
+    CheckoutTarget, Command, CommandAction, CommandValue, HostName, NodeId, PreparedWorkspace, ResolvedPaneCommand,
 };
 use tracing::{debug, error, info};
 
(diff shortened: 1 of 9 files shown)


rjwittams:
Added the suggested host-context symmetry assertion in ef3e7e5a: the resolver test now verifies the prepared checkout key retains its Host-qualified identity. The separate multi-host ownership observation remains intentionally out of scope for this narrow fix.

PR #674, opened 2026-07-11
feat: publish observed Checkout resources

### LINKED ISSUE #614: Git provider contributes observed Checkout resources (discovery only)
## Parent
#604 · Re-sliced 2026-07-10 per the portfolio steer (`project-map/briefs/flotilla-steer-2026-07-08.md`): this issue is now **discovery only** — the view half moved to the generic-table-tabs slice (see #667-to-be / issue filed alongside this edit).

## What to build
On each refresh cycle the git provider writes **observed `Checkout` resources** (Observed spec variant, `flotilla.work/authority: observed`, `flotilla.work/repo` label) into the ephemeral store for every worktree of every registered repo root — create/update/delete as worktrees appear and vanish, new generation after daemon restart. Resources in the store; no presentation work in this slice. The existing ProviderData path stays untouched (deleted later in #616).

## Acceptance criteria
- [ ] Git provider populates observed Checkouts into the ephemeral store each refresh (create/update/delete as worktrees appear/vanish)
- [ ] Facts-in-spec per the contract: ref/branch, path, repo ref
- [ ] Generation semantics: fresh generation after daemon restart; stale-generation reads are distinguishable
- [ ] Observed Checkouts flow through the Aggregator/replica snapshot like any other resource (visible to a store query / replica-snapshot JSON) — no TUI changes required to verify

## Blocked by
None — the substrate (#626/#627) and Aggregator (PR #664 backend) exist.

rjwittams:
## Summary

- publish each refreshed local checkout as an observed `Checkout` resource in the ephemeral store
- reconcile create, update, and deletion by stable path-derived resource identity while preserving adopted resources
- retain the existing `ProviderData` path during the Plane-A transition and cover the projection at the daemon boundary

Closes #614.

## Validation

- `cargo clippy -p flotilla-core -p flotilla-resources --all-targets --locked -- -D warnings`

## Known baseline failures

The full workspace gates currently stop on unrelated existing failures:

- workspace clippy: duplicated `#[cfg(test)]` in `flotilla-daemon/src/runtime/test_git_repo.rs`
- workspace tests: `tests/flotillad_binary.rs` cannot find the expected root `flotillad` binary target

### DIFF
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -1561,6 +1561,19 @@ impl InProcessDaemon {
             updates.push((identity, last_local_providers, re_snapshot));
         }
 
+        let namespace = self.provisioning_namespace().await;
+        for (identity, local_providers, snapshot) in &updates {
+            if snapshot.errors.iter().any(|error| error.category == "checkouts") {
+                warn!(repo = %identity.path, "skipping observed checkout reconciliation after checkout discovery failed");
+                continue;
+            }
+            if let Err(error) =
+                crate::observed_resources::reconcile_checkouts(&self.observed_resource_backend, &namespace, identity, local_providers).await
+            {
+                warn!(repo = %identity.path, %error, "failed to reconcile observed checkouts");
+            }
+        }
+
         // Apply updates under write lock and broadcast
         let mut repos = self.repos.write().await;
         for (identity, last_local_providers, re_snapshot) in updates {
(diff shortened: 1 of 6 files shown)


rjwittams:
Addressed the repo-key concern in ccf436b6: observed Checkout reconciliation now passes the derived URL through canonicalize_repo_url, rejects the explicit unknown identity fallback, and covers both authority-case normalization and no-op preservation on identity failure. Filed #677 for the valid untrack/remote-change cleanup follow-up; it needs last-root and adopted-resource semantics beyond this projection slice.

rjwittams:
Fixed in f9039ca3. Reconciliation now rejects both unknown and local fallback identities, so a repository without a recognized remote neither publishes a path-derived repo key nor leaks its filesystem path through the repo label. Added regression coverage for the local fallback.

PR #676, opened 2026-07-11
Bound resource store and daemon log growth

### LINKED ISSUE #672: Ready-status restamp + no no-op suppression + unbounded resource_events = disk bomb (filled kiwi's disk)
## Incident
Brief: `project-map/briefs/flotilla-runaway-store-2026-07-10.md`. kiwi's 926 GB disk filled (58.8 GB `resources.sqlite`, 56.87 M `resource_events` rows over **20** objects; APFS snapshots amplified). feta reached 35.6 GB with the same signature; its flotillad is now **stopped** (2026-07-11) pending this fix + store reset. kiwi's store was wiped during the incident.

## Root cause — a self-feeding reconcile loop (diagnosed from code, 2026-07-11)

1. **Unconditional Ready restamp.** Once a TaskWorkspace is Ready, *every* reconcile pass ends in `TaskWorkspaceDeps::ready(...)` → `MarkReady { ready_at: now }` with a fresh timestamp (`reconcilers/task_workspace.rs:104` + the `reconcile()` mapping), and `MarkReady::apply` overwrites `status.ready_at` unconditionally (`resources/src/task_workspace.rs:71`).
2. **No no-op suppression.** `apply_status_patch` → `update_status_typed` (sqlite.rs:386, in_memory equivalent) never compares new vs current status: it always bumps `resourceVersion`, rewrites the body, and appends a MODIFIED event.
3. **The event feeds the loop.** The ControllerLoop's own primary watch delivers that MODIFIED back to itself → reconcile → step 1. The loop spins at watch latency — no resync timer involved. 56.8 M events in days is consistent.
4. The brief's "two serializers flapping A↔B" hypothesis is not it: both `.941785` and `.948` are chrono's trailing-zero-trimmed RFC3339 (948000 µs prints as `.948`) — they're two *fresh* `now()` stamps ~6 ms apart. Monotonic restamp, one serializer.
5. **Host heartbeat is a legitimate periodic write** (`runtime.rs:490`, `heartbeat_at: now` per beat) — 47.9 k Host versions is just uptime. Harmless *only* if the event log is bounded, which it isn't:
6. **`resource_events` has no retention** — append-only forever. Any steady writer is a slow disk bomb; the hot loop just found it first.
7. Secondary: rollback-journal mode (read-only queries fail `database is locked (5)` mid-write; journal churn amplifies physical writes); no rotation on daemon/tui logs.

## Fixes (ordered)

1. **Semantic no-op suppression in `update_status` (and `update`) — both backends.** If the encoded new status equals the current one: return the current object, no version bump, no event, no write. This kills the entire class regardless of patch sloppiness, and is the k8s-alike behaviour. Contract-test against both backends.
2. **Transition-only timestamps.** `MarkReady` preserves an existing `ready_at` (`get_or_insert` semantics); same audit for `started_at`/`finished_at` on other patches. With (1) this makes steady-state reconciles produce byte-identical status → zero writes.
3. **Retention on `resource_events`.** Prune to a bounded window (last N versions and/or T minutes), tracked compaction floor; `watch(FromVersion)` below the floor returns a distinct desync error → consumers relist (ControllerLoop supervision and the Aggregator's restart-relist already handle that path). The events table is a watch buffer, not history.
4. **Log rotation** for daemon.log/tui.log (size-capped).
5. **WAL mode + busy_timeout** pragmas on open.
6. **Tripwire:** warn loudly (tracing + host status) when `resource_events` rows exceed a sane multiple of `resource_objects` — 56.8 M over 20 must be unreachable silently.

## Verification (from the brief)
- Daemon idle ⇒ `MAX(rowid)` on `resource_events` flat across minutes; steady-state TaskWorkspace holds a constant `resourceVersion`.
- Heartbeats continue to update Host without unbounded event growth (retention prunes).
- Watch consumer behind the compaction floor relists and converges (test).

## Priority
**P0 — precedes all other batches (#668 etc.).** The dogfood fleet cannot run until this lands; feta additionally needs a store reset after upgrade.

rjwittams:
Closes #672.

## Summary

- suppress semantic no-op resource and status updates so idle reconcilers do not advance resource versions or append events
- keep lifecycle timestamps stable unless their associated phase actually transitions
- bound retained events per resource stream, persist SQLite compaction floors, and return a typed expiry error for lagging watches
- enable SQLite WAL mode and a busy timeout
- publish resource-store diagnostics and explicit growth tripwires through host status and tracing
- cap daemon log files with size-based rotation
- repair the repository-wide Clippy and binary integration-test gates

## Root cause

Reconcilers could repeatedly write semantically unchanged resources. Each write advanced the resource version, appended a SQLite event, and refreshed lifecycle timestamps, allowing an idle daemon to grow its database and logs indefinitely.

## Impact

Idle resource state now converges to a fixed point. Resource history and daemon logs have explicit bounds, and operators can observe retention and event-to-object warnings. A watcher that falls behind compaction receives `WatchExpired`; production controller and aggregator loops recover through their existing supervised list-and-watch restart.

## Follow-ups

- #678 preserves typed watch expiry through HTTP 410 and explores in-place relisting instead of whole-loop restart.
- #679 defines timestamp reset semantics before retryable resource lifecycles are introduced.

## Validation

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1307,6 +1307,7 @@ dependencies = [
  "flotilla-transport",
  "futures",
  "indexmap",
+ "rusqlite",
  "serde",
  "serde_json",
  "tempfile",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -728,15 +728,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "crossbeam-channel"
-version = "0.5.15"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "82b8f8f868b36967f9606790d1903570de9ceaf870a7bf9fbbd3016d636a2cb2"
-dependencies = [
- "crossbeam-utils",
-]
-
-[[package]]
 name = "crossbeam-deque"
 version = "0.8.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1322,7 +1313,6 @@ dependencies = [
  "tokio",
  "tokio-util",
  "tracing",
- "tracing-appender",
  "tracing-subscriber",
  "uuid",
  "visibility",
@@ -1411,7 +1401,6 @@ dependencies = [
  "time",
  "tokio",
  "tracing",
- "tracing-appender",
  "tracing-subscriber",
  "tui-input",
  "tui-tree-widget",
@@ -4151,14 +4140,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "743bd48c283afc0388f9b8827b976905fb217ad9e647fae3a379a9283c4def2c"
 dependencies = [
  "deranged",
- "itoa",
  "libc",
  "num-conv",
  "num_threads",
  "powerfmt",
  "serde_core",
  "time-core",
- "time-macros",
 ]
 
 [[package]]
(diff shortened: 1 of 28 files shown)


rjwittams:
Reviewed against the production paths and #672's acceptance criteria. ControllerLoop and Aggregator currently recover an expired watch by propagating it to their supervisors, which restart from a fresh list; the PR description now states that recovery granularity explicitly. Filed #678 for HTTP 410 typing and in-place relisting. The terminal/workspace/convoy lifecycles are one-shot today, so sticky timestamps are intentional; filed #679 to define reset semantics before retries are introduced. The bon::Builder derives follow CLAUDE.md's requirement for types with more than three fields.

PR #686, opened 2026-07-11
fix: make finalizer deletion idempotent

### LINKED ISSUE #662: Controller loop should treat NotFound as idempotent at post-get write sites
From PR #660 review by @claude: https://github.com/flotilla-org/flotilla/pull/660#issuecomment-4918245815

PR #660 fixed the reported finalizer-removal race by treating `NotFound` as idempotent success after `run_finalizer` completes. The review noted two related post-get write sites that can still propagate `NotFound` and stop the controller loop if the primary disappears between read and write:

- finalizer attachment: `primary.update(...)` after `get()`
- normal status patching: `apply_status_patch(...)` after reconcile

Investigate and normalize these to the same "gone means done" semantics where appropriate, with controller-loop regression coverage for each timing window.

This is intentionally follow-up scope; #660 only targets the teardown race reported in #651.

### LINKED ISSUE #665: delete on an already-marked resource hard-deletes it, skipping pending finalizers
## What happens

Both `SqliteBackend::delete_typed` and `InMemoryBackend::delete_typed` share this shape:

```rust
if !object.metadata.finalizers.is_empty() && object.metadata.deletion_timestamp.is_none() {
    // mark: set deletion_timestamp, emit Modified
} else {
    // hard DELETE, emit Deleted
}
```

An object that is **already marked for deletion but whose finalizers have not been cleared** (finalizers non-empty, `deletion_timestamp` set) falls into the else branch: a *second* `delete()` call removes it immediately and its `run_finalizer` never executes. k8s semantics require repeat deletes to be no-ops while deletion is pending — removal should only happen when the finalizer list is emptied by the reconciler.

## Why the second delete reliably happens

`cleanup_actuations` (convoy reconcile) issues `Actuation::DeleteTaskWorkspace` on **every reconcile** while the workspace is still present in the watch map (`task_workspaces.contains_key`). A workspace pending finalization *is* still present. So any convoy reconcile that lands in the window between "workspace marked" and "task-workspace finalizer completes" — a resync tick, any watch event, or a daemon restart — re-deletes it and hard-removes it mid-cascade. Children (TerminalSessions, per the task-workspace finalizer at `task_workspace.rs` `run_finalizer`) are orphaned: no `deletion_timestamp`, live backing session, `flotilla ls` shows the convoy `running` forever.

## Observed live (2026-07-10, feta, during the #664 parity check)

`day3` (adopted vessel, scratch template): leg completed → daemon restarted seconds later → on restart the convoy resync re-issued `DeleteTaskWorkspace` → `day3-work` hard-deleted with `flotilla.work/task-workspace-teardown` still attached → `terminal-day3-work-shell` stranded: authority `managed`, `deletion_timestamp: None`, finalizer present, shpool session alive, `ls` reporting `running` while the panel row correctly reports the convoy `completed`. Same scenario without a restart passes only because the task-workspace loop usually wins the race.

This is almost certainly the deeper mechanism behind #651 (closed — #660/#661 fixed the symptom of a *missing* resource after finalizer, not the hard-delete that causes the strand).

## Fix directions

1. **Store-level (the real fix):** repeat `delete()` while deletion is pending must be a no-op (return Ok, no event). Hard removal only via the finalizer-clearing update path. Contract-test it once against both backends (the two `delete_typed`s are copy-paste siblings — candidates for sharing).
2. **Belt-and-braces in `cleanup_actuations`:** skip children whose `deletion_timestamp` is already set — deleting something already being deleted is wasted work even with correct store semantics.

## Acceptance

- [ ] Contract test (both backends): create with finalizer → delete → delete again → object still exists, marked; finalizer clear → object removed with Deleted event.
- [ ] Scenario test: convoy completes, task-workspace finalizer delayed, convoy reconciles again in the window → TerminalSession still gets deleted when the finalizer runs.
- [ ] A stranded-session repro (complete + immediate daemon restart) converges after restart instead of stranding.

rjwittams:
Fixes #665
Fixes #662

## What changed

- Make repeated deletes of finalizer-pending resources a no-op in both in-memory and SQLite stores.
- Avoid scheduling another TaskWorkspace delete while that workspace is already pending finalization.
- Treat `NotFound` as idempotent success at all controller post-`get()` write sites.

## Why

A second cleanup delete could hard-delete a TaskWorkspace after its deletion timestamp was set but before its finalizer ran, stranding child TerminalSessions. Separately, a primary disappearing between a controller read and a write could stop the controller loop.

## Impact

Pending finalizers now retain control of teardown, including across a SQLite-backed restart; controllers continue processing later resources after these expected races.

## Validation

### DIFF
--- a/crates/flotilla-controllers/tests/common/mod.rs
+++ b/crates/flotilla-controllers/tests/common/mod.rs
@@ -8,8 +8,8 @@ use flotilla_resources::{
     Convoy, ConvoyRepositorySpec, ConvoySpec, ConvoyStatus, CrewSource, CrewSpec, DockerCheckoutStrategy, DockerEnvironmentSpec,
     DockerPerTaskPlacementPolicySpec, Environment, EnvironmentPhase, EnvironmentSpec, EnvironmentStatus, HostDirectEnvironmentSpec,
     HostDirectPlacementPolicyCheckout, HostDirectPlacementPolicySpec, InputMeta, PlacementPolicy, PlacementPolicySpec, ResourceBackend,
-    TerminalSession, TerminalSessionPhase, TerminalSessionSpec, TerminalSessionStatus, Vessel, VesselRequirement, VesselSpec,
-    WorkflowSnapshot,
+    TerminalSession, TerminalSessionPhase, TerminalSessionSpec, TerminalSessionStatus, Vessel, VesselRequirement, VesselSpec, WorkPhase,
+    WorkState, WorkflowSnapshot,
 };
 use tokio::{
     task::JoinHandle,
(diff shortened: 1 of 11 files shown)


rjwittams:
Verified locally and in CI: all four checks are green. Thanks for the thorough review.

Issue #683, opened 2026-07-11
aggregator: sessions query for free-floating vessels/sessions (attachables, persistent agents)


rjwittams:
Prerequisite for the #667 manifest connector (design grilled 2026-07-11).

*(Updated 2026-07-13: #684/#692 landed ADR 0011 — the wire is typed named-query result sets, not panels. Reworded accordingly; intent unchanged.)*

The `convoys` query is the only named query the Aggregator maintains. The connector's catalog must also list **free-floating vessels/sessions** — running `TerminalSession`s with no convoy label (today's attachable sets; future persistent agents like the yeoman, which may be archipelago-level, i.e. parented to no project). These appear in `flotilla ls`-adjacent fleet rows but in no result set, so the connector has nothing to project.

Add a second named query: a new `QueryId` variant with a typed `SessionRow` (`Rows::Sessions`), a second `QueryProjection<SessionRow>` instantiation in the Aggregator, replica-aware like convoys (its local result set rides `FleetReplicaSnapshot.result_sets`; fleet merge is the same federated query union). Consumers subscribe via `SubscribeQueries` alongside `convoys`. This is also squarely the #666 generic-per-kind-tables direction — same backend need, two consumers (TUI tables and the manifest connector).

`SessionRow` needs (typed fields per ADR 0011): name, repo (when known), host, attach capability (observed workspace/session ref, `Some` = the daemon accepts attach), phase — enough for the connector to publish `[git.repo?] / [flotilla.vessel]` groups with attach recipes.

PR #692, opened 2026-07-13
refactor: split aggregator data plane from table layout — panels become named query result sets

### LINKED ISSUE #684: protocol: split aggregator data plane from table layout — panels become named query result sets
From the #667 design discussion with Robert (2026-07-11). Feeds the #666 grill — this is the answer to its "what survives of the panel wire types" question, pulled forward so #666 designs table config on top of an honest data plane instead of excavating one.

## Problem

The `Panel*` types (`flotilla-protocol/src/panel.rs`, PR #664) are named for their first use — the convoy TUI tab — and weld three different things together:

1. **The query**: which resources, selector, scope, joins. Implicit in aggregator code; `PanelSource`/`PanelScope` gesture at it but nothing consumes them.
2. **The result set**: rows keyed by `ResourceRef`, typed values, hierarchy, capability intents, `seq`, delta maintenance. The actual data plane — what the aggregator really is: **incrementally-maintained result sets of named queries** ("all Convoys, durable ∪ observed, fleet-merged, joined with Presentation attach state").
3. **The layout**: tab composition, panel ids, columns, titles — consumer config for one particular TUI tab, shipped on the wire in every snapshot.

The conflation obscures real inefficiencies:

- **No per-query subscription.** A snapshot is one hardcoded tab; every consumer gets everything, and `seq` is per-tab so nothing resumes at query granularity. The manifest connector (#667) wants "convoys" + "free-floating sessions" (#683); the TUI wants whatever its tabs show.
- **Static config re-shipped as data** — columns/titles/intent definitions repeat per snapshot; they change never, rows change constantly.
- **Stringly rows** — `values: BTreeMap<String, PanelValue>` re-serializes key strings per row and consumers pattern-match `values["phase"].as_str()`. Whether rows are schema'd tuples or flexible maps should be a visible choice about query results, not a side effect of "panels".
- **Joins flattened invisibly** — convoy rows nest task rows with the Presentation join baked into `children`; that it *is* a join is exactly what a consumer reshaping the data needs to know.
- **Replica merge is federated query union**, buried as an implementation detail of one projection.

## Shape

```
wire / data plane:
  ResultSet   { query: QueryId, seq, rows }            // subscribe per named query
  ResultDelta { query: QueryId, seq, changed, removed }
  Row         { resource: ResourceRef, values, children, depends_on, intents }

consumer side (TUI, #666):
  per-kind table config — columns, labels, tab composition, k9s-style
```

- Intents stay on rows: they are capability facts ("the daemon will accept attach on this"), not layout.
- `Panel*`/`TabView` names die with the layout types, which move into TUI config where they belong (#666's territory).
- Per-query subscription + per-query `seq`; replay-on-connect keyed by query.
- Row value representation (schema'd tuples vs map) decided consciously here, with the trade-off written down.

## Sequencing

- After #680 (vocabulary — these types get vessel/work naming there; don't churn twice in flight).
- Before the #667 connector binds (its build is gated behind #672/#680 anyway, so this costs it nothing): the connector consumes *result sets*, not panels.
- #666 designs per-kind table config on top.

rjwittams:
Closes #684.

The `Panel*` wire types (PR #664) welded three things together: the **query** (`PanelSource`/`PanelScope` — zero consumers), the **result set** (rows, seq, deltas — the real data plane), and the **layout** (tabs/columns/titles — re-shipped in every snapshot, ignored by every consumer). This PR makes the wire an honest data plane and kills layout from the wire. ADR 0011 records the decisions; ADR 0005's status is amended (its composition-of-panels view model stands, but consumer-side).

## Wire (flotilla-protocol)

- New `result_set.rs`: `QueryId` (closed enum — `convoys` today, #683 adds sessions), `ResultSet { query, seq, rows }`, `ResultDelta { query, seq, changed, removed }`, and **typed rows** `ConvoyRow`/`VesselRow` with wire `ConvoyPhase`/`WorkPhase` enums. `ResourceRef` moves to its own module (it is row identity, not a panel concept). `panel.rs` is deleted.
- **Typed per-query rows** (decided with the trade-off written down in ADR 0011): no more stringly `values["phase"].as_str()`; joins become visible typed fields — the Presentation join is `VesselRow.attach: Option<String>`, a capability fact. `IntentDefinition`/`IntentTarget`/`RowIntent` die; the namespace/convoy/vessel echo is gone (rows know their identity), capability routing uses `VesselRow.host`.
- **Full per-query subscription**: new `Request::SubscribeQueries { queries: Vec<QueryCursor> }` replaces the connection's subscription set and replays a full `ResultSet` per stale/absent cursor (response mirrors `ReplaySince`). The socket server filters per connection — unsubscribed queries never hit the socket. Query streams leave `ReplaySince` (repo/host streams unchanged); `StreamKey::Panel { tab }` → `StreamKey::Query { query }`.
- Seq semantics documented on `ResultDelta`: contiguous per query (`seq == last_seen + 1`); the client ignores stale deltas (`seq <= last_seen`) and recovers from gaps by re-subscribing with current cursors — which makes the register-before-replay race on subscribe benign.
- **Federation stated in the types**: `FleetReplicaSnapshot.panels` → `result_sets` — replica merge is federated query union of each host's local result sets, origin host stamped onto rows.

## Core / daemon

- `QueryProjection<R>` in `aggregator_projection.rs` is the query-agnostic projection core (local + per-replica rows keyed by `ResourceRef`, per-query seq); the convoys query instantiates it and #683's sessions query is the confirmed second instantiation. The hardcoded `panel_snapshot()` layout builder dies.
- The daemon Aggregator builds `ConvoyRow`/`VesselRow` directly from `ResourceObject<Convoy>` (all `BTreeMap` assembly and phase-label functions deleted); `apply_replica_cache` matches on query id.
- `DaemonHandle::subscribe_queries` added; `InProcessDaemon` implements the replay contract (delivery restriction is a transport concern — the socket server must filter, the in-process broadcast may over-deliver, and this is documented on the trait).

## Client / TUI

- `flotilla-client`: per-query seq tracking and gap detection; gap recovery re-sends `SubscribeQueries` with the full subscribed set.
- TUI subscribes to the convoys query on connect; `panel_row_to_convoy` and all `panel_*` map-lookup helpers collapse into a thin typed `ConvoyRow → ConvoySummary` adapter in `convoy_model.rs`. `flotilla watch` subscribes to all named queries and prints `[query] convoys: …` lines.

## Tests

- Protocol round-trip rewritten for `ResultSet`/`SubscribeQueries`; new socket-level contract test that unsubscribed query events are filtered until `SubscribeQueries` lands and delivered after; client tests for stale-delta ignore and gap→resubscribe (asserting the cursor); in-process subscribe-replay tests including the client-ahead-of-daemon restart guard; daemon integration, convoy e2e, and fixture updates throughout.
- Three TUI tests asserting that completion targets may diverge from row identity were updated/removed: divergence is now structurally impossible on the wire (per-host routing is expressed via `VesselRow.host`, still covered).

## Verification

- `cargo +nightly-2026-03-12 fmt --check`, `cargo clippy --workspace --all-targets --locked -- -D warnings`, `cargo test --workspace --locked` (56 targets, 0 failures), `cargo test -p flotilla-core --features test-support --test in_process_daemon`, `cargo dylint` (no new warnings).
- Grep gates: no `Panel` types remain in flotilla-protocol; no `values.get(` lookups remain in the TUI/aggregator paths.


rjwittams:
Addressed in ea44f50d..HEAD:

1. **local_result_sets ordering** — fixed: now iterates `QueryId::ALL` with an exhaustive per-query match, so adding the #683 sessions query forces this site to be updated (compile error) instead of silently drifting from the docstring.
2. **complete_work always true** — agreed it should become meaningful before general consumers rely on it; deliberately out of scope here (behavior change). Filed as #694.
3. **recover_query_gap coalescing** — it was a deliberate simplification; documented the asymmetry on the function: resubscribes are idempotent (subscription replaced, results seeded monotonically) and stale deltas are dropped in `handle_event` rather than re-triggering recovery.

PR #698, opened 2026-07-13
aggregator: add free-floating sessions query

### LINKED ISSUE #683: aggregator: sessions query for free-floating vessels/sessions (attachables, persistent agents)

rjwittams:
Closes #683.

## What changed

- add a typed `sessions` named query with `SessionRow` and independent query sequencing
- project running free-floating terminal and persistent-agent sessions while excluding convoy-owned sessions
- include session result sets in subscription replay, local fleet snapshots, and replica union
- advertise attach references only when the daemon can resolve them
- recompute attach capabilities when their environment changes
- bump the wire protocol version for the new query variant

## Why

The aggregator only exposed the convoy query after #684. Consumers preparing the free-floating vessel/session UI need a fleet-aware, subscribable result set for running sessions, including truthful attach capability and repository/host context.

## Impact

Clients can subscribe to `QueryId::Sessions` alongside convoys and receive typed full results and deltas. Fleet replicas carry and merge the same query. Sessions associated with convoys remain absent from this result set.

## Validation

- independent standards and issue-spec review, with all findings resolved

### DIFF
--- a/crates/flotilla-core/src/aggregator_projection.rs
+++ b/crates/flotilla-core/src/aggregator_projection.rs
@@ -8,7 +8,7 @@
 use std::{collections::HashMap, sync::Arc};
 
 use flotilla_protocol::{
-    result_set::{ConvoyRow, QueryId, ResultSet, Rows},
+    result_set::{ConvoyRow, QueryId, ResultSet, Rows, SessionRow},
     HostName, ResourceRef,
 };
 use tokio::sync::{RwLock, RwLockWriteGuard};
@@ -26,6 +26,16 @@ impl QueryRow for ConvoyRow {
 
     fn into_rows(rows: Vec<Self>) -> Rows {
         Rows::Convoys(rows)
+    }
+}
+
+impl QueryRow for SessionRow {
+    fn resource(&self) -> &ResourceRef {
+        &self.resource
+    }
+
+    fn into_rows(rows: Vec<Self>) -> Rows {
+        Rows::Sessions(rows)
     }
 }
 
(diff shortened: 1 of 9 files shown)


rjwittams:
Fixed the protocol compatibility issue in 67c7ebd6 by bumping PROTOCOL_VERSION from 9 to 10. The repeated-session-scan observation is valid but outside this PR’s scope; filed as #700 for follow-up.

PR #699, opened 2026-07-13
fix: remove observed Checkouts when a repo is untracked

### LINKED ISSUE #677: Remove observed Checkouts when a repo is untracked
## Context

Follow-up from local review of #674 / #614.

Observed Checkouts are reconciled while a repository is tracked, but untracking the final root for a repository leaves that repository's observed Checkout resources in the ephemeral store until daemon restart. A repository identity/remote change can similarly leave resources under the old repo-key selector.

## What to build

When the final local root for a repository is removed or its repository identity changes, delete only that repository's `LifecycleAuthority::Observed` Checkout resources from the ephemeral store. Preserve adopted resources and do not delete resources while another local root shares the same identity.

## Acceptance criteria

- [ ] Untracking the final local root removes its observed Checkout resources immediately
- [ ] Adopted Checkout resources remain intact
- [ ] Removing one of multiple roots sharing an identity does not remove their shared observed resources
- [ ] A repository identity change does not leave old repo-key-scoped observed Checkouts behind

rjwittams:
## Summary

- remove a repository's observed Checkout resources when its final local root is untracked
- preserve adopted Checkouts and observed Checkouts still contributed by another local root sharing the identity
- migrate already-tracked paths when their discovered repository identity changes, reconciling both the old and new repo-key projections
- serialize observed Checkout reconciliation with removal so an in-flight refresh cannot recreate deleted resources

## Root cause

Observed Checkout cleanup only ran from the refresh reconciliation path. Once the final local root was removed, no later refresh existed to delete that repository's stale observed projection. Re-adding an already tracked path also returned before re-running discovery, so a changed remote identity could leave resources under the old repo key.

## Impact

Untracking is now reflected immediately in the ephemeral resource store. Lifecycle authority remains respected: only `LifecycleAuthority::Observed` Checkouts are deleted, while adopted resources remain intact.

Closes #677.

## Validation

- `TMPDIR="$PWD/.codex-tmp" cargo test -p flotilla-core --locked --features test-support --test in_process_daemon`
- two-axis standards/spec review, with all findings addressed

### DIFF
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -875,6 +875,9 @@ pub struct InProcessDaemon {
     daemon_socket_path: RwLock<Option<PathBuf>>,
     resource_backend: ResourceBackend,
     observed_resource_backend: ResourceBackend,
+    /// Serializes observed Checkout publication with repository removal so a
+    /// refresh captured before untracking cannot recreate deleted resources.
+    observed_checkout_reconciliation: Mutex<()>,
     aggregator_projection_state: RwLock<AggregatorProjectionState>,
     /// Provisioning namespace used by daemon-side resource operations (e.g.
     /// looking up the Convoy whose task is being marked complete). Set by the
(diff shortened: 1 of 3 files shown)


rjwittams:
Addressed the shared-root timing note in c1972ba3: the test now waits for convergence to the surviving root's observed Checkout and asserts its concrete checkout path. The full 82-test daemon integration suite passes. The other notes do not require changes here: rediscovery is necessary to detect identity changes on an already tracked path; RepoUntracked/RepoTracked preserves the existing migration semantics; and ADR 0004 deliberately keeps observed resources in memory, so the global reconciliation lock does not span persistent I/O.

rjwittams:
Addressed the concurrent identity-migration edge case in 698c348a. If removal loses the race, add_repo now re-reads the path mapping: it continues through the existing idempotent insertion path when the path is absent or already at the discovered identity, and preserves the error when the path remains tracked under a different identity. Formatting, Clippy, both identity-migration tests, and the full 82-test daemon integration suite pass locally.

PR #702, opened 2026-07-13
fix: recover expired resource watches in place

### LINKED ISSUE #678: Preserve watch-expiry semantics across HTTP and recover watches in place
From PR #676 review: https://github.com/flotilla-org/flotilla/pull/676#issuecomment-4946081326

PR #676 adds bounded watch history and a typed `ResourceError::WatchExpired`. The current embedded controller and aggregator paths recover correctly by propagating the error to their supervisors, which restart with a fresh list-and-watch. Two refinements remain:

- `HttpBackend` maps Kubernetes HTTP 410 Gone to `ResourceError::Other`, losing the typed expiry signal.
- Embedded consumers restart the entire supervised loop instead of relisting only the expired watch.

## Work

- Preserve watch-expiry semantics when Kubernetes returns HTTP 410, adjusting the error payload if the HTTP response cannot provide the embedded backend compaction fields.
- Handle `WatchExpired` in `ControllerLoop` and `Aggregator` with an in-place relist/watch restart where state-reset semantics allow it.
- Add HTTP wire and production-consumer convergence tests.

## Acceptance

- A stale HTTP watch produces a typed watch-expiry error.
- Controller and aggregator watches relist and converge without requiring whole-loop supervision where safe.
- Other watch errors retain their existing supervised recovery behavior.

rjwittams:
Closes #678

## Summary

- preserve Kubernetes HTTP 410 responses as typed `WatchExpired` errors, including when the response body cannot be read
- recover controller primary and secondary watches in place, with a full primary resync after a secondary expiry so missed deletions converge
- recover each Aggregator resource/source watch independently by relisting only the affected source
- preserve durable/observed source precedence when an observed presentation masks a stale durable attach target
- add HTTP wire tests and production-consumer convergence tests for expiry and non-expiry behavior

## Root cause

The HTTP resource backend mapped Kubernetes 410 responses through the generic status path, discarding watch-expiry semantics and the requested resource version. Embedded consumers also returned watch expiry to their supervisors, restarting the whole loop. A secondary relist alone could not reveal objects deleted during the expired interval, and Aggregator source reconstruction did not let an observed presentation with no workspace mask a lower-precedence durable workspace.

## Impact

Stale HTTP watches now expose the same typed expiry contract as embedded stores. Controllers and the Aggregator recover locally and converge their projections without restarting unrelated watches or the enclosing supervisor. Non-expiry errors still propagate to supervision.

## Validation

- two-axis standards and issue-spec review

### DIFF
--- a/crates/flotilla-daemon/src/aggregator.rs
+++ b/crates/flotilla-daemon/src/aggregator.rs
@@ -2,6 +2,7 @@
 
 use std::collections::HashMap;
 
+use async_trait::async_trait;
 use flotilla_core::aggregator_projection::AggregatorProjectionState;
 use flotilla_protocol::{
     result_set::{ConvoyPhase, ConvoyRow, CrewMemberSummary, QueryId, ResultDelta, Rows, VesselRow, WorkPhase},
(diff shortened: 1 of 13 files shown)


rjwittams:
Fixed in 94c13fdb: removed the orphaned apply_convoy_event and apply_session_event wrappers, and updated the affected tests to exercise the source-aware event path directly. The full-source rebuild on watch recovery is intentional for correctness: after expiry the relist is authoritative and must also remove objects deleted while the watch was unavailable. We can profile and optimize that path separately if source sizes show it is needed.

Issue #681, opened 2026-07-11
feat: crew-level work state with vessel-level roll-up; completion moves to the crew


rjwittams:
Follow-up to #680 (vessel/work rename), designed at the same 2026-07-11 grill. Builds on #675's handoff mechanics.

## Model

Work status belongs to **agent crew members**; the vessel-level entry is a **roll-up**.

- **`CrewWorkState`** per agent crew member: `Working / Done / HandedBack / Failed` (+ timestamps, message). Support processes (`Tool` sources: shells, dev servers) have **no work state** — the agent-vs-support distinction is structural.
- **Stored in `ConvoyStatus`**, keyed `vessel/role` — *not* on `TerminalSession`: work state is workflow state and must survive session death; #675's `MarkStarting` resets session status on revival, which would wipe it.
- **Completion becomes agent-native**: `flotilla crew complete` with ambient `FLOTILLA_CREW_ID` (same pattern as `flotilla crew <role> handoff`). The implementor declares their own work done — a tool call away, per ADR 0008. Human override remains at the roll-up level (`convoy X work <vessel> complete --force` or similar).
- **`ConvoyStatus.work[vessel]` becomes reconciler-derived**: all agent crew aboard done and none handed back ⇒ `Complete`. DAG gating (`depends_on`) reads the roll-up — same shape as today, different writer. `ConvoyPhase` roll-up unchanged.
- **Hand-back**: a handoff targeting crew whose work is `Done` flips it to `Working` (or `HandedBack` on the sender's side), and the vessel roll-up reopens (`Complete → Running` transition becomes legal). This is the "convoy delivered = all crew done *and not had it handed back*" semantic.

## Design details to settle in the PR

- Exact `CrewWorkPhase` vocabulary and whether `HandedBack` is a phase or an event.
- Interaction with `WorkPhase::Ready/Launching` (provisioning-side states stay vessel-level; work states are agent-level).
- Whether crew work state seeds from the brief delivery (Working on first message) or on session start.

## Acceptance

- coder → reviewer → coder e2e (#675's scenario) driven purely by crew-level completion + hand-back, with the vessel roll-up and ConvoyPhase converging correctly.
- Support processes never appear in work state or block roll-ups.

PR #703, opened 2026-07-13
aggregator: make vessel completion capability reflect work state

### LINKED ISSUE #694: aggregator: make VesselRow.complete_work reflect real completability
From PR #692 review by claude[bot] (https://github.com/flotilla-org/flotilla/pull/692#issuecomment-4960140503), observation 2.

`VesselRow.complete_work` is a capability fact per ADR 0011 ("the daemon will accept completing this vessel's work"), but `Aggregator::summarize_vessel` never overrides its `true` default — equivalent to the old model's unconditional `complete-work` intent, so not a regression, but it reads as if it should be meaningful.

Make it reflect real completability before general consumers (e.g. the #667 manifest connector, #666 table intents) rely on it — at minimum `false` for vessels in terminal work phases (complete/failed/cancelled), and possibly gated on convoy phase too.

rjwittams:
## Summary

- derive `VesselRow.complete_work` from the presence of non-terminal vessel work
- reject forced completion when work is already complete, failed, or cancelled
- share terminal-phase semantics between aggregation and command validation

## Why

`complete_work` was always emitted as `true` because the builder default was never overridden. That made the result-set capability disagree with real vessel state, and the command path also accepted already-terminal work.

After #681, crew completion remains reconciler-owned while this field describes the human override surface. Non-terminal work advertises the override; absent or terminal work does not.

## Impact

Consumers can rely on `complete_work` when deciding whether to expose the human completion override. Direct CLI attempts to complete terminal work now return a clear error instead of succeeding as a no-op or rewriting terminal state.

## Validation

- two-axis standards/spec review against issue #694 and the merged #681 completion-authority model

Closes #694

### DIFF
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -3153,6 +3153,9 @@ impl InProcessDaemon {
                     Some(status) if !status.work.contains_key(work) => {
                         flotilla_protocol::CommandValue::Error { message: format!("convoy {convoy} does not contain work {work}") }
                     }
+                    Some(status) if status.work.get(work).is_some_and(|state| state.phase.is_terminal()) => {
+                        flotilla_protocol::CommandValue::Error { message: format!("convoy {convoy} work {work} is already terminal") }
+                    }
                     Some(_) => {
                         match apply_resource_status_patch(
                             &convoys,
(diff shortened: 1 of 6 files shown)


rjwittams:
Addressed in 1545c661: collapsed the duplicate work lookup, changed the VesselRow builder capability default to fail closed, and added coverage for that default. The TOCTOU retry concern is valid but broader than this fix, so I filed #704 to track re-validating preconditions on patch retry.

PR #705, opened 2026-07-13
fix: reduce per-project refresh scan multiplication

### LINKED ISSUE #689: bug: per-project refresh multiplies host-wide scans and overwhelms syspolicyd
## What happened

Running the Flotilla daemon with ten tracked Projects on macOS caused synchronized bursts of short-lived `git`, `gh`, `zellij`, and `shpool` processes during periodic Fleet Observer refreshes. `syspolicyd` sustained roughly one to two CPU cores, grew beyond 3 GiB RSS, repeatedly missed its assessment cache, and eventually reported "Too many open files."

A controlled pause/resume isolated the trigger: before pausing `flotillad`, `syspolicyd` averaged 74% CPU; while the daemon was paused it fell to 0%; after resuming it averaged 129% and peaked at 244%. Codex Computer Use processes remained running throughout the idle period, ruling them out as the immediate trigger.

Host-wide Provider observations appear to be multiplied by the number of tracked Projects. In particular, Presentation Manager, Terminal Pool, and CloudAgent discovery are refreshed independently for every Project alongside project-scoped repository and GitHub observations.

## What I expected

Periodic refresh should keep Project views and Provider health current without overwhelming the host. Tracking another Project should add project-specific observation work, but should not multiply identical host-wide workspace, terminal-session, or CloudAgent scans.

## Steps to reproduce

1. On macOS 26.5.1, run a development build of Flotilla with around ten local Projects tracked.
2. Configure `shpool` as the preferred Terminal Pool and allow the normal Presentation Manager, repository, GitHub, and CloudAgent Providers to be discovered.
3. Leave the daemon running through several periodic refresh cycles.
4. Observe bursts of short-lived `git`, `gh`, `zellij`, and `shpool` children and rising CPU/RSS for `syspolicyd`.
5. Pause `flotillad` with `kill -STOP $(pgrep -x flotillad)` and observe `syspolicyd` become idle.
6. Resume it with `kill -CONT $(pgrep -x flotillad)` and observe the policy load return.

## Additional context

- Observed on commit `ec278544c915b6332a9491322bb9af238b6e9d30`.
- The local `flotillad` was an ad-hoc-signed development binary. Unified logs showed repeated malware scans, cache misses, signing error `-67062`, and error `100024` (too many open files).
- The affected setup refreshed ten Projects in closely aligned batches.
- Related architectural direction: #538 proposes event-driven snapshot deltas instead of fixed polling. This issue is the current severe failure mode and should be treated as a Plane A bugfix rather than new observer capability.

rjwittams:
## Summary

- cache host-scoped provider discovery and workspace/terminal scans per environment with TTL and single-flight semantics
- share Claude and Cursor account-wide session fetches across projects for at least one minute
- split refreshes into staggered 10-second local and 60-second network tiers while keeping manual refresh full
- batch worktree issue metadata into one repository-wide Git config query

## Why

Each tracked project independently repeated host-wide and account-wide provider work on synchronized refresh ticks. On macOS, the resulting bursts of short-lived `git`, `gh`, `zellij`, and `shpool` processes overwhelmed `syspolicyd` as the project count grew.

## Impact

Adding a project now adds project-specific observation work without multiplying identical host-wide scans. Stable per-project offsets spread startup and periodic work, while provider health and errors remain intact across the fast and slow refresh tiers.

## Checks

- parallel Standards and Spec reviews: no remaining findings

Fixes #689

### DIFF
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -56,7 +56,10 @@ use crate::{
     model::{provider_names_from_registry, repo_name, RepoModel},
     path_context::{DaemonHostPath, ExecutionEnvironmentPath},
     providers::{
-        discovery::{discover_providers, run_host_detectors, DiscoveryResult, DiscoveryRuntime, EnvironmentAssertion, EnvironmentBag},
+        discovery::{
+            discover_providers_with_host_scoped, run_host_detectors, DiscoveryResult, DiscoveryRuntime, EnvironmentAssertion,
+            EnvironmentBag,
+        },
         ssh_runner::SshCommandRunner,
         ChannelLabel, CommandRunner,
     },
@@ -569,7 +572,21 @@ async fn discover_repo_for_environment(
     let remote_env = StaticEnvVars::from_bag(&host_bag);
     let env: &dyn crate::providers::discovery::EnvVars = if environment_id == local_environment_id { &*discovery.env } else { &remote_env };
 
-    Ok(discover_providers(&host_bag, &ee_path, &discovery.repo_detectors, &discovery.factories, config, runner, env).await)
+    let host_scoped = discovery
+        .host_scoped_providers
+        .discover_for_environment(environment_id, &host_bag, &discovery.factories, config, &ee_path, Arc::clone(&runner))
+        .await;
+    Ok(discover_providers_with_host_scoped(
+        &host_bag,
+        &ee_path,
+        &discovery.repo_detectors,
+        &discovery.factories,
+        config,
+        runner,
+        env,
+        &host_scoped,
+    )
+    .await)
 }
 
 fn normalize_local_provider_hosts(
--- a/crates/flotilla-core/src/model.rs
+++ b/crates/flotilla-core/src/model.rs
@@ -15,7 +15,7 @@ use crate::{
         registry::{ProviderRegistry, ProviderSet},
         types::RepoCriteria,
     },
-    refresh::RepoRefreshHandle,
+    refresh::{RefreshSchedule, RepoRefreshHandle},
 };
 pub fn labels_from_registry(registry: &ProviderRegistry) -> RepoLabels {
     fn labels<T: ?Sized>(set: &ProviderSet<T>) -> CategoryLabels {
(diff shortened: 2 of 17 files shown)


rjwittams:
Thanks — addressed the actionable parts in 59ac9fc3. SharedScan now caches only successful results, so transient errors are retried immediately; successful results still honor the TTL, matching the existing account-wide cache behavior. I also added the main versus main-2 branch-prefix regression test. On host-scoped discovery, I verified that host detector state is already captured once when EnvironmentManager registers an environment (new_local calls run_host_detectors once and stores the EnvironmentBag). Re-running only factory probes for a later repo would therefore not detect a newly installed binary or exported variable; restarting or re-registering the environment remains the existing host-state refresh boundary. The once-per-environment provider cache makes that lifecycle explicit while eliminating the per-repo multiplication from #689.

rjwittams:
Fixed in 3843702c. Full refresh now runs the fast and slow tier futures with tokio::join, then installs only the ProviderData fields owned by each tier. Periodic fast and slow refreshes remain independent. Added full_refresh_runs_fast_and_slow_tiers_concurrently with a shared barrier so it fails if either tier waits for the other; all 21 refresh tests and strict workspace Clippy pass locally.

rjwittams:
Addressed both findings in 0851615d: managed_terminals and agents are rebuilt from empty maps on each successful projection, with removal regressions for both; SharedScan now uses a generation guard so an invalidation during an in-flight scan prevents that stale result from being cached, also covered by a concurrency test. I also merged current main to reproduce the failed synthetic-merge CI test. The startup stagger exposed stale queued events in its path-only test helper, so 72fe2386 coalesces pre-initial manual triggers and makes the helper match both path and current repository identity. The failing test passed 10 consecutive runs, the full 82-test in_process_daemon suite passed, all 1009 core library tests passed, and strict workspace Clippy passed locally.

rjwittams:
Round 4 review is clean. The CI timeout exposed that startup-trigger coalescing was draining after the initial scan and could swallow a mutation trigger received during that scan. Fixed in 6fa66489 by draining only before the initial scan; added trigger_during_initial_scan_runs_a_follow_up_refresh. The previously failing daemon test passes 5 consecutive runs, the complete daemon library suite passes (172 passed, 1 ignored), all 23 refresh tests pass, and strict workspace Clippy passes.

Issue #690, opened 2026-07-13
TUI repo tabs: refresh no longer updates issues


rjwittams:
## What happened

In the repo tabs of the TUI, refresh isn't working — issue data (at least) does not update. Manual refresh (`r`) appears to have no effect on the issues shown; whether other provider sections silently share the fault is unconfirmed.

Observed by Robert 2026-07-13 on current main (`ec278544`, post-#685 vessel/work rename), daemons freshly restarted on clean stores after the #672 recovery.

## What I expected

`r` (and periodic refresh) in a repo tab updates provider data for that repo, including the issue tracker section.

## Notes

- Plane A bugfix territory (providers → snapshot → TUI pipeline).
- Not yet diagnosed; candidate suspects are the recent large surfaces that touched the refresh/snapshot path — #664 (aggregator replacing the snapshot wire) and #685 (rename sweep) — but that's proximity, not evidence.
- Possibly interacts with #689 (per-project refresh storm): if refresh scheduling is broken in one direction it may be misbehaving in both, or a refresh debounce/failure could be masking issue updates. Worth checking whether issue *incremental* refresh (`issue incremental` tracing) still fires and whether its results reach the snapshot.

Issue #667, opened 2026-07-10
Design the manifest metadata-patch connector (build after Leg-1 contract freeze)


rjwittams:
## Parent
#604 · From the portfolio steer (`project-map/briefs/flotilla-steer-2026-07-08.md`, Ask 1). Grounding: `project-map/specs/pm-capability-model-v1.md` §4–5, `project-map/briefs/manifest-extraction-convoy-2026-07-08.md`.

## What to design (now) and build (later)
A **flotilla-owned producer** — small process or flotillad facility — that watches the resource store and publishes group-targeted metadata patches speaking the **manifest** wire contract: `flotilla.project.* / convoy.* / leg.* / vessel.* / crew.*` keys, including **latent** convoys/vessels (things that could be materialised but aren't). Presentation Managers (wheelhouse, zellij+andamento) consume patches; they never depend on flotilla — the git-watcher remains the independence proof.

What the PM-capability v1 bar needs from it: truthful live+latent convoy/vessel listing and phase/attention facts for badges.

## Constraints
- **Design against spec §4 now; do not bind until the manifest extraction convoy's Leg 1 freezes the contract** (pipe names `manifest-*`, `pm.*` keys, targets `Pane|Workspace|Group|Identity`, `materialize.target = workspace|pane`).
- Transport is a per-host adapter on the manifest side; the connector swaps only its send function.
- Natural implementation substrate: a consumer of the Aggregator/watch stream (PR #664 backend) — the connector is basically a panel-stream-to-manifest-patch projection.

## Acceptance (design phase)
- [x] Design sketch: where it runs (flotillad facility vs sidecar), what it watches, key mapping table (resource/status fields → `flotilla.*` keys), latent-entity derivation
- [x] Reviewed against spec §4–5; gaps the v1 bar needs but the contract lacks are reported back to the portfolio map
- [x] Build ticket sliced (#708 — pulled ahead of Leg 1 per the 2026-07-11 amendment)

## Blocked by
Build: manifest convoy Leg 1 (external). Design: nothing.

rjwittams:
## Design update (2026-07-11, with Robert): pulled forward to v0, plus the pane-identity half

[...]

Issue #666, opened 2026-07-10
Generic per-kind table tabs (k9s-style) — v1 slice of ADR 0012


rjwittams:
## Parent
#604 · Grilled 2026-07-13 → **ADR 0012** (PR #701); re-derived scenario-first 2026-07-14 → **ADR 0014** (merged, amends 0012's generic tier). **Now an umbrella**: implementation split into four slices after Codex's pre-implementation survey (bespoke split/tree screen needs replacement not wrapping; Plane-A action menu not reusable; #729 owns the QueryId reshape).

## Slices
1. #733 — table foundation + convoys/vessels conversion (view-model registry, widget, drill = tab-local address history, vessel attach/complete). **Starts immediately.**
2. #734 — independents table (with #731 rename).
3. #735 — scoped tabs: checkouts + demand-backed issues tables (after #729/#730; start-convoy row action → #732).
4. #736 — filtering: `/` fuzzy + structured scope params in the address.

Recorded drill ruling: Enter mutates the current tab's address (tab-local history), Esc pops; new-tab drill is a later explicit modifier, never the default.

## Explicitly out (recorded follow-ons)
The generic `Resources{kind}` tier — repositioned by ADR 0014 as a possible future tier, columnar (SoA) if ever built. Also: Miller-columns cascade; TOML column overrides; saved queries; fleet rows; connector-absence hints. (The composite scope page — the repo-view successor — was pulled **into** scope 2026-07-20 as #764, with `independents{scope}` as #765; no longer excluded.) Tables never gain create intents beyond row-scoped actions.

## Acceptance (umbrella — closes when all slices land and this passes)
**Venue (2026-07-20 ruling): the acceptance day is conducted on #764's composite project page.** During the transition both the Plane-A repo page and the project page stay openable side by side; the default flips to the project page only after this acceptance passes — old workflow untouched until the fully-convoyed workflow is proven.
A day of convoy-driven work in which the tables are how you **observe** (phases, failure messages visible — never a placeholder; staleness honest per class: `as_of` on demand-backed rows, replica-cache staleness on store-backed; density ≥ `flotilla ls`) and how you **recover** (attach at both granularities, force-complete, teardown; no focus traps; mouse fine) — while normal starts flow through auto-attach without touching the tables. Snapshot diffs reviewed, not blind-accepted.

## Blocked by
Slices: #733 (none) → #734 (#731), #735 (#729/#730), #736 (#735).

rjwittams:
[...]

PR #706, opened 2026-07-13
TUI arbitrary tab model: Views are the addressable unit (ADR 0013)

### LINKED ISSUE #589: TUI: arbitrary tab model

rjwittams:
Implements the arbitrary tab model designed in the #589 grill (see the [design comment](https://github.com/flotilla-org/flotilla/issues/589#issuecomment-4961330086)). Contract in **ADR 0013** (`docs/adr/0013-views-are-the-addressable-unit.md`); CONTEXT.md gains the **View** glossary entry.

## What this does

**Views are the addressable unit.** A `ViewAddress` (flotilla-protocol) is a kind-rooted path — `overview`, `repo/<authority>/<path>`, `convoys/<ns>`, `convoy/<ns>/<name>`, `vessel/<ns>/<convoy>/<vessel>`, `project/<ns>/<name>` — with an optional `flotilla://` prefix. The frozen external surface is exactly kind names + positional param signatures (quasi-external: #667 manifest recipes will embed these).

**The tab set is an explicit, owned list** (`OpenViews`): pinned overview at index 0, open-or-focus identity, drag/`{`/`}` reorder on every tab, close via `S-X` or the active tab's mouse `×`. Persisted to `open-views.toml` (`{address, label_override}`); `tab-order.json` is now a read-only legacy loader and the daemon's two tab-order writers are gone. Registered repo ≠ open tab: `[+]` still registers-and-opens, but repos registered elsewhere just appear on the overview.

**Per-tab loud failure.** A broken address or dangling repo/convoy/vessel renders that tab's error state; other tabs are unaffected; nothing silently falls back or prunes the persisted set.

**Tabs drive subscriptions.** The open-view set computes the `SubscribeQueries` union (recomputed on open/close, cursors kept per query). Plane-A repo streams stay global and outside this lifecycle, dying with Plane A.

**Scoped mode** — the PM capability model's scoped pane: `flotilla view <address>` runs exactly that View with no tab bar and no `open-views.toml` coupling; opens navigate in place. `:open <address>` in the palette does the interactive equivalent.

**New kinds:** `convoy` (full-width vessel tree), `vessel` (detail + pinned attach/complete target), `project` (convoy list scoped by `ConvoyRow.project_ref` — the recentering flagship; more panels ride the same address as their queries land). The convoys machinery is scope-threaded (`ConvoyScope`), which also removed the four hardcoded `"flotilla"` namespaces.

## Behaviour changes to know about

- A repo tab you close stays closed (the point of an ownable tab set). First run seeds the old bar exactly.
- `{`/`}` on the convoys tab now moves the convoys tab (was: silently reordered the hidden repo list pre-#690, then no-op). The pinned overview is the only move/close guard.
- Removing the last repo no longer quits the TUI; the tab dangles loudly instead.
- Esc on the overview returns to the previously active tab.
- Snapshot diffs are exclusively the active tab's new `×` close affordance (verified line-by-line before accepting).

## Notes for review

- `ViewAddress::Repo` renders absolute-path fallback identities (`local//tmp/...`) as one percent-encoded segment — a round-trip bug found by test migration (all local-repo tabs would have loaded as Broken).
- `AttachConvoyVessel` from convoy views now sends a context-free command when no repo is active (previously rode the hidden active repo; the `a` binding is a known dead end per the #664 parity notes either way).
- `ConvoysUiState` stays a singleton shared by convoy-scoped views; keying it per address is deferred until multiple same-scope views are real.
- Dylint inline-path warnings match the pre-existing style of the touched files; left for a dedicated sweep.
- Deferred (recorded as non-decisions in the issue): remote-open into a running TUI, delegate-to-PM outward opens, `custom/<name>` compositions, collision-based label shortening, scoped-mode back stack.

Closes #589. Unblocks #666.


rjwittams:
Both points verified and handled:

**Modal panic (fixed in d5f0a5a6):** confirmed real — `handle_repo_removed` now dismisses open modals when the untracked repo is the active tab's scope, with a regression test (`handle_repo_removed_dismisses_modals_open_on_the_active_repo_tab`). Took the backstop route rather than per-widget guards: modals are opened against the active tab by construction and daemon events are processed before the next render, so dismissal at the removal site closes the whole class — while modals that could newly open on a dangling tab are already gated (action menu and palette repo-context checks go through `active_opt()`, which returns None for untracked repos).

**Dead `resolve_repo_roots`:** confirmed — no callers anywhere in the workspace. Filed as #707 rather than deleting here, because removing it also decides the end-of-life of the legacy `load_tab_order` ordering and `tab-order.json` support, which deserves its own small PR.

rjwittams:
Addressed the offline code-review round in 6475b95a — all 11 findings verified against the codebase first; 10 fixed, 1 partially (with reason):

**Standards**
- `scoped: bool` → `TabSetMode::{Tabbed, Scoped}` on `OpenViews`; the App/RenderContext bools are gone (derived via `views.is_scoped()`). This also resolved the index-0 primitive-obsession finding: pinning is now explicit policy per mode (`is_pinned_index`), not an accident of position.
- New config-test `.unwrap()`s → `.expect()`.
- `SaveTabOrder` → `PersistOpenViews`.
- Scattered view-kind policy: centralized the non-rendering half (query, convoy scope, binding-mode stacks) in `app::view_kind` — key resolution, screen status modes, subscriptions, and the run-loop prefilter all consult it. Rendering dispatch and labels deliberately stay with the widgets (stated in the module docs): a new kind now touches view_kind + one render arm + one label.

**Spec**
- Scoped back stack: in-place opens push the previous address; Esc pops (`OpenViews::back`). Tested.
- Scoped tab bindings: TabPage split into TabPage (app globals) + TabShell (`[ ] { } S-X`); scoped panes compose without TabShell, so no tab keys and no tab-key hints. `keybindings.toml` gains a `tab_shell` section.
- Vessel-view mistargeting (the real behavior bug): `ConvoyScope` gains a `vessel` pin — j/k no-ops on vessel views and attach/complete target the addressed vessel, never the shared selection. Tested with a sibling-vessel discriminator.
- Dangling project addresses: **partial** — the TUI has no projects query, so "empty project" and "no such project" are indistinguishable today; the empty state now names the project and says exactly that instead of implying either. True per-view existence failure needs a projects query (#666 territory).
- Label collisions: render-time pass widens colliding short labels with qualifying params (`leg-1` → `alpha/leg-1`); overrides exempt. Tested.
- ADR 0013 now states how its host-free *identity* rule composes with ADR 0012's host *scope params* (they ride the optional `?key=value` channel — filters, never identity).

fmt/clippy/workspace tests all green locally.

rjwittams:
Fixed in the follow-up commit: label widening is now progressive per kind (vessel → convoy/vessel → namespace/convoy/vessel), so cross-namespace vessel collisions qualify all the way to the namespace instead of stalling at an ambiguous convoy/vessel. Same-convoy collisions still get the shorter form. Covered by `cross_namespace_vessel_collisions_widen_to_the_namespace`.

PR #711, opened 2026-07-13
fix: clean stale aggregator presentation joins

### LINKED ISSUE #669: Harden aggregator lifecycle coverage (post-result-set split)
From the automated re-review of PR #664: https://github.com/flotilla-org/flotilla/pull/664#issuecomment-4934965594

*(Updated 2026-07-13: #692 / ADR 0011 replaced the `Panel*` wire types with typed named-query result sets, which deleted most of this list at the root — the shared stringly field/key vocabulary, the `PanelValue` accessor surface, `PanelField` serde tagging, and the `PanelDelta`/non-string-`PanelValue` round-trip gaps no longer exist as concepts. `ResultSet`/`ResultDelta` round-trip coverage and a socket-level subscription contract test landed with #692.)*

Still live:

- Broaden Aggregator integration coverage beyond the happy path, including generation/reset and presentation-map cleanup behavior.
- Remove the redundant full `ResultSet` broadcast during supervised Aggregator restart if measurement or tracing confirms it is unnecessary.

The generic table/view semantics and functional attach UX remain tracked by #666 rather than this issue.

rjwittams:
Closes #669

## Summary

- scope Aggregator presentation workspace joins to vessels in the current effective Convoy projection
- discard lingering workspace contributions as soon as their Convoy is deleted
- add a focused lifecycle regression for Convoy deletion while its Presentation remains live

## Root cause

The Aggregator rebuilt `presentation_workspaces` from every live Presentation, independently of whether the referenced Convoy and vessel still existed. Presentation cleanup normally follows Convoy cleanup, but during that interval the derived join retained a stale attach contribution.

Later result-set and watch-recovery work already covered the broader lifecycle items originally recorded in #669. This change addresses the remaining presentation-map cleanup without mutating the authoritative durable or observed Presentation source caches.

## Impact

Deleting a Convoy now removes its workspace contribution from the Aggregator join immediately, even if owner cleanup has not deleted the Presentation yet. Existing durable/observed precedence and watch-relist behavior are unchanged.

## Validation

- focused Aggregator behavior and integration tests
- two-axis standards and issue-spec review

### DIFF
--- a/crates/flotilla-daemon/src/aggregator.rs
+++ b/crates/flotilla-daemon/src/aggregator.rs
@@ -391,11 +391,31 @@ impl Aggregator {
     }
 
     async fn rebuild_local_projection(&mut self) {
+        let mut effective_convoys = HashMap::new();
+        for source in LOCAL_SOURCE_PRECEDENCE {
+            let Some(convoys) = self.convoys_by_source.get(&source) else { continue };
+            effective_convoys.extend(convoys.iter().map(|(reference, convoy)| (reference.clone(), convoy.clone())));
+        }
+        let presentation_keys = effective_convoys
+            .values()
+            .flat_map(|convoy| {
+                convoy.status.as_ref().and_then(|status| status.workflow_snapshot.as_ref()).into_iter().flat_map(|snapshot| {
+                    snapshot
+                        .vessels
+                        .iter()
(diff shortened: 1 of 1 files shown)


rjwittams:
Reviewed; no code change needed. The suggested vessel-removal case is not a live lifecycle because workflow_snapshot is immutable after bootstrap, and the per-vessel key clones are negligible at expected vessel counts. The focused Convoy-deletion regression covers the remaining #669 cleanup.

PR #712, opened 2026-07-13
fix: never mistake an incompatible daemon for a stale socket in connect_or_spawn

rjwittams:
Ports the surviving half of an uncommitted pre-#692 draft found in the main checkout during the fleet upgrade (preserved on `handshake-hardening-draft-wip`). The other half — client-side version check in `do_client_hello` and the server replying with its Hello before closing on mismatch — already landed with #692; this PR carries only what main lacks.

## The gap

`connect_or_spawn` probed for an existing daemon with plain `SocketDaemon::connect`, which performs **no Hello handshake at all** — and treated *any* connect failure as "no daemon here", falling through to the spawn path. Consequences with protocol v8's handshake enforcement:

- a client that can't handshake with a **live but version-mismatched daemon** would delete the socket out from under it and **spawn a second daemon** on the same state dir;
- the mismatch diagnostic ("daemon has vN, client has vM — restart the daemon") that #692 added could never surface through the `connect_or_spawn` path, because the stateless probe doesn't exchange versions.

Exactly the window a fleet upgrade opens — found while doing one.

## Change

- New `connect_existing_stateful`: connection-refused ⇒ `Ok(None)` (spawn path allowed); connection accepted ⇒ the stateful Hello **must** succeed, and any handshake failure (version mismatch, malformed reply) is returned verbatim instead of being treated as a stale socket.
- All four probe sites in `connect_or_spawn` (initial, post-lock-wait, last-chance, post-spawn poll) go through it.
- Tests: `do_client_hello` rejects a stale-version daemon with a diagnostic naming both versions; `connect_or_spawn` against a live mismatched listener errors instead of reusing or replacing it.

## Validation

- `cargo test --workspace --locked` (56 suites)

### DIFF
--- a/crates/flotilla-client/src/lib.rs
+++ b/crates/flotilla-client/src/lib.rs
@@ -320,6 +320,13 @@ pub async fn connect_or_spawn(
     // handshake failure means a daemon is listening but is incompatible or
     // malformed; surface that error instead of treating the socket as stale
     // and silently spawning a second daemon over the live process.
+    //
+    // Deliberately no retry at this probe, unlike the two probes below: those
+    // sit in windows where a *just-spawned* daemon is expected and an error is
+    // probably a startup race, while a steady-state daemon that fails a
+    // 5s-bounded handshake is a condition the caller should hear about
+    // immediately — this is the interactive path, and retries would only
+    // delay an error the user has to act on anyway.
     if let Some(daemon) = connect_existing_stateful(socket_path).await? {
         return Ok(daemon);
     }
@@ -405,7 +412,9 @@ pub async fn connect_or_spawn(
         spawn_daemon(config_dir, config_dir_override, socket_override)?;
     }
 
-    // Poll for connection with a 10s deadline. Handshake errors here are
+    // Poll for connection with a 10s deadline (soft: the deadline is checked
+    // between probes, and a probe can block up to HELLO_HANDSHAKE_TIMEOUT, so
+    // the true worst-case is deadline + handshake timeout). Handshake errors here are
     // retried rather than propagated: the daemon on this socket was spawned by
     // us moments ago, so an error is far more likely a startup race (accepted
     // before the serve loop is up) than a genuine incompatibility. The last
--- a/crates/flotilla-client/src/lib.rs
+++ b/crates/flotilla-client/src/lib.rs
@@ -316,8 +316,11 @@ pub async fn connect_or_spawn(
     config_dir_override: Option<&Path>,
     socket_override: Option<&Path>,
 ) -> Result<Arc<SocketDaemon>, String> {
-    // Try to connect to existing daemon
-    if let Ok(daemon) = SocketDaemon::connect(socket_path).await {
+    // An existing socket must complete the stateful Hello handshake. A
+    // handshake failure means a daemon is listening but is incompatible or
+    // malformed; surface that error instead of treating the socket as stale
+    // and silently spawning a second daemon over the live process.
+    if let Some(daemon) = connect_existing_stateful(socket_path).await? {
         return Ok(daemon);
     }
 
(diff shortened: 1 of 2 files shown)


rjwittams:
Both points taken as fixes, pushed:

1. **Bounded handshake**: connect_existing_stateful now wraps the stateful Hello in a 5s tokio timeout (HELLO_HANDSHAKE_TIMEOUT). Expiry is a hard error naming the socket ('accepted the connection but did not complete the Hello handshake … it may be wedged'), never Ok(None) — so a stalled daemon can neither hang the CLI/TUI nor be spawned over. New test connect_or_spawn_reports_wedged_daemon_instead_of_hanging_or_respawning pins it with an accept-and-never-reply listener.

2. **Spawn-wait loop**: handshake errors during the post-spawn poll are now retried until the 10s deadline (the daemon there was spawned by us, so errors are overwhelmingly startup races); if the deadline expires the *last specific error* is surfaced instead of the generic timeout message, so a genuine problem still reads as itself.

rjwittams:
Fixed the remaining probe site (the post-lock-wait connect). One deliberate deviation from the suggested 'warn + continue' shape: falling through after exhausted retries would reach the **spawn** path, which deletes the socket — so a persistently-erroring live daemon (e.g. genuine version mismatch on the other process's daemon) would get spawned over, reopening the exact bug this PR fixes. Instead: a handshake error spends a retry attempt with a warn (absorbing the startup race, as the retry loop intended), and on exhaustion the branch returns the last handshake error **iff the final probe found a listener that kept failing** — spawn only happens when the last evidence says nothing is listening. Line 368's last-chance probe keeps its early-return, per your note that surfacing the specific error there is an improvement.

rjwittams:
Round-3 points, all four addressed:

1. **First-probe asymmetry — deliberate, now documented in code.** The internal probes retry because they sit in windows where a just-spawned daemon is expected (an error is probably a startup race); the first probe faces a steady-state daemon, where a failed 5s-bounded handshake is a condition the caller should hear immediately — and it's the interactive path (TUI startup, every CLI noun), so retries would only delay an error the user must act on anyway. Comment added at the probe site.

2. **Stateless-connect CLI paths → filed as #716.** Agreed it deserves fixing as its own slice: route run_status/run_watch/run_replica_snapshot through the stateful handshake, and reconsider whether stateless `connect` stays public once only tests use it.

3. **Lock-wait retry test — declining for now.** Exercising that branch requires choreographing flock acquisition across the spawn-lock (a second holder acquiring and releasing at the right moment) plus a listener that fails one handshake then recovers — heavy fixture machinery for straight-line control flow whose interesting property (never spawn when the last probe found a listener) is enforced structurally by the early return. If it regresses in practice it'll surface as #716-style diagnostics, not silent damage.

4. **Soft deadline — acknowledged in the comment at the poll loop** (worst case = deadline + handshake timeout). Left the behaviour as-is: tightening it means threading remaining-time through the helper for a bound nobody depends on.

rjwittams:
Confirmed — real bug, and your analysis of the mechanism was exactly right: `continue` re-entered `acquire_spawn_lock` against a lock the releasing winner had left uncontended, so the retry 'budget' was an illusion and the exhaustion check was dead code in the two-process case. My earlier claim that the property was structurally enforced was wrong.

Fixed with direction (a), plus the test this branch should have had all along:

- **Race-free final probe under the held lock** (before `remove_file`/`spawn_daemon`): a daemon found there returns `Ok`; a handshake error returns hard ('won the spawn lock, but a live daemon is on the socket and failing the handshake — not spawning over it'); only `Ok(None)` proceeds to spawn. Because the lock is exclusively held, nothing can race the check. This also closes the narrower *pre-existing* first-probe→lock-win→spawn window that existed before this PR.
- **`connect_or_spawn_never_spawns_over_daemon_that_appeared_during_lock_contention`** plays out your exact scenario: the test acts as process B (holds the spawn lock, binds a stale-version listener after A's first probe finds nothing, releases the lock); A must surface the version-mismatch error and — asserted directly — B's socket must still exist, i.e. the spawn path was never reached.

PR #713, opened 2026-07-13
aggregator: avoid repeated session scans for attach capabilities

### LINKED ISSUE #700: aggregator: avoid repeated session scans for attach capabilities
From PR #698 review by @claude[bot]: https://github.com/flotilla-org/flotilla/pull/698#issuecomment-4961119484

The sessions query currently computes each row's attach capability by calling `resolve_attach_command_internal`. That path lists and scans both durable and observed terminal-session stores before resolving the matched session's environment and terminal pool.

During aggregator bootstrap and environment-driven recomputation, doing this once per session makes projection work quadratic in the number of sessions. Current session counts make this acceptable, but the projection should avoid repeated full-store scans as fleets grow.

Investigate a batch or direct-resource attach-capability API that preserves the invariant that `SessionRow.attach: Some` means the daemon can resolve the reference, while listing session stores at most once per projection batch.

Context: follow-up from #683; intentionally kept out of PR #698.

rjwittams:
## What changed

- build a reusable attach-candidate index from durable and observed terminal sessions
- resolve all projected session attach capabilities through one batch call per aggregator rebuild
- preserve exact and prefix matching, ambiguity handling, replica routing, and early blank-reference validation
- add coverage proving multiple projected rows use a single batch resolver invocation

## Why

`Aggregator::summarize_session` previously resolved each session independently. Every resolution listed both session stores and scanned all candidates, making a projection rebuild O(n²).

The batch API lists each store once and reuses the resulting index across all projected session references.

## Impact

Running session rows continue to expose `attach: Some` when resolvable, while projection cost now scales with a single store scan per rebuild.

Closes #700.

## Validation

- parallel Standards and issue-spec reviews, with all findings addressed

### DIFF
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -282,6 +282,68 @@ fn fleet_row_attach_reference_label(row: &FleetListRow) -> String {
     }
 }
 
+enum AttachTarget {
+    Local(Box<flotilla_resources::ResourceObject<ResourceTerminalSession>>),
+    Replica { host: HostName },
+}
+
+impl AttachTarget {
+    async fn resolve(&self, daemon: &InProcessDaemon, reference: &str) -> Result<String, String> {
+        match self {
+            Self::Local(session) => daemon.attach_command_for_session(reference, session).await,
+            Self::Replica { host } => daemon.recursive_attach_command_for_remote(host, reference).await,
+        }
+    }
+}
+
+struct AttachCandidate {
+    label: String,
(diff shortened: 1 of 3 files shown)


rjwittams:
Addressed both readability nits in d282c53f: renamed the shadowed inner prefix variable to candidate_reference and documented why the public blank-reference guard intentionally runs before candidate-index construction. Pinned formatting, workspace Clippy, all attach-related core tests, and the focused aggregator batching regression test pass locally.

PR #715, opened 2026-07-13
Disambiguate command subjects with @ addressing

### LINKED ISSUE #697: Disambiguate command verbs and nested nouns from positional subjects
> *This was generated by AI during triage.*

## Problem

Flotilla's composable command grammar is:

```text
[target] noun [subject] verb [verb-specific args]
```

Noun groups that support both subjectless verbs and positional subjects currently let Clap choose between those interpretations with `subcommand_precedence_over_arg`. A valid subject whose text matches a verb (or, at a routing layer, a nested noun) can therefore become unaddressable.

The original crew-specific report was only partly accurate: `complete` and `fail` roles are already addressable when followed by `handoff`; `list` is explicitly rejected. The same structural problem is clearer for subjects such as a checkout branch named `status`, a host named `status`, or a repo selector named `refresh`.

This is a shared command-grammar bug, not a crew-only special case. It relates to #477.

## Decision

Keep the composable subject-before-verb grammar and continue using Clap.

Add an explicit subject address marker:

```text
@<name>
```

A marked token is always a subject, never a verb or nested noun.

Examples:

```text
flotilla crew complete
flotilla crew @complete handoff --message "please continue"
flotilla crew @list handoff --message "please review"

flotilla host @status status
flotilla host @checkout repo @refresh work
flotilla checkout @status status
```

Ordinary non-colliding subjects remain unchanged.

Leading `@` is reserved for Flotilla-owned names, including workflow vessel and crew role names, and validation must reject such declarations. Opaque external identifiers may legitimately begin with `@`; noun groups using the marker must therefore also accept `--subject <literal>` as an unambiguous literal fallback.

## Acceptance criteria

- [ ] `crew @list handoff`, `crew @complete handoff`, and `crew @fail handoff` target those literal roles while unmarked `crew list`, `crew complete`, and `crew fail` retain their command meanings.
- [ ] Ambiguous host, repo, checkout, issue, and environment subjects can be forced with `@<name>` wherever those noun groups mix positional subjects with subjectless verbs or routing.
- [ ] Those noun groups accept `--subject <literal>`, mutually exclusive with the positional subject, without interpreting a leading `@` in the flag value.
- [ ] Nested routing remains composable, including a routed target or inner subject whose name collides with a verb or noun.
- [ ] Command display/parse round-trips preserve marked and explicit-subject forms.
- [ ] Completion surfaces render colliding dynamic subjects in their addressable `@<name>` form.
- [ ] Relevant parse errors explain the `@<name>` or `--subject` escape.
- [ ] Workflow validation rejects vessel and crew role names beginning with the reserved `@` prefix.
- [ ] Focused parser, completion, palette, and workflow-validation tests cover collisions and literal external identifiers beginning with `@`.

## Out of scope

- Reordering the grammar to verb-before-subject.
- Replacing Clap.
- Restricting opaque external identifier namespaces such as Git branch names.
- Backward-compatibility or migration handling for previously accepted workflow templates.

rjwittams:
## Summary

- preserve Flotilla's subject-before-verb grammar while allowing `@name` to force a positional subject
- add `--subject <literal>` for opaque external identifiers that legitimately begin with `@`
- apply the shared disambiguation to crew, host, repo, checkout, issue, and environment commands, including nested routing
- make completions and TUI palette prefills emit addressable subjects and explain the escape syntax in relevant parse errors
- reserve leading `@` in workflow vessel and crew-role names

## Root cause

Noun groups with both optional positional subjects and subjectless verbs rely on Clap's subcommand precedence. A subject whose text matches a verb or nested noun is therefore parsed as command grammar instead of data and becomes unaddressable.

## User impact

Ordinary commands remain unchanged. Colliding names can now be addressed explicitly, for example:

```text
flotilla crew @complete handoff --message "please continue"
flotilla checkout @status status
flotilla host @checkout repo @refresh work
```

External names beginning with `@` use the literal fallback:

```text
flotilla checkout --subject @topic status
```

## Validation

- independent standards and issue-spec review

Closes #697

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1231,6 +1231,7 @@ dependencies = [
 name = "flotilla-commands"
 version = "0.1.0"
 dependencies = [
+ "bon",
  "clap",
  "flotilla-protocol",
  "tempfile",
--- a/crates/flotilla-commands/Cargo.toml
+++ b/crates/flotilla-commands/Cargo.toml
@@ -6,6 +6,7 @@ license.workspace = true
 
 [dependencies]
 flotilla-protocol = { path = "../flotilla-protocol" }
+bon = { workspace = true }
 clap = { version = "4", features = ["derive", "string"] }
 
 [dev-dependencies]
--- a/crates/flotilla-commands/src/commands/checkout.rs
+++ b/crates/flotilla-commands/src/commands/checkout.rs
@@ -5,6 +5,7 @@ use flotilla_protocol::{CheckoutSelector, CheckoutTarget, Command, CommandAction
 
 use crate::{
     resolved::{HostResolution, RepoContext},
+    subject::{resolve_subject, write_subject},
     Resolved,
 };
 
@@ -13,7 +14,11 @@ use crate::{
 #[command(subcommand_precedence_over_arg = true, subcommand_negates_reqs = true)]
 pub struct CheckoutNoun {
     /// Branch name or checkout path
+    #[arg(value_name = "SUBJECT", conflicts_with = "explicit_subject")]
     pub subject: Option<String>,
+    /// Literal subject; use for external names beginning with `@`
+    #[arg(long = "subject", value_name = "SUBJECT", conflicts_with = "subject")]
+    pub explicit_subject: Option<String>,
 
     #[command(subcommand)]
     pub verb: Option<CheckoutVerb>,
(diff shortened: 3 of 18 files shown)


rjwittams:
Addressed the performance and synchronization points in 7d639647: the root and host Clap command trees are now cached in OnceLock values, so palette candidates reuse metadata, and the crew pseudo-command set is centralized for both resolution and completion. I kept the CrewNoun bon::Builder derive because CLAUDE.md's documented convention explicitly requires #[derive(bon::Builder)] for types with more than three fields or many optional fields; CrewNoun has nine mostly-optional fields, so removing it would violate the repository standard even though current construction is primarily through Clap.

rjwittams:
Fixed the issue-prefill gap in 6ed7a750. palette_prefill now routes issue keys through address_subject_for_cli(SubjectNoun::Issue, ...), and a regression test verifies that an issue key named open produces issue @open. Focused command-palette tests and TUI clippy pass locally.

PR #718, opened 2026-07-13
fix: make forced work completion atomic

### LINKED ISSUE #704: Make forced work completion validation atomic with status patch
From PR #703 review by @claude: https://github.com/flotilla-org/flotilla/pull/703#issuecomment-4961231460

Forced vessel-work completion currently validates the observed `WorkPhase`, then calls `apply_resource_status_patch`. That helper re-fetches and reapplies the infallible `ForceWorkCompleted` patch on optimistic-concurrency conflicts. If reconciliation changes the work to `Failed` or `Cancelled` between validation and a retry, the retry can overwrite that terminal phase with `Complete`.

Make terminal-state validation atomic with the status mutation. This likely requires either a condition-aware/fallible status patch or another compare-and-set shape that revalidates against every fetched version.

Context: #694 aligns the Aggregator capability with command validation for the ordinary path. This follow-up covers the narrower concurrency guarantee without expanding that PR into status-patch API design.

rjwittams:
## Summary

- add a checked status-patch helper that revalidates each freshly fetched resource version
- use it for forced convoy-work completion so optimistic-concurrency retries cannot overwrite terminal work
- preserve rejection of every terminal phase through the shared `WorkPhase::is_terminal()` rule
- add a deterministic in-memory conflict test covering the race

## Root cause

Forced completion validated a convoy once, then called the retrying status-patch helper. If reconciliation changed the work to `Failed` or `Cancelled` after validation, a conflict retry re-fetched that terminal status and blindly reapplied `ForceWorkCompleted`, overwriting it with `Complete`.

## Impact

The validation now runs inside the same resource-version retry loop as the mutation. A concurrent status change either makes the versioned write conflict and trigger revalidation, or the write succeeds against exactly the version that was validated.

## Validation

- parallel standards and spec reviews against issue #704

Fixes #704

### DIFF
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -26,14 +26,15 @@ use flotilla_protocol::{
     StatusResponse, StreamKey, SystemInfo, ToolInventory, TopologyResponse, TopologyRoute,
 };
 use flotilla_resources::{
-    apply_status_patch as apply_resource_status_patch, external_patches as convoy_external_patches, terminal_session_attach_target,
-    Checkout as ResourceCheckout, CheckoutPhase as ResourceCheckoutPhase, CheckoutSpec as ResourceCheckoutSpec,
-    CheckoutStatus as ResourceCheckoutStatus, Convoy as ResourceConvoy, ConvoyRepositorySpec, ConvoySpec, ConvoyStatusPatch, CrewSource,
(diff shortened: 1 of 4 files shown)


rjwittams:
Added a short comment explaining that the synchronous barrier intentionally blocks one Tokio worker while the second worker lands the concurrent status update. The focused status-patch tests and pinned format check still pass.

Issue #708, opened 2026-07-13
Build the manifest metadata-patch connector v0 (flotilla pm connect + attach pane stamping)


rjwittams:
## Parent

#604 · Design: #667 (grilled 2026-07-11, currency-checked 2026-07-13 — design doc: `project-map/reports/manifest-connector-design-2026-07-11.html`). All prerequisites landed (#680, #682, #683, gate #672); worktree `manifest-metadata-patch-connector` exists on current main.

## What to build

The manifest metadata-patch connector v0, per the #667 design:

- **Half 1 — attach-time pane stamping:** `flotilla attach`, when running inside a PM, publishes pane → identity (`flotilla.session` = `<host>/<ns>/<session>` canonical join key + denormalized ids, no TTL) into the PM's metadata plane before exec.
- **Half 2 — the catalog:** `flotilla pm connect`, one per PM instance, PM-adjacent (never a flotillad facility). Dials the local daemon as a client, subscribes (`SubscribeQueries`) to the `convoys` and `sessions` named queries, and projects the fleet-merged rows into Group/Identity-targeted metadata patches: project → convoy → vessel hierarchy, latent entities included, phase/attention badges, recipes. TTL ≈30s, re-asserted ≈10s.

v0 binds to andamento's **current** pipe names and key spellings (`andamento-apply-metadata-patch`, `tab.*`, `git.repo`) — deliberate deviation from bind-after-Leg-1, recorded on #667. All spellings live in one wire module so the Leg-1 rename is one file.

## Dependency direction (decided 2026-07-13)

The wire module **mirrors** `andamento-shared`'s patch serde for v0 (`MetadataPatch`/`MetadataTarget`/`MetadataValueUpdate`), verified by fixture round-trip tests against andamento's actual JSON. This mirror is a stopgap, not the architecture:

- **End state:** andamento, wheelhouse/uishell, **and flotilla** all depend on the shared `flotilla-org/manifest` crate once the manifest extraction convoy's Leg 1 creates it. PM cores must not depend on flotilla; flotilla depending on a PM's crate (`andamento-shared`) is even more backwards. The shared crate holds the non-flotilla-specific things: patch types, targets, store/resolver semantics, transport adapters.
- **Stays in flotilla:** everything stringly flotilla-specific — the `flotilla.*` key spellings, the projection from result-set rows, recipe templates. These depend on flotilla metadata by construction and belong here.
[...]

PR #720, opened 2026-07-13
feat: daily-driver layout running the manifest connector

rjwittams:
Adds `layouts/daily-driver.kdl` — andamento's daily layout composed with one extra pane running `flotilla pm connect --flotilla-bin $FLOTILLA_BIN` (the git-watcher's seat, per the #667 design) — and `scripts/run-daily-driver.sh` which builds the andamento plugins, warns when the dev binaries are ad-hoc signed (#689), and execs zellij with the layout.

Kept on the flotilla side deliberately: andamento's own `layouts/andamento.kdl` stays flotilla-free, preserving the git-watcher as the manifest independence proof.

This is the acceptance rig for #708's daily-driver checklist.

### DIFF
--- a/layouts/daily-driver.kdl
+++ b/layouts/daily-driver.kdl
@@ -1,0 +1,81 @@
+// Daily-driver layout: zellij + andamento + the flotilla manifest connector.
+//
+// This is the composed layout the PM-capability daily driver runs
+// (flotilla-org/flotilla#708). It is derived from andamento's own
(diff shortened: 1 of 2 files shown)


rjwittams:
Fixed in 386e0259. The macOS-specific signing warning loop now runs only when codesign is available, so Linux hosts no longer get a misleading ad-hoc-signing warning.

PR #721, opened 2026-07-13
fix: migrate legacy repo config filenames

### LINKED ISSUE #691: repo remove misses configs saved under the legacy filename scheme — untracked repos resurrect on restart
## What happened

`~/.config/flotilla/repos/` contains files in two naming schemes: URL-encoded paths (`%2FUsers%2Frobert%2Fdev%2Fghostty.toml`) and an older slugified scheme (`users-robert-dev-facefacts.toml`). Both load fine at daemon startup, but `flotilla repo remove` only deletes the file under the current (URL-encoded) name.

Observed 2026-07-13 while pruning tracked repos (#689 mitigation): removing four repos updated `tab-order.json` and the running daemon for all four, but only kitty-image-tests (URL-encoded config) lost its file — FaceFacts, flotilla-site, and rjw-cc kept their legacy-scheme tomls, so a daemon restart would silently re-track them. Cleaned up by hand this time.

## What I expected

Untracking a repo removes its persisted config regardless of which filename scheme it was saved under.

## Fix shape (no-compat phase)

Either delete by matching the `path` field across all files in `repos/` rather than by computed filename, or migrate legacy-named files to the current scheme at load time (rewrite + delete old). Given the no-backwards-compatibility phase, load-time migration is probably the cleanest way to make the scheme single again. Three legacy-scheme files remain live locally (cleat, flotilla, reticulate) — any load-time migration gets exercised immediately.

rjwittams:
## Summary

- canonicalize parseable repo config filenames to the URL-encoded scheme during startup
- preserve complete per-repo TOML contents while migrating and keep canonical files authoritative when duplicates exist
- deduplicate loaded repo paths and remove the legacy slug implementation from production code

## Root cause

Repo loading trusted the `path` field in every TOML regardless of filename, while repo removal deleted only the current URL-encoded filename. A repo stored under the legacy slug filename therefore returned after daemon restart, and its per-repo overrides were invisible to canonical lookup.

## Impact

Untracking a migrated repo now removes its only persisted config, so it cannot silently resurrect after restart. Existing checkout overrides survive migration byte-for-byte.

## Validation

- two-axis standards/spec review against issue #691

Closes #691


rjwittams:
Added a structured info log when migration discards a duplicate legacy config, including both the legacy and canonical paths. The degenerate collision order remains intentionally unchanged: an existing canonical config is authoritative, and loaded repo paths are sorted and deduplicated.

PR #722, opened 2026-07-13
fix: delete dead repo root resolver

### LINKED ISSUE #707: Delete (or re-wire) the dead resolve_repo_roots startup path
From PR #706 review by claude-review: https://github.com/flotilla-org/flotilla/pull/706#issuecomment-4962186926

`flotilla_core::config::resolve_repo_roots` has no callers anywhere in the workspace — its startup role was replaced by direct `TrackRepoPath` dispatch before #706, and #706 edited its body (removed the tab-order write) without noticing it was dead.

Deleting it pulls a thread worth deciding deliberately:
- It was the only consumer of the legacy `load_tab_order` **ordering** (persisted-order-first repo registration). With it gone, daemon repo order comes from `load_repos()` (alphabetical) + CLI args + cwd detection — decide whether that ordering story matters post-ADR-0013 (tab order is Surface-owned now; registration order only affects the overview listing and first-run seeding).
- If it goes, `load_tab_order` keeps exactly one purpose (seeding `open-views.toml` order on migration — currently unused since seeding reads daemon list order) and can likely be deleted with it, taking `tab-order.json` support out entirely.

Small PR: delete `resolve_repo_roots`, decide the `load_tab_order` / `tab-order.json` end-of-life, adjust config tests.

rjwittams:
## Summary

- remove the dead `resolve_repo_roots` path and legacy `tab-order.json` loader
- preserve no-argument current-repository discovery in the live startup `TrackRepoPath` path
- make explicit CLI repository roots intentionally take precedence over cwd discovery
- update active configuration docs to name `open-views.toml`

## Root cause

Embedded-mode removal left repository-root resolution disconnected from startup. The legacy tab-order loader also no longer seeded any live view state, so both paths had become dead while the useful current-working-directory fallback had been lost.

When explicit roots are supplied, startup tracks exactly those roots. Cwd discovery is only a no-argument fallback, avoiding an unexpected extra repository alongside an explicit selection.

## Validation

- `cargo test --workspace --locked` (GitHub Actions)
- focused startup repository selection tests
- `cargo test -p flotilla-core --locked config::tests`

All GitHub Actions checks pass. The workspace suite also passed locally before the final rebase; a post-rebase daemon timeout under concurrent local Rust build load did not reproduce in CI.

Closes #707

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -188,7 +188,7 @@ Key bindings are configurable via TOML. Defaults: `j/k` navigate, `Enter` execut
 
 Stored in `~/.config/flotilla/`:
 - `repos/*.toml` — one per tracked repo (`path = "..."`, plus per-provider config: `change_request`, `issue_tracker`, `cloud_agent`, `ai_utility`, `workspace_manager`, `terminal_pool`, `vcs.git`)
-- `tab-order.json` — array of repo paths
+- `open-views.toml` — ordered TUI-owned View addresses and label overrides
 - `keybindings.toml` — user key binding overrides
 
 Workspace templates: `.flotilla/workspace.yaml` in repo root.
--- a/crates/flotilla-core/src/config.rs
+++ b/crates/flotilla-core/src/config.rs
@@ -1,7 +1,7 @@
 use std::{
     collections::{BTreeMap, HashMap},
     path::{Path, PathBuf},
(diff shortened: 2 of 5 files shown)


rjwittams:
Fixed in 3526244a. Explicit roots are now canonicalized and deduplicated in argument order, with a regression test covering repeated CLI paths.

PR #723, opened 2026-07-13
fix: require hello handshake for socket connections

### LINKED ISSUE #716: run_status / run_watch / run_replica_snapshot bypass the Hello handshake — version mismatch undiagnosable on those paths
From PR #712 review by claude[bot]: https://github.com/flotilla-org/flotilla/pull/712#issuecomment-4963059400 (point 2).

`run_status` and `run_watch` (`crates/flotilla-tui/src/cli.rs:500`, `:626`) and `run_replica_snapshot` (`src/main.rs:459`) call the stateless `SocketDaemon::connect` directly. They don't spawn on failure, so they don't hit the spawn-over-live-daemon bug #712 fixed — but they also never exchange protocol versions, so against a stale/incompatible daemon they fail with whatever confusing error falls out of cross-version wire traffic instead of the clear "daemon protocol version mismatch … restart the daemon" diagnostic.

Route them through the stateful handshake (`connect_stateful` / `connect_existing_stateful`-style probing) so every entry point reports version mismatch as itself. Consider whether the stateless `SocketDaemon::connect` should remain public at all once no production path uses it (tests in `flotilla-daemon/tests/socket_roundtrip.rs` still do).

rjwittams:
## Summary

- make `SocketDaemon::connect` perform the bounded stateful Hello handshake by default
- remove the opt-in `connect_stateful` socket constructor so production socket callers cannot bypass version validation
- update real-socket client tests to negotiate Hello and cover a stale daemon version directly

## Why

`status`, `watch`, and `replica-snapshot` all used the public socket constructor without exchanging protocol versions. Against an incompatible daemon they could therefore fail with confusing cross-version wire errors instead of the existing actionable version-mismatch diagnostic.

The public socket boundary now guarantees a compatible, responsive daemon before normal request handling begins, while retaining non-spawning behavior for these commands.

Closes #716.

## Validation

- `cargo test -p flotilla-client --locked`
- `cargo test -p flotilla-daemon --locked --test socket_roundtrip`
- independent Standards and Spec review against #716

### DIFF
--- a/crates/flotilla-client/src/lib.rs
+++ b/crates/flotilla-client/src/lib.rs
@@ -98,22 +98,13 @@ pub struct SocketDaemon {
 impl SocketDaemon {
     /// Connect to a running daemon at the given Unix socket path.
     ///
-    /// Builds a session from the socket and then starts the shared client
-    /// reader/pending-request machinery on top of it.
+    /// Requires a bounded, stateful Hello handshake before starting the shared
+    /// client reader/pending-request machinery. This makes protocol-version
+    /// validation part of every real socket connection rather than an opt-in
+    /// call-site choice.
     pub async fn connect(socket_path: &Path) -> Result<Arc<Self>, String> {
         let session = connect_unix_message_session(socket_path).await?;
-        Self::from_session(session)
-    }
-
-    /// Connect to a running daemon with a stateful Hello handshake.
-    ///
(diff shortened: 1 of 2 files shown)


rjwittams:
Reviewed — thank you. Confirmed there are no actionable findings; the bounded public connection handshake and non-spawning caller behavior are intentional.

PR #724, opened 2026-07-13
Fix Zellij workspace creation focus race

### LINKED ISSUE #709: zellij workspace creation races on session focus — write-chars/new-pane land in the wrong pane/tab
## What happened

Creating a workspace via the zellij presentation manager (Plane A) races on the session's *current focus*. Observed by Robert 2026-07-13/14, from the flotilla TUI inside zellij:

- the first pane's command was **typed into the flotilla TUI's own command palette** (the attach command appeared as palette input)
- subsequent panes opened **in the wrong tab**

## Mechanism (confirmed in code — `providers/presentation/zellij.rs`, `create_workspace`)

After the qualified `new-tab --name --cwd` (which returns a tab id), every following step is **focus-relative against session-global state**:

1. `write-chars` (first pane's command) types into whatever pane is focused *right now* — if focus has returned to the TUI's pane, the command lands in the TUI (palette open ⇒ palette input). `write-chars` has no target argument; it can only ever be focus-relative.
2. `new-pane` (all further panes/stacks) creates relative to the focused tab/pane — focus moved ⇒ wrong tab.
3. Final focus positioning is a counted `focus-previous-pane` walk — same dependence.
4. Sequencing is `sleep(150ms)` / `sleep(100ms)` between actions: timing, not ordering. Any focus change in the window — the user typing, andamento, the TUI itself, a concurrent flotilla command — redirects the remaining sequence.

## Does `new-tab --layout` keep the rail? — VERIFIED: no (tiled), with a floating nuance

Neither assumed nor guessed — read from the zellij source (the `rjwittams/zellij` fork at `~/dev/zellij`) and andamento's layouts:

- `zellij-server/src/plugins/mod.rs:538`: `tab_layout = tab_layout.or_else(|| Some(layout.new_tab().0));` — the session's `new_tab_template` (`Layout::new_tab()`, `zellij-utils/src/input/layout.rs:1695`, returns `self.template`) is consulted **only when no layout is passed**. A CLI-provided layout **replaces** the tiled template wholesale; there is no composition.
- Floating nuance, same handler: `if floating_panes_layout.is_empty() { layout.new_tab().1 }` — the template's **floating** panes *do* apply when the CLI layout defines none.
- `~/dev/andamento/layouts/andamento.kdl:35-45` (and `-native`): the rail (`plugin location="andamento-rail"`, 23% vertical) and the status bar are **tiled** panes in `new_tab_template`.

**Conclusion: in an andamento session, `new-tab --layout` produces a tab with no rail and no status bar.** The atomic-layout fix is ruled out as-is.

## Template-splicing IS doable — verified (this changes the ranking)

Robert's condition — "possible iff you take the default tab layout *from the running zellij* and fill it in" — holds. Verified chain in the zellij source:

- `zellij action dump-layout` → `ScreenInstruction::DumpLayout` (`zellij-server/src/screen.rs:7273`) → `get_layout_metadata` → `SessionLayoutMetadata::new(self.default_layout.clone())` (`screen.rs:5839`) — the **running session's** layout, template included.
- The serializer emits the template into the dumped KDL: `serialize_new_tab_template(global_layout_manifest.default_layout.template, ...)` (`zellij-utils/src/session_serialization.rs:64,496`), pinned by test `can_serialize_new_tab_template` (line 2103).

## Fix space (grounded, re-ranked)

1. **Recommended fix: dump → splice → atomic layout.** Per workspace creation: `dump-layout` from the live session → parse KDL → extract `new_tab_template` → replace its content slot with the workspace's pane tree (rule: the first empty non-plugin `pane` — exactly the slot zellij itself fills with the shell on a plain `new-tab`; andamento's template has exactly one) → `zellij action new-tab --layout <generated> --name --cwd`. Atomic, no `write-chars`, no focus-relative steps, no sleeps — and **no coupling**: the template comes from whatever session is running (any PM, bare zellij, custom user templates). Sessions with *no* template dump none → splice degrades to using the workspace tree bare, which is exactly right. Fixture-test the dump→parse→splice→emit round-trip against andamento's actual template.
2. **Fallback / independent hardening: eliminate `write-chars` regardless.** Even if (1) hits a snag (dump-layout precision, multi-slot templates), running the first pane's command as a `new-pane -- sh -c <cmd>` instead of typed keystrokes kills the palette-typing failure class on its own; `go-to-tab-by-id` re-assertion before each `new-pane` shrinks (not removes) wrong-tab.
3. **Fork options** (`rjwittams/zellij`): qualified targets (`new-pane --tab-id`) or a native template-composing `new-tab --layout-in-template`. No longer required for correctness — (1) achieves it CLI-only — but the native flag would let zellij own the splice rule instead of clients reimplementing it; decide with andamento's long-term interests, upstream if wanted.

Plane A bugfix (allowed under the freeze). Worth fixing before the #708 connector makes zellij workspace creation a more-travelled path.

rjwittams:
Closes #709.

## What changed

- Start the first workspace command as part of `zellij action new-tab`, preserving Zellij's configured `new_tab_template`.
- Target every subsequent pane with the stable tab ID returned by `new-tab`, and restore requested pane focus with stable pane IDs.
- Remove relative `write-chars`, focus walking, and timing sleeps from workspace creation.
- Require Zellij 0.44.1, the first version with the complete stable-target CLI used here.
- Add deterministic command-boundary tests and re-record the live Zellij replay fixture.

## Why

Workspace creation previously depended on whichever tab or pane happened to be focused between separate actions. A user focus change could therefore send the first command into Flotilla's command palette and create later panes on another tab. Supplying the initial command to `new-tab` and qualifying all later mutations removes that race without replacing the native Andamento tab template.

## Impact

Zellij workspaces are now created deterministically on one tab. The first command starts in the intended terminal pane, later panes cannot be redirected by focus changes, and native Andamento rail/status plugins remain present.

## Validation

- `RUSTC_WRAPPER= cargo clippy --workspace --all-targets --locked -- -D warnings`
- `RUSTC_WRAPPER= cargo test --workspace --locked`
- Native Andamento Zellij built with `scripts/build-native-andamento --locked --bin zellij`; live tab creation retained rail/status plugins and stable `--tab-id` pane placement.

### DIFF
--- a/crates/flotilla-core/src/providers/discovery/factories/zellij.rs
+++ b/crates/flotilla-core/src/providers/discovery/factories/zellij.rs
@@ -62,7 +62,7 @@ mod tests {
         let bag = EnvironmentBag::new().with(EnvironmentAssertion::env_var("ZELLIJ", "0"));
         let dir = tempfile::tempdir().expect("failed to create tempdir");
         let config = ConfigStore::with_base(dir.path());
-        let runner = Arc::new(DiscoveryMockRunner::builder().on_run("zellij", &["--version"], Ok("zellij 0.42.2".into())).build());
+        let runner = Arc::new(DiscoveryMockRunner::builder().on_run("zellij", &["--version"], Ok("zellij 0.44.1".into())).build());
         let result = ZellijPresentationManagerFactory.probe(&bag, &config, &ExecutionEnvironmentPath::new("/repo"), runner).await;
         assert!(result.is_ok());
     }
@@ -105,7 +105,7 @@ mod tests {
             .with(EnvironmentAssertion::env_var("ZELLIJ_SESSION_NAME", "my-session"));
         let dir = tempfile::tempdir().expect("failed to create tempdir");
         let config = ConfigStore::with_base(dir.path());
-        let runner = Arc::new(DiscoveryMockRunner::builder().on_run("zellij", &["--version"], Ok("zellij 0.42.2".into())).build());
+        let runner = Arc::new(DiscoveryMockRunner::builder().on_run("zellij", &["--version"], Ok("zellij 0.44.1".into())).build());
         let result = ZellijPresentationManagerFactory.probe(&bag, &config, &ExecutionEnvironmentPath::new("/repo"), runner).await;
         assert!(result.is_ok(), "factory should succeed when ZELLIJ_SESSION_NAME is set");
     }
--- a/crates/flotilla-core/src/providers/presentation/fixtures/zellij_workspaces.yaml
+++ b/crates/flotilla-core/src/providers/presentation/fixtures/zellij_workspaces.yaml
@@ -9,19 +9,13 @@ rounds:
     - feat-123
     - '--cwd'
     - /tmp
+    - '--'
+    - sh
+    - '-c'
+    - '{main_command}'
     cwd: '.'
     stdout: |
       1
-    exit_code: 0
-  - channel: command
-    cmd: zellij
-    args:
-    - action
-    - write-chars
-    - |
-      {main_command}
-    cwd: '.'
-    stdout: ''
     exit_code: 0
   - channel: command
     cmd: zellij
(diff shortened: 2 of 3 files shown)


rjwittams:
Thanks for the detailed verification. The live native validation supplied the JSON shape used by the focused-initial-pane unit test; I filed #725 to add a dedicated REPLAY=record contract fixture for that list-panes path. No code changes are needed for the non-blocking parse_pane_id observation.

PR #726, opened 2026-07-14
fix: keep attachable set identity stable

### LINKED ISSUE #719: AttachableSets leak and split: checkout's set id ≠ the live set for the same workspace; orphan empty sets accumulate
## What happened

Observed by Robert 2026-07-14 in `flotilla repo flotilla-org/flotilla work --json` (Plane A): attachable sets look like they're leaking. Two distinct symptoms in the output:

1. **Every fresh worktree's Checkout row carries an `attachable_set_id` that does not match the AttachableSet item of the same workspace name.** E.g. the connector checkout points at `c86eceea-…` while the set named `flotilla.manifest-metadata-patch-connector` — the one actually holding the 3 live terminals — is `5c4a345f-…`. Same for all five current worktrees (atomic-work-completion-validation, fix/workspace-focus-race, fix/untracked-repos-legacy-config, fix/dead-resolve-repo-roots, skip-hello-handshake-version-mismatch). Because the ids differ, the `AttachableSet` correlation key never joins them: the sets float as separate work items with no branch/issue association, and the checkout rows show 0 terminals despite their workspaces running 3 each.
2. **Orphan sets accumulate**: an anonymous set (`ab08d14e-…`, no description, 0 terminals) and `flotilla.mh-tests-2` (0 terminals) linger with nothing attached.

## Mechanism pointers (`crates/flotilla-core/src/attachable/store.rs`)

- `ensure_terminal_set_with_change` (line ~177) dedupes by the exact tuple `(host_affinity, checkout, environment_id)` and mints a fresh UUID otherwise. Two callers describing the same workspace with different tuples (e.g. one passing `checkout: Some(path)`, another `checkout: None` + host affinity, or differing `environment_id`) get **two sets**.
- Ids are fresh UUIDs per allocation; anything that persists a set id (workspace refs, resources, template identity) keeps it across daemon restarts while a restarted in-memory registry re-mints — persisted-vs-live divergence is the other candidate for the same symptom.
- Empty sets are never garbage-collected once their members go.

Diagnosis should establish which callers create the checkout-linked set vs the terminal-holding set and unify the dedupe key (or thread the set id through instead of re-ensuring); plus a cleanup rule for member-less sets.

## Why it matters now

Plane A attachablesets are the daily-driver workflow while convoys mature (the current 6 parallel worktrees all run through them); the split breaks branch/issue ↔ terminal correlation in the repo views, which is exactly the surface used to keep the parallel agents straight. Plane A bugfix, allowed under the freeze.

rjwittams:
## Summary

- thread the checkout's canonical attachable-set ID through terminal preparation instead of allocating a second hostname-qualified set
- preserve stable checkout identity for normal workspace creation and both existing/fresh teleport flows
- remove unbound attachable sets when their final member moves or is removed, while preserving intentionally empty bound sets
- add regressions covering checkout plans, teleport flows, and both attachable-store implementations

## Root cause

Workspace preparation created a stable HostId-qualified set for the checkout, but terminal preparation reconstructed the checkout from `HostName::local()` and re-ran set allocation. Because set deduplication uses the exact host/check-out/environment tuple, this produced a second set that held the live terminals while the checkout and workspace binding pointed at the first.

## Impact

Checkout rows, terminal sessions, and workspace bindings now share one attachable-set identity, restoring branch/issue-to-terminal correlation and preventing future empty source-set accumulation.

Existing stale registry files are not migrated, consistent with the repository's no-backward-compatibility policy.

## Validation

Closes #719

### DIFF
--- a/crates/flotilla-core/src/attachable/store.rs
+++ b/crates/flotilla-core/src/attachable/store.rs
@@ -357,12 +357,25 @@ impl AttachableStoreState {
     }
 
     fn remove_member_link(&mut self, set_id: &AttachableSetId, attachable_id: &AttachableId) -> bool {
-        if let Some(set) = self.registry.sets.get_mut(set_id) {
+        let changed = if let Some(set) = self.registry.sets.get_mut(set_id) {
             let original_len = set.members.len();
             set.members.retain(|member| member != attachable_id);
-            return set.members.len() != original_len;
+            set.members.len() != original_len
+        } else {
+            false
+        };
+        if changed {
+            let is_empty = self.registry.sets.get(set_id).is_some_and(|set| set.members.is_empty());
+            let is_bound = self
+                .registry
+                .bindings
+                .iter()
+                .any(|binding| binding.object_kind == BindingObjectKind::AttachableSet && binding.object_id == set_id.as_str());
+            if is_empty && !is_bound {
+                self.registry.sets.shift_remove(set_id);
+            }
         }
-        false
+        changed
     }
 
     fn remove_set(&mut self, id: &AttachableSetId) -> Option<RemovedSetInfo> {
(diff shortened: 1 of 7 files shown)


rjwittams:
Addressed both notes in a2df2036: the neighboring file-backed executor tests now use isolated temp config directories, and the attachable-store contract suite now explicitly verifies that a bound source set survives after its final member moves, for both file-backed and in-memory implementations.

Issue #688, opened 2026-07-13
Cross-root semantics for shared identities (projects, repos): scenario-first brainstorm


rjwittams:
## Question

Federation (ADR 0002) is deliberately not CRDT/consensus. So what should a user **observe** when more than one root knows about "the same" thing — the same project, the same repo, work on the same convoy seen from different machines? Today the answer is accidental: `Project` is a plain local resource (`flotilla-resources/src/project.rs`), so two roots can silently hold divergent records for one name and every consumer (convoy create, aggregator fleet merge, the #667 connector's project spine) picks whichever its root has.

## Method ruling (Robert, 2026-07-13)

Two constraints on how this brainstorm is conducted, recorded so they survive until the session:

1. **Scenario-first, not model-first.** Walk through concrete situations and state what you would *expect* to happen in each — then design the data layer to deliver those expectations. Do not evaluate backwards from the replication model to "what would happen in that model" and call the result semantics. Technical considerations must not drive observable behaviour (tail wagging dog).
2. **Layering:** the **Tender is entirely about hosts, transports, and forwarding**. The control-plane data layer — replication semantics, identity, conflict behaviour, however it works — is **fully flotilla's**. This is a flotilla data-layer design, not a Tender requirement; nothing from here goes on #608.

## Seed scenarios (expectations to be written at the session; extend freely)

1. Create project `flotilla` on kiwi. Later, on feta, run `flotilla convoy create --project flotilla`. What should happen?
2. Both roots independently created a project named `flotilla` (accident, weeks apart, different `repositories` lists). What should `ls` / panels / convoy create show and do on each root?
3. Deliberate two-machine life: laptop (kiwi) + workstation (feta), one project, editing its config from whichever machine you're on — including while the laptop is off-network. Reconnect: expectation?
4. Edit divergence in one window: `display_name` changed on kiwi, `repositories` changed on feta. Expectation on reconnect?
5. Delete a project on one root. What do other roots show, immediately and later?
6. The same git repo known to both roots via different remotes/aliases (#645's gap). One project references it — is it "the same repo" everywhere?
7. A convoy launched from kiwi with vessels placed on feta; kiwi lid-closes mid-run (the T3 scenario). Where does the convoy's record live, who can act on it, what does feta show?
8. A root dies permanently (disk loss). Which records should have survived elsewhere, and in what form — readable memory, or recoverable authority?
[...]

PR #741, opened 2026-07-14
docs: ADR 0016 — overlay replication for cross-root state

rjwittams:
The mechanism half of the #688 grill (the semantics half is recorded on the issue). Decides, against the fixed expectations:

- **Overlay model**: per-root authored logs, no root ever writes another's log (ADR 0002's invariant held by construction); **durable** materialized replicas (never caches — ephemeral replicas would regress the merged view when an origin sleeps); relay of origin-immutable log segments; deterministic merge view — agreement by confluence, not consensus.
- **Field merge**: per-field dots + per-record seen-vector, MV-register semantics. Causal supersession applies silently; true concurrency surfaces as conflicts; the clock never decides; resolution is an ordinary superseding write. Explicitly not last-writer-wins.
- **One channel**: the store's existing k8s-style HTTP watch API over a forwarded UDS; a replicator per peer; live tail (no cadence knob); per-kind opt-in by class; `FleetReplicaSnapshot` + Plane-A peer transport become deletable.
- **Deletion**: authored tombstones, delete-vs-edit surfaces as a conflict, **permanent retention** (no GC — every horizon reopens resurrection with weeks-sleeping laptops in the fleet).
- **Host death**: `flotilla host <name> declare-lost` / `found` (human decree — no timeout can tell dead from asleep); Host spec is definitions-class, status stays convergent facts; definitions need **no adoption** (replicas are permanent and survivor edits supersede — authorship is not ownership).
- **Homing granularity**: vessels home at their placement host; the convoy record is coordination only and spans logs; per-vessel re-provision on vessel-host death; **succession** (not re-provision) when the convoy's home dies or hands off — a hundred-vessel convoy is never torn down because its record-keeper died.
- **Conflict UX**: detection is a read-path property; `<kind> <name> conflicts` / `resolve` CLI verbs; table row badges per ADR 0012; no new query family.

Also adds glossary entries: Convergent Facts, Definition (resource class), Home-bound Runtime, Succession.

Refs #688.

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -232,6 +232,41 @@ domain semantics; the whole product family uses it. Named for a tender boat —
 it ferries and services connections between ship and shore.
 _Avoid_: Bridge, peer, gateway, proxy, router, tug.
 
+**Convergent Facts**:
+The cross-root resource class whose records are natural-keyed and true
+everywhere at once — **Repository**, Host status, observed **Checkouts**.
+Duplication across roots is harmless and union is the entire merge; no
+conflicts are possible. (ADR 0016.)
+_Avoid_: Reference data (that is external, per ADR 0006), observed data
+(broader).
+
+**Definition (resource class)**:
+The cross-root resource class of human-named, human-authored records —
+**Project**, **WorkflowTemplate**, Host spec. Replicated everywhere,
+editable from any root including offline; merged field-by-field: a write
+made having seen the current value supersedes silently, concurrent
+same-field writes surface as conflicts and are never auto-resolved; deletes
+are permanent tombstones. (ADR 0016.)
+_Avoid_: Config, global resource, CRDT-as-a-noun.
+
+**Home-bound Runtime**:
+The cross-root resource class reconciled only at its home, which is fixed at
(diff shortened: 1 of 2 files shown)


rjwittams:
All four addressed: softened the ADR 0012 badge citations to say the badge-with-actions extension is not yet specified there; heading/body now say "running work", not "atoms"; renamed to "no takeover step" with an explicit disambiguation from ADR 0003's adoption axis; and the false-positive race is now ruled rather than open — **a home that learns of a successor to its own record stands down** (causal record decides; partition-era writes surface in the loud return event; concurrent successors are detectable via duplicate `succeeded_from` and surface as an ordinary conflict). Never-fork holds one level up.

rjwittams:
Good catch — added a bullet to the channel section: the cross-root classes are *merge* classes, orthogonal to storage class. Durable-log kinds (definitions, Repository, home-bound runtime) replicate resumably across restarts on both ends; generational observed kinds (Host status, observed Checkouts) follow ADR 0004 — replica survives the holder's restarts, origin generation change forces discard + re-list, and origin *absence* is not a generation change (last generation retained as last-known, staleness-stamped).

Issue #732, opened 2026-07-14
Start convoy from issue: intent verb + admission completion + snapshot Brief


rjwittams:
## Parent
#629 · Story D of the M1 user-story grill (2026-07-14) → ADR 0015 (PR #727), amended by the #728 design grill and ADR 0014's source-qualified issue-reference ruling.

## What to build
The end-to-end path from an issue row to a working crew, per ADR 0015:

- **Intent verb**: one flag-complete, non-interactively drivable `flotilla convoy start` invocation. The caller supplies a Project plus an issue ref/ID (provider-native shorthand is fine); the TUI issue-row action submits the same structured intent. The future governor is just another caller.
- **Issue Source resolution**: resolve through the Project used to start work—its `issue_source` override when present, otherwise the relevant Repository Forge. Issue IDs are opaque strings, not numbers.
- **Admission completion** on the convoy's home daemon, before the spec persists: branch + convoy name generated together (reuse the Plane-A AiUtility call, new call site) with a deterministic fallback based on slugified issue title + opaque external ID when keyless/offline. Checkout path derives from branch. Store never holds a half-spec.
- **Workflow default**: unspecified workflow resolves to the Project's concrete `default_workflow_ref`; admission fails visibly while the referenced WorkflowTemplate or Repository is locally unresolved.
- **Issue reference on work**: ConvoySpec carries the relevant RepositoryKey, canonical Issue Source identity, opaque external issue ID, and display snapshot (title, state, `as_of`). The source identity is captured with the work so a later Project issue-source change does not rewrite history.
- **Snapshot Brief**: generated preamble (convoy, branch, repo, role, completion verbs) + issue title/body/labels stamped `as_of`, with the full source-qualified ref alongside + the human free-text slot.
- Auto-attach on create per the recorded default (interactive starts land in the crew session).

Explicitly out: governor/bosun (upstream completers, later), human-approval gate (policy knob on this pipeline, later), multiple simultaneous Project issue sources, and hull christening (record the field direction only if the allocation path is touched).

## Acceptance criteria
- [ ] From a Project issue row, one action produces a convoy with coherent generated branch/convoy names and lands attached in a genuinely contained crew session
- [ ] The same result is reachable from one non-interactive CLI invocation
- [ ] GitHub-style numeric and Linear-style opaque issue IDs both round-trip
- [ ] The persisted work reference remains resolvable/understandable after the Project changes its Issue Source
- [ ] Fully specified intent receives no completion (governor-path property)
[...]

Issue #729, opened 2026-07-14
Demand-backed query materialization + issues{scope} query family


rjwittams:
## Parent
#629 · Story B of the M1 user-story grill (2026-07-14) → ADR 0014 (PR #727), demand-backed + issues sections, as amended by the #728 design grill.

## What to build
The Aggregator's second materialization class and its first family, `issues{scope}`:

- **Parameterized QueryId**: family + owned scope params (Project ref / typed RepositoryKey). Scope rides the View address per ADR 0013's `?key=value` channel.
- **Issue Source resolution**:
  - Repository scope uses that Repository's Forge-backed issue source.
  - Project scope uses the Project's single `issue_source` override when present.
  - Without an override, Project scope expands to the constituent Repositories' Forge sources and unions their windows inside the Aggregator.
  - No resolvable source returns an explicit `IssueSourceUnavailable` condition, never an ordinary empty result set.
- **Portable binding, local provider**: Forge/Issue Source identifies the external service + scope only. Each host independently selects an installed provider and credentials capable of serving it; provider implementation names and credentials never enter Project/Repository definitions.
- **Opaque IDs**: issue IDs are strings. GitHub `728` and Linear `WIDGET-123` are both valid; nothing assumes numeric issue numbers.
- **Demand-backed materializer**: exists only while ≥1 subscription; window = first page of open issues, updated-desc; kept fresh through the selected IssueProvider's `list_issues_changed_since`; discarded on unsubscribe. Nothing persists.
- **Fetch-more as an intent on the result set**: appends the next page as ordinary deltas; paging state server-side; consumers never stitch pages. Search beyond the window = separate ephemeral scoped query over `search_issues`.
- **Set metadata**: demand-backed ResultSets carry `as_of` and `has_more`; store-backed sets omit them. Seq/gap contract unchanged.
- **Reuse**: existing IssueProvider adapters are the fetch layer (paged list, search, fetch-by-id, changed-since + ETag). Replay fixtures carry over.

Explicitly out: storing issues (the Forge/Issue Source is the system of record — ADR 0014 forbids replication), issue mutation features, multiple simultaneous Project issue-source overrides, and `change_requests{scope}` (same shape, later).

## Acceptance criteria
- [ ] Subscribing `issues{repo}` yields a windowed ResultSet with `as_of`/`has_more`; unsubscribing tears the materialization down
- [ ] Project scope with no override returns the union of constituent Repository Forge windows
[...]

PR #751, opened 2026-07-15
Source-addressed issue fetch seam

### LINKED ISSUE #747: Source-addressed issue fetch seam: IssueRef, grown Issue shape, fetch-by-id
## Parent
Extracted from #729 (its stage 5) to break the #732 → #729 serialization. Both consume this; neither needs the other.

## What to build

The shared issue-data substrate both #729 and #732 sit on, per the rulings already recorded on those issues:

- **`IssueRef { source, id }`** in protocol — source-qualified, opaque id (ADR 0014 as amended: GitHub `728` and Linear `WIDGET-123` are both opaque; two repos' issues must never collide in a Project union).
- **Grown normalized `Issue` shape**: body, state, `as_of` in addition to today's title/labels (today's shape has none of these — verified). #732's Brief snapshot and #729's window rows both need it.
- **Source-addressed fetch seam**: provider calls take an external source identity (Forge / Project Issue Source per ADR 0014 glossary), with host-local provider/credential selection behind the seam — never a checkout path (cloud-only Projects have no checkout). Converge the two existing issue interfaces (`IssueProvider` / `IssueQueryService`) onto it rather than adding a third; Plane-A implementations remain the fetch layer, rebound.
- **Verbs**: fetch-by-id first (unblocks #732's admission snapshot); paged list / search / changed-since rebound to the seam behind it (consumed by #729's materializer).
- Project → Issue Source resolution (single override, else constituent Repositories' Forges; unresolvable = unavailable, never empty — ADR 0014 as amended).

## Acceptance
- [ ] Fetch an issue by opaque id from a source identity with no checkout on the host
- [ ] `IssueRef` + grown `Issue` defined once in protocol; no second definition in #729/#732
- [ ] Existing issue interfaces converged; GitHub adapter passes existing behavior through the new seam (replay fixtures re-recorded, not hand-edited)
- [ ] Project issue-source resolution surfaces "unavailable" as a typed outcome

## Blocked by
None — can start immediately. Unblocks #732 (admission snapshot + Brief) and #729's materializer stage; #729's delta/registry stages (1–4) proceed in parallel regardless.

rjwittams:
Closes #747

## What changed

- Added canonical protocol-level `IssueSource`, `IssueRef`, `IssueState`, and the normalized issue snapshot fields `body`, `state`, and `as_of`.
- Converged `IssueProvider` and `IssueQueryService` into one source-addressed provider interface covering fetch-by-id, paged list/search, and changed-since.
- Added typed Project issue-source resolution: an explicit override wins, otherwise repository Forge identities form a deterministic union, with typed unavailable outcomes.
- Rebound the GitHub adapter to the new seam, made it a host-scoped capability independent of checkout paths, and re-recorded its replay fixture.
- Added a public no-checkout fetch path and migrated existing daemon, executor, and TUI consumers.

## Why

Issue identities were source-local strings and provider calls were coupled to checkout paths. That cannot represent cloud-only Projects safely and allows same-number issues from different repositories to collide in Project-level collections. The new seam makes the external source identity explicit while keeping credentials and provider selection local to the host.

## Impact

This unblocks the issue snapshot work in #732 and the materializer work in #729. Existing repository issue browsing and branch-name flows continue through the unified provider interface.

## Validation

- GitHub record/replay query plus fetch-by-id coverage
- No-checkout source-addressed integration coverage


rjwittams:
Fixed the batch-fetch regression in 9429d10f: the shared IssueProvider default now runs up to 10 fetches concurrently (matching the old GitHub adapter), preserves input order, and has a regression test that asserts the concurrency bound. I also documented why bare GitHub service names are accepted. The fail-closed Project resolution is intentional: issue #747 specifies that an unresolvable constituent yields typed unavailable rather than a partial/empty source result.

PR #760, opened 2026-07-18
Harden recovery across attachables, workspaces, and daemon clients

### LINKED ISSUE #752: Discovered attachable sets never correlate to their worktree (checkout only set on the executor path)
Dogfood finding (daily driver, 2026-07-15, Robert): attachable sets in the flotilla repo don't correlate to their worktree rows on the Plane-A repo page.

## Mechanism

- Correlation emits `CorrelationKey::CheckoutPath` for a set **only when `set.checkout` is `Some`** (`data.rs` `correlate()`, attachable-set arm).
- `set.checkout` is only populated by `ensure_terminal_set_with_change(..., Some(checkout), ...)` — and the only callers are in `executor/workspace.rs` (the workspace open/create path).
- Sets that come back via **discovery** (e.g. shpool sessions rediscovered after a daemon restart or a store reset — which just happened fleet-wide with the pre-744 archive) get `checkout: None`, so they float uncorrelated forever.

The irony: discovery *has* the data — shpool sessions carry their working dir (the daemon's own workspace-prepare log resolves `shpool ... attach --force --dir '/Users/robert/dev/flotilla.materialization-query-scope' <id>` per member). The set's checkout can be derived from member sessions' dir facts.

## Fix direction

Populate `set.checkout` from member terminal dir facts at refresh/assembly time when absent (convergent observed fact — the dir all members share), rather than only at executor creation. Keep the executor path as-is.

## Acceptance
- [ ] A shpool set whose sessions run in a worktree correlates to that worktree row after a fresh daemon start with an empty attachable registry
- [ ] Plane-A freeze respected: bugfix in the existing pipeline, no new capability

Plane A (bugfix-only) — this is a correctness fix in existing correlation.

### LINKED ISSUE #753: Workspace open reports Succeeded and does nothing when no WorkspaceManager is active
Dogfood finding (daily driver, 2026-07-15, Robert): double-clicking a worktree row in the TUI (worktree with a live shpool attachable set) showed the workspace-prepare progress, then nothing happened.

## What actually happened (daemon log, 13:09:06–08)

- The zellij workspace-manager provider was **never activated**: `MissingBinary("zellij >= 0.44.1 required, found 0.44.0")` — the daemon's PATH found a stale `~/.cargo/bin/zellij` 0.44.0, while the daily driver's native-plugins zellij is 0.45.0 and would pass the gate.
- The open-workspace flow nevertheless ran: step 0 "Prepare workspace" resolved all three member shpool attach commands and reported **Succeeded**; step 1 "Attach workspace" reported **Succeeded in ~50µs** — with no workspace manager to attach anything to. A silent no-op reported as success.

## Two defects

1. **Honesty**: with no active `WorkspaceManager`, the open-workspace command must fail loudly — ideally carrying the discovery reason ("zellij >= 0.44.1 required, found 0.44.0"), which the daemon already knows and logs at DEBUG. A step must not report `Succeeded` when it performed no observable action.
2. **Environment/UX**: the version-gate failure was invisible — `flotilla status` health lists active providers but nothing surfaces *why* an expected provider is missing. The unmet-requirements list exists (`providers not activated: missing requirements`); it should be reachable from status/TUI, not only DEBUG logs.

Plane A (bugfix-only): defect 1 is a correctness fix on the existing action path. Defect 2's surfacing may be minimal (status line) to respect the freeze.

## Environment note (not part of the fix)
The immediate daily-driver unblock is aligning the daemon's zellij with the running spike build (0.45.0) — separate env action, tracked in the acceptance-run notes.

### LINKED ISSUE #754: pm connect: 'received delta for unknown repo, requesting replay' warn at startup
Dogfood finding (daily driver, 2026-07-15, Robert), in the `flotilla pm connect` pane:

```
WARN flotilla_client: received delta for unknown repo, requesting replay repo_identity=github.com:flotilla-org/cleat repo=Some("/Users/robert/dev/cleat")
```

Gap recovery did its job (replay requested, connector recovered) — but a WARN on an apparently ordinary startup ordering (delta arriving before the client learned the repo) is noise that trains people to ignore warnings. Observed once, shortly after fleet restart on fresh stores.

## Work
- Determine whether this is the expected subscribe-before-repo-list race (client subscribes, daemon registers repos asynchronously, first delta for a late repo precedes its announcement). If so, downgrade to `debug!` or close the ordering hole (announce repo set before deltas).
- If it recurs in steady state (not just startup), that's a real identity-mapping bug between delta `repo_identity` and the client's repo table — diagnose then.

## Acceptance
- [ ] Clean startup of `pm connect` against a warm daemon produces no WARN-level log lines
- [ ] Genuine mid-session unknown-repo deltas (if possible) still warn

### LINKED ISSUE #755: Daemon EOF strands event subscribers; long-lived clients must reconnect with exponential backoff
Dogfood finding (daily driver, 2026-07-15, Robert): across a deliberate daemon restart, the `flotilla pm connect` pane logged `ERROR flotilla_client: daemon connection closed (EOF)` and became inoperative — the andamento catalog silently froze until manual restart.

Ruling: **long-lived clients reconnect with exponential backoff rather than becoming inoperative.**

## Mechanism (verified)

Three stacked defects:

1. **The broadcast never closes.** `SocketDaemon`'s reader task breaks on EOF and fails pending *requests*, but `event_tx` is owned by the `SocketDaemon` struct itself, so subscribers' `events.recv()` never yields `RecvError::Closed`. The connector's `Closed` arm (`pm_connect.rs` `run_connector`) is unreachable; its select loop keeps firing the TTL reassert tick, republishing the stale catalog forever. Inoperative-but-alive.
2. **The reconnect loop can kill the process.** `pm_connect::run` loops with a fixed 2s sleep, but `connect_or_spawn(...).await?` propagates — one failed reconnect attempt while the daemon is down exits the process.
3. **No backoff.** Fixed 2s retry, no jitter, no cap.

## Contract

- **Client layer**: on transport EOF/read error, surface termination to broadcast subscribers deterministically (terminal connection-lost event, or drop the sender). No consumer should be able to await a closed connection forever.
- **pm connect**: reconnect with exponential backoff + jitter + cap (e.g. 500ms → 30s); a failed attempt never exits the process; on reconnect, resubscribe — stale cursors already yield full `ResultSet`s via the gap machinery, and the TTL fade + re-assert design (already documented in the module) covers the away period.
- **Incompatible daemon** (a no-compat upgrade landed while the client was away): the typed handshake failure (#723) must terminate the retry loop with a clear version message — daemon-away retries forever, daemon-incompatible exits loudly. The two must be distinguishable.
- **Audit the other long-lived consumers**: `flotilla watch`, the TUI event loop — same stranding applies to anything holding a `subscribe()` receiver.

## Acceptance
- [ ] Kill + restart the daemon under a running `pm connect`: connector reconnects within backoff, catalog re-published, no manual intervention
- [ ] Stop the daemon for longer than several backoff steps: connector keeps retrying at the cap, recovers when the daemon returns
- [ ] Reconnect to a handshake-incompatible daemon: clean exit with the version message
- [ ] A clean reconnect logs at INFO, not ERROR

rjwittams:
## What changed

- recover discovered attachable sets from provider-safe terminal session metadata and correlate them to their checkout after store reset
- fail workspace-open commands loudly when no active workspace manager can satisfy the request, and expose unmet requirements in status output
- suppress the expected unknown-repository replay warning during initial synchronization while retaining warnings for genuine mid-session gaps
- close event subscriptions deterministically on daemon EOF and reconnect `pm connect`, `flotilla watch`, and the TUI with capped exponential backoff and jitter
- stop retrying on handshake-incompatible daemons and preserve the clear version error
- add focused regression and acceptance coverage for each path

## Why

These four dogfood findings shared a recovery theme: discovered state could not be reconstructed after a reset, missing workspace capabilities were reported as success, normal startup recovery was noisy, and daemon restarts stranded long-lived clients. Together they made recovery unreliable and, in some cases, silently stale.

## User impact

Attachable sessions recover their checkout association after registry loss, workspace opens report unavailable providers instead of succeeding without effect, startup logs are quieter, and long-lived clients recover automatically after daemon restarts.

## Validation

- `cargo test --workspace --locked --features flotilla-daemon/skip-no-sandbox-tests`
- `cargo test -p flotilla-tui --lib cli::tests --locked`

Closes #752
Closes #753
Closes #754
Closes #755


rjwittams:
Fixed in ab66f389. Checkout inference now requires both the member working directory and the attachable set namespace to match: stable local host IDs are checked against the refresh host identity, legacy host names against host affinity, and environment paths against environment_id. Added a regression that rejects a sole same-path remote checkout and then selects the matching local checkout when both exist. Full fmt, clippy, and workspace tests pass locally.

Issue #763, opened 2026-07-18
Reconcile adopted checkout persistence across durable and observed stores


rjwittams:
From PR #762 review by @claude: https://github.com/flotilla-org/flotilla/pull/762#issuecomment-5011125273

Adopted checkout creation currently persists the controller-facing durable resource and the query-facing observed fact sequentially. A failure between those writes can temporarily leave only one representation. Retrying the same adoption self-heals because matching conflicts are idempotent, but there is no periodic reconciliation for adopted authority resources.

Design and implement an explicit consistency mechanism across these stores. This should define rollback or reconciliation semantics for failures during create and status update without deleting a pre-existing adopted checkout owned by an earlier successful operation.

Context: PR #762 intentionally keeps both representations because vessel controllers consume the durable checkout while `checkouts{scope}` projects observed facts only.

rjwittams:
Implemented by #767 (durable-first persistence, restore + periodic reconciliation across restarts, interrupted-status healing — the closes-link was just missing). The deliberately-deferred piece — removing observed projections when durable checkouts are *deleted* — is #768, waiting on an adopted-checkout deletion path existing at all.

PR #767, opened 2026-07-20
fix: reconcile adopted checkout persistence

rjwittams:
## Summary
- persist adopted checkouts durably before projecting them into the observed store
- restore and periodically reconcile adopted checkout projections across daemon restarts
- preserve durable status, heal interrupted status writes, and isolate observed-store collisions

## Validation
- daemon unit tests: 202 passed
- daemon aggregator tests: 6 passed
- core adopted-checkout reconciliation tests passed

### DIFF
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -582,8 +582,14 @@ async fn create_adopted_checkout_resource(
     );
     let status = ResourceCheckoutStatus::builder().phase(ResourceCheckoutPhase::Ready).path(path_str).build();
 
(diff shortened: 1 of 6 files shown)


rjwittams:
Thanks for the review. I filed #768 to track observed-projection cleanup when adopted Checkout deletion is introduced. The joined reconciliation error remains intentionally log-only today; callers do not branch on individual item failures.

rjwittams:
Sanity-checked: retaining a matching durable Failed status is intentional. The durable resource is authoritative, and #763 explicitly requires preserving any existing durable status rather than inferring a transition from a later ConvoyCreate request. A future reconciler-owned Failed-to-Ready transition would need its own explicit state-transition rule.

PR #769, opened 2026-07-20
Stream zellij metadata patches through a persistent pipe

### LINKED ISSUE #757: pm connect: batch metadata patches into one pipe payload per publish
Found while investigating the zellij action-queue wedging (daily driver, 2026-07-15). Not claimed as the wedge's root cause — but it is standing, wrongly-shaped load. (Original framing said "batch into one payload"; corrected by Robert: the whole point of `zellij pipe` is the **long-running streaming mode**.)

## Current shape (wrong)

`ZellijPipeSink::send` spawns a **one-shot `zellij pipe --name … -- <payload>` subprocess per `MetadataPatch`** (`crates/flotilla-manifest/src/sink.rs`), and the connector re-asserts the full catalog every 10s. That is spawn + server connection + handshake per message, O(catalog) every interval, forever.

## The designed mode (verified in zellij `cli_client.rs`)

`zellij pipe` with piped stdin and no argv payload runs a streaming loop: one line from stdin → one message to the server → wait for `UnblockCliPipeInput` (per-message backpressure from the consuming plugin) → next line. One persistent child, plugin-paced.

## Work

- `ZellijPipeSink` holds a **long-running `zellij pipe --name andamento-apply-metadata-patch [--plugin url]` child** and writes one line per patch to its stdin. No per-patch spawning.
- Payloads must be single-line (one stdin line = one message) — assert/escape in `to_pipe_payload`.
- Child lifecycle: respawn on exit (zellij restart); a blocked stdin write is the plugin applying backpressure — pace the connector by it, and surface (log/status) if it stays blocked past a threshold instead of wedging silently. Coordinate with #755's reconnect discipline — same supervision shape, different peer.
- The one-shot `zellij pipe` stamp call at workspace creation (zellij provider) is low-frequency and can stay one-shot.
- Wheelhouse's `UnixSocketSink` is unaffected; the transport choice stays in the sink.

## Acceptance
- [ ] Steady-state daily driver keeps exactly one zellij pipe child, regardless of catalog size or publish rate
- [ ] Plugin backpressure paces the connector (no unbounded queueing on either side); sustained blockage is surfaced, not silent
- [ ] Pipe child death (zellij restart) is recovered automatically; catalog TTL/fade semantics unchanged

rjwittams:
Closes #757.

## What changed

- keep one long-running `zellij pipe` child and send one metadata patch per stdin line
- serialize writes so plugin backpressure paces the connector, with a warning after sustained blockage
- recover from child exit or broken stdin with capped exponential respawn backoff
- enforce the single-line JSON payload invariant
- cover streaming, restart recovery, broken-pipe retry, backpressure, and escaped-newline framing

## Why

`ZellijPipeSink::send` previously launched a one-shot subprocess for every patch. Catalog reassertion therefore paid process startup and Zellij connection/handshake cost once per catalog entry every ten seconds. Zellij's intended pipe mode is a persistent stdin stream with plugin-driven per-message backpressure.

## Impact

Steady-state `pm connect` now maintains one Zellij pipe process regardless of catalog size or publish rate. Wheelhouse's Unix socket transport and catalog TTL/fade behavior are unchanged.

## Validation

### DIFF
--- a/crates/flotilla-manifest/Cargo.toml
+++ b/crates/flotilla-manifest/Cargo.toml
@@ -10,7 +10,7 @@ bon = { workspace = true }
 flotilla-protocol = { path = "../flotilla-protocol" }
 serde = { workspace = true }
 serde_json = { workspace = true }
-tokio = { version = "1", features = ["process", "net", "io-util"] }
+tokio = { version = "1", features = ["process", "net", "io-util", "sync", "time"] }
 tracing = { workspace = true }
 
 [dev-dependencies]
--- a/crates/flotilla-manifest/src/sink.rs
+++ b/crates/flotilla-manifest/src/sink.rs
@@ -3,12 +3,23 @@
 //! Per the manifest architecture, producers swap only their send function —
 //! the same projection drives zellij (CLI pipe) and wheelhouse (unix socket).
 
-use std::path::PathBuf;
+use std::{path::PathBuf, process::Stdio, time::Duration};
 
 use async_trait::async_trait;
-use tokio::{io::AsyncWriteExt, net::UnixStream, process::Command};
+use tokio::{
+    io::AsyncWriteExt,
+    net::UnixStream,
+    process::{Child, ChildStdin, Command},
+    sync::Mutex,
+};
+use tracing::warn;
 
 use crate::{keys::APPLY_METADATA_PATCH_PIPE, wire::MetadataPatch};
+
+const BLOCKED_WRITE_WARNING_AFTER: Duration = Duration::from_secs(5);
+const RESPAWN_INITIAL_DELAY: Duration = Duration::from_millis(500);
+const RESPAWN_MAX_DELAY: Duration = Duration::from_secs(30);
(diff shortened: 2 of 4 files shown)


rjwittams:
Thanks — I checked this path carefully and am keeping the current accounting because these are two distinct failure events, not one event applied twice.

The first `schedule_respawn()` records the dead/broken existing child and its deadline is consumed by `ensure_process()` before it attempts the replacement. If that new `spawn_process()` attempt also fails, the second `schedule_respawn()` records that separate failed reconnect attempt. Without it, `respawn_not_before` has already been taken, so the next catalog patch would attempt another spawn immediately and defeat the reconnect backoff. This matches #755's discipline of advancing backoff for each failed reconnect attempt.

The inherited stderr and single sustained-block warning are deliberate here: zellij's diagnostic remains visible in the connector pane, while the structured warning surfaces the transport state; one warning satisfies #757 without periodic log spam. The single-line assertion is the requested invariant guard around serde's escaping guarantee.

PR #770, opened 2026-07-20
Materialize Projects for every tracked repository

### LINKED ISSUE #766: Materialize degenerate Projects for every tracked repo (project-only view scope)
## Parent
#735 / #764 (2026-07-20 scope-unification ruling: the view layer speaks Project only).

## What to build

Every tracked repo gets its **degenerate whole-repo Project** (per #744's shape) materialized automatically:

- **At repo-track time**: tracking a repo creates its degenerate Project with a deterministic name derived from the repo slug (collision-aware across namespaces). Tracking *is* the human intent — the definitions-class creation is not speculative.
- **Backfill**: currently-tracked repos without a Project get theirs on daemon start (idempotent, create-if-absent).
- `flotilla project add` remains the explicit path (multi-repo, custom config); a degenerate Project later "growing up" (more repos, issue-source override) is an ordinary Project edit, not a migration.

This is what makes project-only scope honest: opening any repo opens its Project's page (#764) with no virtual-project special case, and start-convoy always has a project + default workflow.

## Future notes (recorded for later design, out of scope)

From Robert, 2026-07-20:
- Later, a **'project setup' agentic flow**: a conversation + investigation establishing a project's real scope — natural once 'always checkout every constituent repo' isn't mandatory (per-vessel narrowing from #756 is the mechanism). The Project trends toward a **'manifest of relevant repos'** rather than 'code always modified together'.
- **Issue-tracker selection** needs working through later — especially when a project includes upstream open-source components (whose Forges you observe but don't triage in).
- Project data itself possibly **source-control backed**, holding structured data plus 'prompty' instructions for agents working in the project's context.

## Acceptance
- [ ] Tracking a new repo yields a usable `project/<ns>/<name>` address with no manual step
- [ ] Existing tracked repos backfilled idempotently
- [ ] Degenerate Projects are ordinary Projects (editable, growable) — no special kind

rjwittams:
## Summary

- materialize an ordinary whole-repository Project whenever a repository is tracked
- backfill Projects for repositories that were already tracked when the daemon starts
- allocate deterministic, collision-safe Project names while preserving Projects that users have expanded or customized
- fail tracking or startup when the required Project cannot be materialized
- support inspecting brand-new Git repositories whose `HEAD` is still unborn

## Why

The view layer now scopes work through Projects. Repository tracking is the human intent that establishes the default Project, so every tracked repository must have a usable Project without a separate setup step or a virtual-Project special case.

Closes #766.

## User impact

Tracking a repository immediately yields a usable Project backed by the `single-agent-contained` workflow. Existing tracked repositories receive the same Project idempotently on restart, and those Projects remain ordinary editable resources that can grow to include more repositories or an issue-source override.

## Validation

- `RUSTC_WRAPPER= cargo clippy --workspace --all-targets --locked -- -D warnings`
- `TMPDIR="$PWD/.codex-tmp" RUSTC_WRAPPER= cargo test --workspace --locked --features flotilla-daemon/skip-no-sandbox-tests`

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -22,7 +22,10 @@ _Avoid_: using "Island" as a code/identifier name — it is the metaphor, not th
 **Project**:
 The definition resource that structurally groups work (convoys) belonging to
 one codebase, which may span one or more **Repositories** or repository slices.
-The concrete form of the **Island** metaphor.
+Tracking a Repository materialises an ordinary whole-repository Project as the
+human-intent-backed default; that Project remains user-editable and may grow to
+cover more Repositories or select a different Issue Source. The concrete form
+of the **Island** metaphor.
 _Avoid_: Repo (a Project may span more than the bare git remote), Workspace.
 
 **Repository**:
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -2310,15 +2310,15 @@ async fn ensure_single_agent_contained_workflow(backend: &ResourceBackend, names
     }
 }
 
-fn normalize_project_name(name: &str) -> Result<String, String> {
-    let normalized = normalized_project_name(name)?;
+fn validate_project_name(name: &str) -> Result<(), String> {
+    let normalized = normalize_project_name(name)?;
     if normalized != name {
         return Err(format!("project name `{name}` is invalid; use `{normalized}`"));
     }
-    Ok(normalized)
+    Ok(())
 }
 
-fn normalized_project_name(name: &str) -> Result<String, String> {
+fn normalize_project_name(name: &str) -> Result<String, String> {
     let normalized = name
         .trim()
         .chars()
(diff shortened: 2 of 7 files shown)


rjwittams:
Addressed the startup concern in `b4c9fbaf`. Backfill now warns and skips a tracked path when repository inspection fails, so a stale path or unavailable mount cannot prevent the daemon from starting. Failures after successful inspection (for example, a durable resource backend error while creating the Project) still fail startup rather than being silently swallowed. The regression test now asserts successful startup with an uninspectable tracked repository.

I also changed the add-repository error to the user-facing `cannot track repository …` form.

For the unborn-HEAD note, the real integration test `startup_registration_skips_repos_without_origin_and_gates_docker_policy` initializes a repository with `git init` and no commit; that test exposed the failure before the fallback and passes with it. I added a short code comment explaining why `symbolic-ref` is used, while retaining the focused mock test for exact command behavior.

rjwittams:
Conscious call on deterministic-name exhaustion: manual intervention is the intended recovery when all three candidates are occupied by unrelated Projects.

The sequence deliberately widens from the human-friendly leaf slug, to the catalog slug, to the catalog slug plus repository-key prefix. An unrelated Project occupying that final identity-derived address is a namespace ownership conflict, not a normal hash collision. Inventing a numeric fourth candidate would make the selected address depend on creation order and hide that conflict. Returning an explicit error also avoids overwriting or silently adopting a user-authored Project.

For an ordinary `repo add`, the failure happens before the repository is tracked. The startup case requires an already-tracked repository to lack its Project while all three addresses were subsequently claimed, which indicates state needing operator repair. I’m therefore retaining the explicit failure for that post-inspection materialization error.

Issue #773, opened 2026-07-20
Suppress project-table rows represented by higher resources


rjwittams:
Parent: #764.

A Project page should present the highest-level live representation of work exactly once. When a resource already represents another item in a higher section, omit the lower item. For example, after starting a Convoy from a scoped Issue, the issue disappears from Issues and is represented by the Convoy in Convoys; likewise, a Checkout owned or used by a Convoy is not duplicated in Checkouts.

This is a Plane-B query/result-set concern, not a TUI-only hide rule. ConvoySpec.issue already stores a durable, source-qualified IssueRef plus snapshot, but ConvoyRow does not project it. Convoys also own/adopt checkout references, while scoped checkout rows have no convoy owner/ref.

Define a typed higher-resource claim/representation relationship; project enough identity into the relevant query inputs or result sets to suppress by identity rather than title, branch, path, or provider-local-ID heuristics. Specify terminal/deletion semantics, preserve demand metadata/pagination, and cover live transitions: starting moves an Issue upward; relinquishing its claim restores the lower row.

Acceptance: one representation of a started Issue (Convoys); no duplicate claimed Checkout; source-qualified deterministic identity across hosts; live delta and terminal-state tests.

rjwittams:
Concrete instance from the 2026-07-21 dogfood round (#782-#790 batch): the four failed convoys each provisioned a worktree under `~/dev/flotilla-repos/<convoy>/…`, and those checkouts appear in the project page's Checkouts panel alongside the convoys that own them. `rebuild_checkout_rows` (`crates/flotilla-daemon/src/aggregator.rs:731`) projects every `CheckoutSpec::Observed` in scope with no claimed-by-convoy exclusion, which is the gap this issue describes. Note the claim must also survive the convoy reaching `Failed` — these convoys are dead but still own their checkouts (cleanup story is #785), so terminal-state semantics matter for the suppression rule too.

PR #775, opened 2026-07-20
feat: unify find interaction

### LINKED ISSUE #736: Tables v1 slice 4: finding and narrowing table work
## Parent
#666 (umbrella) · ADR 0012 §Filtering, ADR 0014 (scope params).

## What to build
Establish one global interaction model for finding and narrowing work across Views. Keys, command palette entries, contextual menus, and future surface affordances are presentations of the same available Intents, resolved from the active View address and optional selected row.

- `:` opens the global command palette. It remains one palette, with availability and presentation informed by the active View and selection; it is not a separate palette implementation per View.
- `/` invokes **Find** consistently across Views, seeded with the active View scope. Find is a user intent, not a choice between a local cache filter and a remote provider search. Its implementation may use loaded rows or source-backed queries, but that distinction must not leak into the interaction.
- Model view refinement separately from Find. Refinement is a typed, addressable narrowing of the View and may grow beyond the initial host/project/phase-class examples; do not freeze a prematurely narrow scope type or binding.
- The reusable table and Plane-A repo page consume the shared interaction vocabulary rather than assigning conflicting meanings to the same key.

## Out of scope
- Search ranking, provider/source fanout, and persisted saved searches.
- #773's result-representation behavior: when an associated Issue matches, Find should ultimately surface its higher Convoy representation plus match evidence, but the claim/projection/data-plane work is owned by #773.

## Acceptance criteria
- [ ] `:` always opens the global command palette; its available actions are resolved from the active View and optional selection
- [ ] `/` has one Find meaning across table and repo Views; it does not mean a cache-specific filter or a separate source-search mode
- [ ] The interaction resolver is the shared seam for palette/action availability and key eligibility
- [ ] View refinement remains distinct from Find and can be encoded in a future View address without a forced migration

## Blocked by
None — #735 is complete.

rjwittams:
## Summary

Unifies the TUI's global interaction model for finding work.

- `:` opens the global command palette; `/` opens Find.
- Key handling, palette entries, and status-bar hints use the same active-view resolver.
- The resolver retains typed selected-row context for future contextual interactions and associated-resource results.
- Removes the visible local/remote search split without committing the result model needed by #773.

## Validation

Closes #736.


rjwittams:
Resolved in 56343562 after merging current `main` into this branch. `ProjectPageWidget` now handles `Action::OpenFind` by opening `TableSearchWidget::find` for its focused panel, and the obsolete filter/source-search action references are removed. The merged branch passes formatter, strict Clippy, and the sandbox-safe workspace test suite locally.

rjwittams:
Fixed in 56343562. After merging current main, Project panels route `Action::OpenFind` to `TableSearchWidget::find` and no longer refer to removed actions or constructors. Formatter, strict Clippy, and the workspace test suite pass locally on the merged branch.

PR #778, opened 2026-07-21
Add project list command

### LINKED ISSUE #776: flotilla project list: enumerate projects (names, repos, addresses)
## What to build

`flotilla project` has `add` and `apply` but no `list` — with #770 materializing a degenerate Project for every tracked repo, there's now no way to discover project names (needed for `project/<namespace>/<name>` addresses, `:open` in the palette, `flotilla view` deep links) except knowing the slug convention.

Add `flotilla project list` (subject-verb convention allows the bare-noun list form, matching `flotilla host list`):

- One row per Project: namespace/name, constituent repository count (or slugs when few), issue-source override if set, default workflow ref.
- Include the view address (`project/<ns>/<name>`) — the thing you paste into `:open` / `flotilla view`.
- `--json` for scripting, matching the global flag.

## Acceptance
- [ ] Lists all Projects including materialized degenerate ones
- [ ] Address column usable verbatim with `flotilla view` and the palette `open` command
- [ ] `--json` output stable and typed

rjwittams:
## Summary

- add `flotilla project list` with a concise human-readable table
- expose stable typed JSON output through the global `--json` flag
- include repository summaries, issue-source overrides, default workflows, and pasteable project addresses
- enumerate both explicit and materialized degenerate projects in deterministic order

## Why

Operators need a discoverable inventory of known projects and canonical addresses they can paste into other Flotilla commands. The protocol also needs a stable machine-readable shape for scripting.

## Impact

Users can now list every known project from the CLI and consume the same data programmatically as typed JSON. Existing commands and project materialization behavior are unchanged.

## Validation

- `cargo dylint --all -- --all-targets`

Closes #776


rjwittams:
Addressed the issue-source presentation nit in `88c22adc`. Project-list output now uses the repository's existing `service / scope` convention, so an arbitrary scope is not rendered as though it were a URL path. The formatter regression test was updated, and the pinned formatting check, workspace Clippy with warnings denied, and the sandbox-safe workspace test suite all pass.

The pagination observation needs no follow-up in this PR: project listing intentionally matches the existing unbounded host-list query shape, and the review identified no current scale or correctness problem.

rjwittams:
Addressed the redundant-count concern in `70da54bd`. `ProjectListEntry` now exposes only the authoritative `repositories` collection; human output derives its few-versus-many threshold and count from `repositories.len()`, and the daemon test preserves the repository-slice dedup assertion through the collection length. This removes the possibility of two stable JSON fields drifting out of agreement.

Focused protocol/core/TUI tests and the exact pinned formatting, workspace Clippy, and sandbox-safe workspace test gates all pass.

PR #794, opened 2026-07-21
fix: keep convoy starts responsive

### LINKED ISSUE #782: fix: ConvoyStart executes inline in the daemon and freezes the TUI for 30-60s
## What happens

Starting a convoy from an issue row (project page → action menu → start convoy) freezes the whole TUI for 30–60s. Keypresses buffer during the freeze and re-fire the action when the loop unblocks, so one user attempt became four convoys at exactly ~61s intervals. Found during #666 dogfood (2026-07-21).

## Diagnosis

`ConvoyStart` is the one command that violates the daemon's fast-ack contract. Every other command `tokio::spawn`s its work and returns the command id immediately (`crates/flotilla-core/src/in_process.rs:4811`); the `ConvoyStart` branch (`in_process.rs:4328-4365`) runs **inline** before returning:

1. `admit_convoy_start`, which includes a network call to the Anthropic API for convoy/branch name generation (`in_process.rs:2575`) through `ReqwestHttpClient` — which is built with `reqwest::Client::new()` and has **no timeout at all** (`providers/mod.rs:451`).
2. With `auto_attach` (hardcoded `true` in the TUI intent, `crates/flotilla-tui/src/app/key_handlers.rs:405`), a 60s poll loop waiting for an attachable crew session (`in_process.rs:4341-4350`) that **never checks the convoy phase** — the dogfood convoys went `Failed` at +7s and the handler spun the remaining ~53s anyway.

Meanwhile:
- The socket client request timeout is 30s (`crates/flotilla-client/src/lib.rs:512`), so the eventual response is dropped as `received response for unknown request id` and the user never sees the error.
- The TUI event loop drains the command queue with a synchronous `dispatch(cmd).await` (`crates/flotilla-tui/src/run.rs:140-142`), so the UI (including Escape and the painted action menu) is frozen for the full client timeout.
- The daemon serializes requests per client, so each buffered retry dispatches exactly when the previous 61s handler returns — the retry storm.

## Fix shape

- Spawn the `ConvoyStart` work like every other command and return the command id immediately; the TUI already handles `ConvoyStarted { attach_command }` arriving via `CommandFinished` (`crates/flotilla-tui/src/app/executor.rs:169-175`).
- Make the auto-attach wait watch the convoy and bail out as soon as the phase becomes `Failed`, carrying the per-work failure message (not the generic convoy-level message) into the command error.
- Give `ReqwestHttpClient` a sane timeout so name generation can never hang admission indefinitely.

## Acceptance criteria

- [ ] Starting a convoy returns control to the TUI immediately (spinner/pending row, no event-loop freeze)
- [ ] A convoy that fails during bootstrap reports the real failure to the initiating client within a few seconds, not after a 60s timeout
- [ ] Repeated Enter presses during an in-flight start do not create additional convoys

rjwittams:
## Summary

- run convoy admission and auto-attach asynchronously so the TUI remains responsive
- reject duplicate in-flight convoy starts and surface context-free progress in project view
- stop auto-attach promptly on failed work, preserving the detailed failure message
- bound Anthropic and general Reqwest request durations

Closes #782

## Testing

- `cargo check --workspace --all-targets --locked`
- focused convoy-start, Anthropic-timeout, and TUI status-bar regression tests

The full local workspace suite reaches one unrelated macOS path-alias failure in `dispatch_add_list_remove_repo_round_trip`: the actual path begins with `/private/var` while the expected path begins with `/var`.

### DIFF
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -8,7 +8,7 @@ use std::{
     path::{Path, PathBuf},
     sync::{
         atomic::{AtomicU64, Ordering},
-        Arc,
+        Arc, Weak,
     },
     time::Duration,
 };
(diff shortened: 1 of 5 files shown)


rjwittams:
Agreed this is a valid defensive hardening opportunity rather than a current blocker: the existing fallible paths return `Result`, but an unexpected future panic could strand the admission key. Filed #796 to add panic-safe cleanup and regression coverage: https://github.com/flotilla-org/flotilla/issues/796

PR #795, opened 2026-07-21
fix: align tab close hotspot with rendered affordance

### LINKED ISSUE #787: fix: tab close-button hotspot misaligned under Chevron bar style — visible × dead, seam click closes active tab
## What happens

Two dogfood reports with one root cause:

- Clicking the visible `×` on the active tab does nothing.
- A click just right of the `×` closes the **active** tab even when aiming at the neighbouring tab — which is how the project-page tab silently disappeared.

## Diagnosis

The close hotspot is hard-coded to the last 3 cells of the active tab segment (`crates/flotilla-tui/src/widgets/tabs.rs:212-214`). That math only holds for the Pipe bar style. The default theme (`catppuccin-mocha`, `src/main.rs:405`) uses `BarKind::Chevron` (`theme.rs:167`), where `ThemedRibbonStyle::render_item` wraps every segment as `CHEVRON + " {label} " + CHEVRON` (`crates/flotilla-tui/src/widgets/segment_bar.rs:156-170`). With the label built as `"<text> × "` (`tabs.rs:176`), the `×` lands at `width-4` — one cell left of the hotspot. So:

- The visible `×` resolves to `SwitchTo(active)` — a no-op (`tabs.rs:231-239`, `open_views.rs:370-378`).
- The pad/chevron seam resolves to `Close(active)` — closing whatever is active. And since a closed view is persisted out of `open-views.toml` immediately, the loss survives restart.

## Fix shape

Compute the close hotspot from the rendered segment geometry (or have the bar style report the `×` cell) instead of assuming Pipe layout; cover both bar styles with a hit-test unit test. Consider requiring the click to land exactly on the `×` cell so near-misses select rather than close.

## Acceptance criteria

- [ ] Clicking the `×` closes the active tab under both Pipe and Chevron bar styles
- [ ] Clicking anywhere else on a tab (including the seam) never closes; it selects
- [ ] Hit-test unit tests cover both bar styles

rjwittams:
## Summary

- derive the active tab close hotspot from the rendered segment geometry
- restrict closing to the exact trailing `×` cell for Pipe and Chevron bar styles
- cover every rendered tab cell, Chevron seams, and labels containing their own `×`

## Root cause

The close hotspot was hard-coded to the final three cells of the active segment. Chevron rendering adds padding and chevrons around the label, shifting the visible close affordance one cell outside that region while placing the trailing seam inside it.

## Validation

Closes #787

### DIFF
--- a/crates/flotilla-tui/src/segment_bar.rs
+++ b/crates/flotilla-tui/src/segment_bar.rs
@@ -29,6 +29,12 @@ impl RenderedItem {
         Self { spans, width }
     }
 
+    /// Return the display-cell offset of the last `needle` in the rendered text.
+    pub(crate) fn last_cell_offset_of(&self, needle: &str) -> Option<usize> {
+        let text: String = self.spans.iter().map(|span| span.content.as_ref()).collect();
+        text.rfind(needle).map(|byte_offset| text[..byte_offset].width())
+    }
+
     pub fn empty() -> Self {
         Self { spans: vec![], width: 0 }
     }
--- a/crates/flotilla-tui/src/widgets/tabs.rs
+++ b/crates/flotilla-tui/src/widgets/tabs.rs
@@ -201,17 +201,23 @@ impl Tabs {
             BarKind::Pipe => Box::new(ThemedTabBarStyle { theme, site: &theme.tab_bar }),
             BarKind::Chevron => Box::new(ThemedRibbonStyle { theme, site: &theme.tab_bar }),
         };
+        let close_offset = (active_idx != 0)
+            .then(|| tab_style.render_item(&items[active_idx]).last_cell_offset_of(CLOSE_AFFORDANCE))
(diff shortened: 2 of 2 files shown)


rjwittams:
Thanks for the close read. I’m leaving both non-blocking nits as-is:

1. The extra `render_item` call is a deliberate local tradeoff. It is one pure, small span render per frame, and avoiding it would require broadening `segment_bar::render`/`HitRegion` with affordance-specific geometry that no other consumer needs.
2. `last_cell_offset_of` is crate-private implementation detail. The regression test covers it through the public behavior seam (`Tabs::render` followed by `handle_click`) across both rendering styles, duplicate `×` content, every tab cell, and clipping-safe hit geometry. That matches this repository’s documented preference for behavior tests over tests coupled to internal helpers.

No code change needed from these nits.

PR #797, opened 2026-07-21
fix: cap splash image at native resolution and log picked protocol

rjwittams:
## What

The splash was scaling the 1024x559 splash image up to fill the terminal area before transmission. Pixel protocols (kitty) transmit the rendered area's pixels as uncompressed RGBA (`f=32,t=d` — the only medium ratatui-image implements), so on a retina font this shipped 9.5MB of base64 escape data inside a 112x30 zellij pane and 43MB on a full-screen ghostty — for a 233KB source image. Through a multiplexer that must re-parse every byte, this made the splash glacial (measured ~10s ingest at ~1MB/s through zellij vs ~1.2s directly in ghostty).

- Cap the rendered area so the transmitted image never exceeds the image's native pixel size (no-op for halfblocks, which has no pixel cost). Cuts the payload to ~3MB in both cases and eliminates the upscale resize cost.
- Log the picked protocol, font size, query duration, and first-draw duration from the real splash path so `tui.log` answers "what did detection pick" without a debugging session.
- Add `examples/splash_probe.rs`: runs the same detection + draw with timings printed after terminal restore; run it inside a multiplexer and directly in the host terminal to localize where splash time goes (query / encode+transmit / placement backpressure).

Measured with the probe on kiwi (ghostty 1.2, zellij fork, font 16x34); numbers in the table are in the linked discussion context.

Follow-ups tracked separately: ratatui-image only implements `f=32,t=d` (no `o=z`, no `t=s`/`t=f`, no PNG) — upstream opportunity; zellij-side `t=d` ingest rate (~1MB/s) profiling happens in the zellij fork.

### DIFF
--- a/crates/flotilla-tui/examples/splash_probe.rs
+++ b/crates/flotilla-tui/examples/splash_probe.rs
@@ -1,0 +1,111 @@
+//! Measure image-protocol detection and splash render cost in the current
+//! terminal. Run inside a zellij pane and directly in the host terminal to
+//! compare:
+//!
+//! ```sh
+//! cargo run -p flotilla-tui --example splash_probe [query_timeout_ms]
+//! ```
+
+use std::time::{Duration, Instant};
+
+use ratatui_image::{
+    picker::{cap_parser::QueryStdioOptions, Picker},
+    StatefulImage,
+};
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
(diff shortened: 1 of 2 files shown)


rjwittams:
Both observations implemented: the cap math is extracted as `splash::splash_scale` with four unit tests (native cap, small-area fit, halfblocks uncapped, aspect preservation), and the probe now calls the same function so it mirrors production by construction — with a `--uncapped` flag retained deliberately, since the oversized payload is useful for stressing multiplexer ingest throughput (the zellij-side follow-up this PR's description mentions).

PR #799, opened 2026-07-21
fix: validate convoy agent adapter availability

### LINKED ISSUE #784: fix: convoy admission doesn't validate agent adapter availability; failures surface as generic 'work failure detected'
## What happens

Every convoy started from an issue during the #666 dogfood round failed ~7s after creation with `agent adapter codex unavailable for environment env-<convoy>-work` (`crates/flotilla-daemon/src/runtime.rs:1133`), and the convoy-level status flattened that to a generic `work failure detected` — the real message only lives on the per-work status, and the project page truncates it.

Root cause chain:

- The default workflow `single-agent-contained` runs one contained vessel with crew capability `code`.
- The seeded capability table hard-maps `code` → the `codex` adapter (`crates/flotilla-core/src/agent_adapter.rs:83`).
- The default docker placement policy provisions the contained environment from a stock `ubuntu:24.04` image, so no agent adapter of any kind can resolve there.

So the failure was 100% predictable at admission time, but admission (`admit_convoy_start`) validates project, workflow, and placement-policy containment — not whether the required agent adapter can resolve in the target environment. The user pays a create → reconcile → provision → fail round trip (and, combined with the auto-attach wait bug, a 60s freeze) for a convoy that could never run.

## Fix shape

- Admission-time validation: resolve each vessel's crew capability against the capability table and check the target environment / placement policy can supply the adapter; reject with a direct message ("workflow requires agent adapter `codex`, which is not available in placement `docker-on-…` (image `ubuntu:24.04`)").
- Failure surfacing: when a work fails, propagate the work's failure message into the convoy-level `status.message` instead of the generic `work failure detected`, so every surface (project page convoys panel, convoy view, CLI) shows the real reason without drill-down.

Separate product question (not this issue): what the dogfood-ready contained image should be — an image with agent tooling baked in vs. defaulting the workflow to an uncontained stance until one exists.

## Acceptance criteria

- [ ] Starting a convoy whose required adapter cannot resolve in the target environment fails at admission with a message naming the adapter and the environment/image
- [ ] A convoy that fails after admission carries the failing work's message in convoy-level `status.message`
- [ ] Existing failure surfaces (project page, convoy view) show that message

rjwittams:
## What changed

- validate every agent crew capability against the selected placement before persisting a convoy
- advertise host-direct adapters through live host capabilities and let Docker image recipes declare bundled adapters
- reject the stock `ubuntu:24.04` contained placement with an error naming the missing adapter, placement, and image
- propagate a failed work item's concrete message into convoy-level status

## Why

The default contained workflow resolves `code` to the `codex` adapter, but the stock Docker image contains no adapter. That made a guaranteed failure pass admission, spend time provisioning, and then surface only the generic `work failure detected` message.

Admission now fails immediately, while post-admission failures retain their actionable cause across existing project, convoy, and CLI views.

## Impact

Impossible convoy starts are rejected before persistence. Operators see the concrete adapter or work failure instead of waiting for provisioning and drilling into per-work status.

## Validation

- `TMPDIR=/private/tmp cargo test --workspace --locked`
- independent Standards and Spec reviews

Closes #784

### DIFF
--- a/crates/flotilla-controllers/tests/common/mod.rs
+++ b/crates/flotilla-controllers/tests/common/mod.rs
@@ -179,6 +179,7 @@ pub async fn create_docker_worktree_policy(backend: &ResourceBackend, namespace:
             .docker_per_vessel(DockerPerVesselPlacementPolicySpec {
                 host_ref: fixture.host_ref,
                 image: fixture.image,
+                agent_adapters: Default::default(),
                 default_cwd: fixture.default_cwd,
                 env: Default::default(),
                 checkout: DockerCheckoutStrategy::WorktreeOnHostAndMount { mount_path: fixture.mount_path },
--- a/crates/flotilla-controllers/tests/vessel_reconciler.rs
+++ b/crates/flotilla-controllers/tests/vessel_reconciler.rs
@@ -674,6 +674,7 @@ async fn multi_repository_docker_fresh_clone_uses_per_repository_paths() {
             .docker_per_vessel(DockerPerVesselPlacementPolicySpec {
                 host_ref: HOST_REF.to_string(),
                 image: "ghcr.io/flotilla/dev:latest".to_string(),
+                agent_adapters: Default::default(),
                 default_cwd: None,
                 env: Default::default(),
                 checkout: DockerCheckoutStrategy::FreshCloneInContainer { clone_path: "/workspace/".to_string() },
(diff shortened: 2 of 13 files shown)


rjwittams:
Fixed in `da40be5c`.

Both `ConvoyStart` and the still-reachable `ConvoyCreate` path now use the same workflow-aware placement resolver, including contained-placement and required-adapter validation before any repository or adopted-checkout side effects. The Docker regression test now exercises both command paths and asserts neither persists a convoy.

I also added host-direct coverage using a live `HostStatus` that advertises `claude-code` but not the required `codex` adapter. The full `flotilla-core` library suite, focused admission test, formatting, and workspace clippy all pass locally.

rjwittams:
Thanks for the re-review and confirmation that both findings are resolved.

No further PR code changes are needed. I filed #804 for the valid out-of-scope Docker placement-policy authoring gap: https://github.com/flotilla-org/flotilla/issues/804. The multiple-failure ordering observation remains deliberately non-blocking; we can choose a causal ordering rule when concurrent heterogeneous failures become a demonstrated requirement.

rjwittams:
Thanks for the final review and confirmation that the current diff has no remaining bugs or test gaps. The Docker placement-policy authoring gap remains tracked in #804; the multiple-failure ordering note is intentionally deferred pending a demonstrated causal-ordering requirement.

PR #800, opened 2026-07-21
fix: dispatch TUI commands outside event loop

### LINKED ISSUE #783: fix: TUI event loop blocks on command dispatch — any slow daemon ack freezes the UI
## What happens

The TUI event loop drains its command queue with a synchronous `dispatch(cmd).await` (`crates/flotilla-tui/src/run.rs:140-142` → `crates/flotilla-tui/src/app/executor.rs:56`). Any daemon request that is slow to ack freezes the entire UI — no key handling, no render — for up to the 30s client timeout (`crates/flotilla-client/src/lib.rs:512`). Buffered keys then replay against whatever is on screen, which is how one convoy-start attempt became four (see the ConvoyStart issue from the same dogfood round).

This is the client-side half of the convoy-start lockup, but it is a standing hazard independent of any one command: the fast-ack contract is only a convention, and a single slow handler (or a saturated daemon) freezes the TUI.

## Fix shape

Move command dispatch off the event loop: spawn the execute request and deliver its ack/error back as an app event, keeping `&mut App` free between submission and completion. Pending state already exists (`PendingAction` / `pending_actions`) for the in-flight indicator; the dispatch path should feed it asynchronously. Modal loading states that today rely on the synchronous error return (`reset_loading_mode`, `executor.rs:79`) need to reset from the async error event instead.

## Acceptance criteria

- [ ] No `daemon.execute(...).await` on the event-loop path; the UI renders and handles keys while a command ack is outstanding
- [ ] A deliberately delayed daemon ack (test double) leaves the TUI responsive and shows the pending indicator
- [ ] Command errors (including request timeout) surface as status messages and reset loading modals, as today

rjwittams:
## Summary

- dispatch daemon commands and attach queries from background tasks instead of awaiting them on the TUI event loop
- represent pre-ack submissions explicitly and reconcile command-finish/ack ordering
- preserve completion events across foreground terminal attaches and reset pending/loading state on errors
- add delayed-ack, rendered-spinner, ordering, timeout, and attach completion regression coverage

## Root cause

The command executor awaited daemon acknowledgements while holding the mutable `App` borrow on the event-loop path. Any delayed acknowledgement therefore stopped input handling and rendering until the daemon responded.

## Impact

The TUI remains responsive and continues rendering pending feedback while daemon acknowledgements are outstanding. Timeout and dispatch failures still surface through the status line and clear loading modals.

## Validation

Closes #783

### DIFF
--- a/crates/flotilla-tui/src/app/executor.rs
+++ b/crates/flotilla-tui/src/app/executor.rs
@@ -55,8 +55,17 @@ pub fn dispatch(cmd: Command, app: &mut App, pending_ctx: Option<PendingActionCo
     if let Some(ctx) = &pending_ctx {
         let action = PendingAction { status: PendingStatus::Submitting, description: ctx.description.clone() };
         if let Some(page) = app.screen.repo_pages.get_mut(&ctx.repo_identity) {
+            if page
+                .pending_actions
+                .get(&ctx.identity)
+                .is_some_and(|pending| matches!(pending.status, PendingStatus::Submitting | PendingStatus::InFlight { .. }))
+            {
+                app.model.status_message = Some("An action is already in progress for this item".into());
+                return;
+            }
             page.pending_actions.insert(ctx.identity.clone(), action);
         }
+        app.pending_dispatch_acks += 1;
     }
 
     let daemon = app.daemon.clone();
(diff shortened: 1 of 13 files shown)


rjwittams:
Fixed both concerns in 1dc5ccde.

The early-finish cache is now a map retained only while this TUI has pending-action acknowledgements. It has no arbitrary capacity eviction, explicitly accepts that lifecycle traffic is system-wide, and clears as soon as the last local acknowledgement is reconciled. The new `unrelated_finishes_cannot_evict_a_finish_awaiting_its_ack` regression covers the original finish followed by 66 unrelated finishes before its acknowledgement.

Dispatch now also rejects a second Submitting/InFlight action for the same work-item identity before it reaches the daemon, preserving the first command's pending state. `dispatch_rejects_a_second_pending_action_for_the_same_identity` covers that path.

Pinned formatting, workspace Clippy, and the full TUI suite (942 unit tests plus integration and snapshot suites) pass locally.

rjwittams:
Reviewed; no further code change needed for the two explicitly non-blocking observations.

`DaemonHandle::execute` reports operational failures and client timeouts through `Result`, which the completion event already handles. Recovering an unexpected panic inside a spawned daemon future would introduce a new panic-recovery policy not used elsewhere in these crates; such a panic indicates a broader invariant failure rather than an ordinary dispatch outcome.

The missing-page counter behavior is intentional pairing: a supplied pending context increments exactly once and its completion decrements exactly once even if the view disappears or was already absent. UI updates then safely no-op, while acknowledgement/finish reconciliation remains balanced.

PR #801, opened 2026-07-21
feat: add convoy delete action

### LINKED ISSUE #785: feat: delete/cancel action for convoys (TUI row action + CLI verb)
## What to build

There is currently no way to get rid of a failed convoy from any user surface. After the #666 dogfood round, four `Failed` convoys sit in the store permanently:

- The convoy table projection exposes no row actions at all (`crates/flotilla-tui/src/widgets/table_view.rs:621-623`, `actions: &[]`) — no delete/cancel/retry.
- The `convoy` CLI noun has only `work`, `start`, `create` (`crates/flotilla-commands/src/commands/convoy.rs:23-26`).
- The reconciler tears down child vessels/presentations on failure but leaves the Convoy resource itself (`crates/flotilla-resources/src/convoy/reconcile.rs:184-190`). Deletion exists only at the ResourceClient/backend layer, unwired to any surface.

Add convoy deletion end-to-end: a `flotilla convoy delete <name>` CLI verb and a TUI row action on convoy rows (action menu + delete-confirm flow, consistent with the existing `d` delete grammar), both routed through resource deletion so the `flotilla.work/convoy-teardown` finalizer runs the teardown path. Cancel-while-running can ride the same path if deletion of an active convoy triggers teardown correctly; if that needs separate design, scope this issue to terminal-phase (Failed/Completed) convoys and note the follow-up.

## Acceptance criteria

- [ ] `flotilla convoy delete <name>` removes a Failed convoy (finalizer teardown runs, resource gone from store and all views)
- [ ] Convoy rows in the TUI offer a delete action with the standard confirm flow
- [ ] Deleting a convoy cleans up its child resources (vessels/environments/terminal sessions) via the existing teardown finalizer

rjwittams:
## Summary

- add `flotilla convoy delete <name>` to the CLI and protocol
- route convoy deletion through the resource resolver so existing finalizers clean up managed children
- add a host-aware Convoy row Delete action in the TUI with confirmation before dispatch

## Why

Convoys could be created and completed, but users had no direct CLI or TUI action to delete or gracefully cancel one. Using the existing deletion/finalizer path keeps active and terminal convoy cleanup consistent.

## User impact

Users can delete a convoy from either the CLI or its TUI row. Active convoys follow the same finalizer path, tearing down managed vessels, presentations, checkouts, environments, and terminal sessions.

## Validation

- parallel standards and spec review; all findings addressed

Closes #785

### DIFF
--- a/crates/flotilla-commands/src/commands/convoy.rs
+++ b/crates/flotilla-commands/src/commands/convoy.rs
@@ -23,6 +23,11 @@ pub struct ConvoyNoun {
 pub enum ConvoyVerb {
     /// Manage the work aboard a convoy's vessels
     Work(ConvoyWorkNoun),
+    /// Delete a convoy and tear down its managed resources
+    Delete {
+        /// Convoy resource name
+        name: String,
+    },
     /// Start a convoy through Project-scoped admission completion
     Start {
         /// Project whose definitions and repository set admission snapshots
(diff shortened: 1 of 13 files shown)


rjwittams:
Reviewed the observations; no code change is needed for this PR. The confirmation constructor intentionally fails fast on an internal invariant and has a single typed call site, CLI namespace selection matches the existing convoy verb convention, and the positional-subject guard follows the established `convoy start` shape. Thanks for the review.

PR #802, opened 2026-07-21
fix: make project panels directly navigable

### LINKED ISSUE #788: fix: project-page Issues panel fetch-on-scroll traps downward panel navigation
## What happens

On the project page, paging down through the Issues panel never reaches the Independents panel below it — downward keyboard navigation is trapped in Issues. (Reported during #666 dogfood as the page feeling "permanently focused into the issues section".)

## Diagnosis

The Issues panel is demand-backed with fetch-on-scroll: `SelectNext` fires `fetch_more_query` (`crates/flotilla-tui/src/widgets/project_page.rs:270-281`) and `should_fetch_more` triggers within 5 rows of the bottom (`crates/flotilla-tui/src/widgets/table.rs:46-49`). For a large issue source (flotilla has hundreds of issues), every approach to the bottom loads another page, so the "last row" — the crossing point into the next panel — recedes indefinitely. Upward navigation (`k`) and mouse clicks on panel headers still cross panels (`project_page.rs:341-357`), which is the current escape hatch.

## Fix shape

Panel crossing shouldn't depend on exhausting the data. Options (pick during fix): a dedicated next/previous-panel binding (e.g. Tab/S-Tab or J/K) that crosses regardless of row position; or make `SelectNext` at the bottom of the *loaded* rows cross into the next panel when a fetch is already pending rather than waiting. A paging-parity pass with the old repo-page section navigation grammar would resolve it — the repo page had explicit section-jump keys.

## Acceptance criteria

- [ ] Keyboard-only navigation can reach every project-page panel with a bounded number of keypresses, independent of issue count
- [ ] Fetch-on-scroll still works for actually browsing deep into the issue list

rjwittams:
## Summary

- add direct Tab / Shift+Tab navigation between panels on Project pages
- keep `j`/`k` row browsing and near-bottom issue fetches unchanged
- expose configurable `next_panel` / `prev_panel` actions and document the new navigation
- add a regression test that simulates a fast-growing issue window

## Root cause

The Issues panel prefetches within five rows of the loaded window's end. When each fetch completes before the next downward keypress, the last loaded row recedes at the same rate as selection, so row navigation never reaches the crossing point into Independents.

Panel navigation is now orthogonal to row exhaustion and network timing: Tab and Shift+Tab move directly to the adjacent panel header.

## Validation

- `cargo clippy -p flotilla-core -p flotilla-tui --all-targets --locked -- -D warnings`
- `TMPDIR="$PWD/.codex-tmp" cargo test -p flotilla-tui --locked` (938 unit, 2 E2E, 1 high-fidelity, 50 snapshot tests)
- `TMPDIR="$PWD/.codex-tmp" cargo test -p flotilla-core --locked config` (49 tests)

Closes #788

### DIFF
--- a/crates/flotilla-core/src/config.rs
+++ b/crates/flotilla-core/src/config.rs
@@ -133,6 +133,8 @@ pub struct KeysConfig {
     #[serde(default)]
     pub convoys: HashMap<String, String>,
     #[serde(default)]
+    pub project: HashMap<String, String>,
+    #[serde(default)]
     pub convoy_vessels: HashMap<String, String>,
     #[serde(default)]
     pub action_menu: HashMap<String, String>,
--- a/crates/flotilla-tui/src/app/view_kind.rs
+++ b/crates/flotilla-tui/src/app/view_kind.rs
@@ -67,7 +67,7 @@ pub(crate) fn kind_modes(address: Option<&ViewAddress>) -> Vec<BindingModeId> {
         ) => {
             vec![BindingModeId::Convoys]
         }
-        Some(ViewAddress::Project { .. }) => vec![BindingModeId::Convoys, BindingModeId::DemandTable],
+        Some(ViewAddress::Project { .. }) => vec![BindingModeId::Convoys, BindingModeId::DemandTable, BindingModeId::Project],
         Some(ViewAddress::Issues { .. }) => vec![BindingModeId::Convoys, BindingModeId::DemandTable],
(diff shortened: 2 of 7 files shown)


rjwittams:
Confirmed: production key resolution and status hints for a Project view are driven by the composed modes in `view_kind`, as noted. Keeping `ProjectPageWidget::binding_mode()` aligned with the widget's actual interaction vocabulary preserves the `InteractiveWidget` contract for direct embedding and harness use, so no code change is needed for this non-blocking observation.

PR #803, opened 2026-07-21
fix: decode view labels and restore panel dividers

### LINKED ISSUE #790: fix: project-page panel headers — percent-encoded addresses leak into UI; adopt repo-page divider chrome
## What happens

Two cosmetic dogfood findings about the project page's panel headers, one fix unit:

1. **Percent-encoding leaks into the UI.** Headers render e.g. `Independents › independents?project=flotilla%2Fflotilla`. The `%2F` is not a stray web-ism in the data — it's the `ViewAddress` `Display` impl (`crates/flotilla-protocol/src/view_address.rs:96-137`), which percent-encodes because `Display`/`FromStr` are the round-trippable wire/persistence contract (serde via `collect_str`, `open-views.toml`, materialize recipes; ADR 0013 grammar). The bug is using the wire form as display text: `crates/flotilla-tui/src/widgets/project_page.rs:208` formats `layout.panel.target` straight into the header.

2. **Headers look plain vs the old repo page.** Old repo-page section headers are a full-width divider rule `── Label ──` (`crates/flotilla-tui/src/widgets/split_table.rs:789-800`, muted dashes framing a `section_header`-colored label). New project-page headers are a bare bold `Paragraph` (`theme.header_style()`, `crates/flotilla-tui/src/theme.rs:258-260`, +REVERSED when focused) — richer metadata (`· as of…`, `· more available`, `· ⚠ conditions`) but no chrome.

## Fix shape

- Add a `human_label()` (decoded, human-readable) form to `ViewAddress` — the scope fields are already stored decoded (`QueryScope { namespace, name }`) — and use it in panel headers (and anywhere else an address is shown to a human, e.g. tab labels, palette completions). **Do not touch `Display`/`FromStr`** — that's the canonical grammar.
- Bring the divider-rule chrome to the project-page panel headers so sections read like the old repo page, keeping the new metadata suffixes.

## Acceptance criteria

- [ ] No percent-encoded text anywhere in rendered UI; addresses shown to humans use the decoded form
- [ ] `ViewAddress` `Display`/`FromStr` round-trip unchanged (existing tests still pin the grammar)
- [ ] Project-page panel headers use the divider chrome consistent with the repo page, retaining the as-of/more/conditions metadata

rjwittams:
Closes #790

## What changed

- added a decoded, human-facing `ViewAddress::human_label()` while preserving canonical `Display`/`FromStr` behavior
- used decoded labels in project headers, breadcrumbs, human CLI output, and user-visible address errors
- extracted the repo-page divider renderer and reused it for project-panel headers
- retained panel metadata and focused-header styling

## Root cause

Human-facing TUI surfaces formatted `ViewAddress` through its canonical persistence/wire `Display` implementation, which correctly percent-encodes reserved characters. Project panels also used plain paragraph headers instead of the established section-divider chrome.

## Validation

- focused protocol and TUI render tests after rebasing onto current `main`

### DIFF
--- a/crates/flotilla-protocol/src/view_address.rs
+++ b/crates/flotilla-protocol/src/view_address.rs
@@ -76,6 +76,35 @@ impl ViewAddress {
             Self::Repo { .. } => "repo",
         }
     }
+
+    /// A decoded address-shaped label for human-facing surfaces.
+    ///
+    /// Unlike [`fmt::Display`], this form is not a persistence or parsing
+    /// contract: reserved characters remain readable instead of being
+    /// percent-encoded.
+    pub fn human_label(&self) -> String {
+        match self {
+            Self::Overview => "overview".to_string(),
+            Self::Convoys { namespace } => format!("convoys/{namespace}"),
+            Self::Independents { scope: Some(scope) } => {
+                format!("independents?project={}/{}", scope.namespace, scope.name)
+            }
+            Self::Independents { scope: None } => "independents".to_string(),
+            Self::Convoy { namespace, name } => format!("convoy/{namespace}/{name}"),
+            Self::Vessel { namespace, convoy, vessel } => format!("vessel/{namespace}/{convoy}/{vessel}"),
+            Self::Project { namespace, name } => format!("project/{namespace}/{name}"),
+            Self::Issues { scope } => format!("issues?project={}/{}", scope.namespace, scope.name),
+            Self::Checkouts { scope: Some(scope) } => format!("checkouts?project={}/{}", scope.namespace, scope.name),
+            Self::Checkouts { scope: None } => "checkouts".to_string(),
+            Self::Repo { identity, repository_key } => {
+                let mut label = format!("repo/{}/{}", identity.authority, identity.path);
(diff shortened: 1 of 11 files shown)


rjwittams:
Fixed in a976fe0c. The shared divider now accepts a row style separately from its label style, so focused project headers retain the previous full-row REVERSED affordance while keeping the new divider chrome. Added `focused_panel_header_reverses_the_full_divider_row` to cover both ends of the header row; project-page tests, pinned formatting, and TUI clippy all pass locally.

rjwittams:
Checked the active-but-not-header-focused state and am keeping the current styling intentionally. The issue asks project headers to adopt the established repo-page divider chrome, where muted dashes frame the styled label; the full-row REVERSED treatment is specifically the keyboard/header-focus affordance. Applying the active label style across the dashes would weaken that convention and make active-but-unfocused compete with actual header focus. No code change needed for this non-blocking note.

rjwittams:
Acknowledged. The rebase preserves both the upstream panel-navigation regression test and this PR's full-row focus regression test; no further code changes are needed.

PR #815, opened 2026-07-21
fix: harden convoy-start cleanup against worker panics

### LINKED ISSUE #796: Harden convoy-start dedup cleanup against worker panics
From PR #794 review by @claude: https://github.com/flotilla-org/flotilla/pull/794#issuecomment-5038704934

`pending_convoy_starts` is removed after the asynchronous convoy-start worker returns normally. The current admission, watch, and attach paths propagate fallible operations through `Result` and have no known panic path, so this is not a live blocker for #782.

Add defensive cleanup at the spawned-task boundary so an unexpected future panic cannot permanently strand the deduplication key or leave the initiating command without a `CommandFinished` event. Consider a drop guard or an unwind boundary, and add a regression test that injects a worker panic while proving a subsequent matching convoy start can proceed.

Context: the deduplication set is an admission gate, so leaked entries block matching starts until daemon restart.

rjwittams:
## Summary

- supervise asynchronous convoy-start workers with an unwind boundary
- clear the pending deduplication key before publishing command completion
- turn unexpected worker panics into a visible `CommandFinished` error
- add regression coverage proving an identical start can retry after a panic

Closes #796

## Verification

- `cargo test -p flotilla-core --locked --features test-support --test in_process_daemon convoy_start_`
- `cargo test --workspace --locked -- --skip server::tests::dispatch_add_list_remove_repo_round_trip`

The unfiltered workspace suite has one unrelated deterministic macOS temp-path assertion mismatch (`/private/var/...` versus `/var/...`) in `dispatch_add_list_remove_repo_round_trip`; the isolated test fails identically on `main`-equivalent code.

### DIFF
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -5,6 +5,7 @@
 
 use std::{
     collections::{BTreeMap, BTreeSet, HashMap, HashSet},
+    panic::AssertUnwindSafe,
     path::{Path, PathBuf},
     sync::{
         atomic::{AtomicU64, Ordering},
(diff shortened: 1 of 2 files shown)


rjwittams:
Reviewed; no code changes needed. The review found no blocking issues. The two observations are accurate but outside #796's contract: this change guarantees panic-safe dedup cleanup and a `CommandFinished` result, while strict ordering between separate commands and rollback of hypothetical partial resource mutations would require separate transactional semantics. The current admission path has no known panic after partial persistence.

PR #821, opened 2026-07-22
fix: qualify presentation tabs with convoy identity

### LINKED ISSUE #808: fix: presentation tab named after bare vessel name — every single-vessel convoy's tab is 'work'
Observed in the first successful convoy-from-issue run (2026-07-21, flotilla-org/flotilla#796's convoy): the zellij tab for the convoy is named **work**.

`create_presentation_actuation` sets `PresentationSpec.name = vessel.to_string()` (`crates/flotilla-resources/src/convoy/reconcile.rs:752`), and the seeded `single-agent-contained` workflow names its only vessel `work` — so every convoy started from the default workflow gets an identical, meaningless tab name, and two concurrent convoys produce two indistinguishable 'work' tabs. The convoy identity rides only in the process-selector labels, not the display name.

Presentations are per-vessel (one tab per vessel, panes filtered to that vessel's crew), so the name needs the convoy for identity and the vessel only for disambiguation:

- Single-vessel workflow: tab name = convoy name
- Multi-vessel workflow: `<convoy>:<vessel>`

This mirrors the TUI's tab-label convention (short by default, widen with qualifiers only when two labels would otherwise collide).

Note for later, out of scope: AI-generated convoy names run long for tab real estate (`convoy-start-dedup-panic-cleanup` is 31 chars); a short display alias generated alongside the convoy name would help every surface, and the andamento-style convoy-tab-with-vessel-tree is future presentation work.

Acceptance:
- [ ] Single-vessel convoy tab is named after the convoy
- [ ] Multi-vessel convoy tabs are `<convoy>:<vessel>`
- [ ] Two concurrent convoys have distinguishable tabs

rjwittams:
## Summary

- name single-vessel presentation tabs after the convoy
- name multi-vessel tabs as `<convoy>:<vessel>`
- cover distinct concurrent convoys and persisted multi-vessel presentations

## Root cause

Presentation creation used the bare vessel name as `PresentationSpec.name`. The default workflow calls its only vessel `work`, so every convoy launched from that workflow produced an indistinguishable `work` tab even though convoy identity was already present in selectors and labels.

## Impact

Presentation tabs now carry convoy identity while retaining the vessel qualifier only when it is needed to distinguish multiple vessels in the same convoy.

## Validation

- `TMPDIR=/private/tmp cargo test --workspace --locked`

Fixes #808

### DIFF
--- a/crates/flotilla-resources/src/convoy/reconcile.rs
+++ b/crates/flotilla-resources/src/convoy/reconcile.rs
@@ -730,6 +730,17 @@ fn create_vessel_outcome(convoy: &ResourceObject<Convoy>, vessel: &str, _now: Da
 }
 
 fn create_presentation_actuation(convoy: &ResourceObject<Convoy>, vessel: &str) -> Actuation {
+    let presentation_name = if convoy
+        .status
+        .as_ref()
+        .and_then(|status| status.workflow_snapshot.as_ref())
+        .is_some_and(|snapshot| snapshot.vessels.len() == 1)
+    {
+        convoy.metadata.name.clone()
+    } else {
+        format!("{}:{vessel}", convoy.metadata.name)
+    };
+
     Actuation::CreatePresentation {
         meta: InputMeta::builder()
             .name(vessel_resource_name(&convoy.metadata.name, vessel))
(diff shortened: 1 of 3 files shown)


rjwittams:
Thanks for the review. No code changes needed for the two non-blocking observations:

- The workflow snapshot is immutable after bootstrap, so the persisted convoy status and the cleanup prediction cannot disagree about vessel count. Passing predicted status through would widen the helper signature without changing behavior.
- The single-vessel transformation is used once and deliberately shows the three correlated structures being narrowed together. It already starts from the shared fixture; extracting a one-use helper would hide that setup without adding reuse.

The naming behavior and coverage therefore remain as reviewed.

PR #826, opened 2026-07-22
feat: run host-direct convoys on remote peers

### LINKED ISSUE #812: feat: remote convoy placement — dispatch to feta/udder via host-direct placement policies
## What to build

Parallelism: all convoys currently place host-direct on the local host (kiwi). Feta and udder are peered, run current builds, and idle. Make convoy placement target remote hosts: host-direct placement policies for peer hosts, admission validating the remote host's advertised agent adapters (the Host status capability set — the same mechanism #799/#806 fixed locally), checkout provisioning on the remote host's checkout root, and attach routed via the existing multi-host command routing.

This is the first real test of the multi-host control plane doing work (not just observing). The host/environment model design (#625) is the deep version; this slice uses what exists: Host resources with capability status + placement policies + remote command dispatch (#273-era plumbing).

## Acceptance criteria

- [ ] A placement policy targeting feta or udder admits a convoy whose adapter the remote host advertises
- [ ] The vessel's checkout and terminal sessions provision on the remote host; attach works from kiwi's TUI
- [ ] Three convoys running simultaneously across kiwi+feta+udder, all visible on the project page with correct host attribution

## Blocked by

None hard; #625 informs but this slice deliberately precedes it.

rjwittams:
## Summary

- advertise each peer's agent adapters and terminal pools as host capabilities
- materialize ready peer `Host` resources and host-direct placement policies
- admit remote convoy starts on the presentation host, then forward immutable workflow, placement, and convoy snapshots to the selected execution host
- route remote auto-attach results back through the selected host and preserve host attribution in fleet views
- reject placement/target mismatches and prevent clients from bypassing admission with prepared peer commands

## Why

Convoy placement previously assumed that the daemon receiving the user command also owned the checkout and terminal session. A host-direct policy targeting a connected peer therefore could not safely launch there, and remote attach commands were not usable from the presentation host.

This change makes peer capability discovery feed placement admission and gives remote convoy launch an explicit, validated peer boundary without depending on identically named resources in two daemon-local stores.

## Validation

- focused core tests for peer admission/routing and simultaneous kiwi/feta/udder fleet attribution
- focused daemon tests for remote snapshot persistence and routed auto-attach
- `cargo test -p flotilla-protocol --locked`
- workspace clippy passed before the final review fixes; GitHub CI will run the complete repository gates

Closes #812

### DIFF
--- a/crates/flotilla-core/src/executor.rs
+++ b/crates/flotilla-core/src/executor.rs
@@ -339,6 +339,7 @@ pub async fn build_plan(
         | CommandAction::CrewHandoff { .. }
         | CommandAction::ConvoyCreate { .. }
         | CommandAction::ConvoyStart { .. }
+        | CommandAction::ConvoyStartPrepared { .. }
         | CommandAction::WorkflowTemplateApply { .. }
         | CommandAction::ProjectAdd { .. }
         | CommandAction::ProjectApply { .. }
--- a/crates/flotilla-core/src/host_registry.rs
+++ b/crates/flotilla-core/src/host_registry.rs
@@ -117,6 +117,10 @@ impl HostRegistry {
         self.node_environments.read().await.get(node_id).cloned()
     }
 
+    pub(crate) async fn node_id_for_environment(&self, environment_id: &EnvironmentId) -> Option<NodeId> {
+        self.hosts.read().await.get(environment_id).filter(|state| !state.removed).map(|state| state.node_id.clone())
+    }
+
     pub(crate) async fn get_host_status(
         &self,
         environment_id: &EnvironmentId,
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -1778,12 +1778,10 @@ impl InProcessDaemon {
             .get(policy_name)
             .await
             .map_err(|error| format!("placement policy {policy_name}: {error}"))?;
-        let host_ref = policy
-            .spec
-            .host_direct
-            .as_ref()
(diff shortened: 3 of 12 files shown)


rjwittams:
Addressed the two lifecycle findings in `ed3194b9`:

1. Prepared workflow and placement resources are now content-addressed, so deleting and reusing a convoy name with changed specs selects new snapshot names instead of colliding with stale resources. A regression test verifies the names change with content.
2. Auto-materialized peer policies now reconcile their selected terminal pool when the peer's healthy pool advertisement changes, with coverage in the peer-summary test.

For sleep survival, the execution host now owns the durable convoy, checkout, and terminal-session resources; the dispatching host is not in their runtime path. Wake-up reconvergence continues through the existing replica refresh/generation replacement path (including stale/unreachable preservation), rather than new cross-root replication. The new three-host fleet test covers host attribution, while the existing stale/refresh tests cover the observer losing and refreshing its remote view. This PR therefore keeps `Closes #812`; live lid-close validation remains an operational acceptance check of those existing reconnect paths, not a deferred implementation slice.

rjwittams:
Fixed in `e10794f6`. Non-host-direct policies now resolve to local execution (`Ok(None)`) instead of being rejected by the remote-routing layer. The peer placement test now includes a pool-only placement policy and verifies it remains local, preserving the existing docker/pool admission path.

rjwittams:
Agreed this is a valid display/staleness follow-up rather than an execution-safety blocker. Filed #832 to make persisted peer Host readiness track disconnect/reconnect while retaining useful last-known capabilities: https://github.com/flotilla-org/flotilla/issues/832

PR #842, opened 2026-07-22
fix: sort project issues by id

### LINKED ISSUE #824: fix: project-page issues panel sorts by last-updated — default should be id
The project page's Issues panel currently orders by last-updated, which makes the list churn as activity touches old issues and makes 'where is #NNN' scanning impossible. Default sort should be by issue id (descending feels natural — newest filed first — but pick one and make it stable); last-updated can return later as an optional sort once panel-level sort configuration exists (that's #813-adjacent surface, not this fix).

From Robert's dogfood notes 2026-07-22 ('soon').

Acceptance:
- [ ] Issues panel default-sorted by id, stable across refreshes
- [ ] No churn of row positions when unrelated issue activity updates timestamps

rjwittams:
Closes #824.

Project issue rows now use a stable descending ID order across initial materialization, refreshes, and client-side deltas. Numeric IDs sort numerically; opaque IDs have a deterministic fallback order.

### DIFF
--- a/crates/flotilla-core/src/query_registry.rs
+++ b/crates/flotilla-core/src/query_registry.rs
@@ -104,7 +104,7 @@ impl QueryRegistry {
             by_reference.insert(row.reference.clone(), row.clone());
         }
         *rows = by_reference.into_values().collect();
-        rows.sort_by(|left, right| right.issue.as_of.cmp(&left.issue.as_of).then_with(|| left.reference.cmp(&right.reference)));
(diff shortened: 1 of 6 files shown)


rjwittams:
The mixed-format comparator concern is valid but outside this GitHub-only fix. I filed https://github.com/flotilla-org/flotilla/issues/848 to define and test a transitive numeric/opaque ordering before another issue provider is added.

PR #855, opened 2026-07-22
fix: resolve convoy PR refs from peer data

### LINKED ISSUE #846: fix: convoy PR references only resolve for local-vessel convoys
Observed 2026-07-22 with five convoys running (all dispatched from kiwi, all in kiwi's store): every convoy branch has an **open PR** —

| convoy | vessel host | PR | TUI shows ref |
|---|---|---|---|
| socket-path-sunlen-fix | kiwi | #844 | yes |
| issues-panel-default-sort-id | kiwi | #842 | yes |
| convoy-lifecycle-teardown | feta | #839 | **no** |
| crew-handoff-provenance | feta | #841 | **no** |
| multi-select-dispatch | feta | #845 | **no** |

— yet the convoy rows show PR references only where the vessel is local. The #827 resolver (`resolve_convoy_change_request`) is branch-keyed against the *local tracked repo's* change-request provider and the convoy's admission-snapshotted `repository_keys`, all of which are identical across the five, so host should be irrelevant; 3/3 vs 2/2 says something host-correlated leaks into the path.

Hypotheses to check first:
- refresh triggering: `refresh_change_request` on convoy events + `RepoRefreshCompleted` → does refresh-all actually re-resolve convoys that previously resolved to None, or does something (retain/remove logic in `convoy_change_requests`) skip them?
- ResourceRef/host qualification: rows for remote-vessel convoys built through a path that keys `convoy_change_requests` differently (host-qualified ref) than the resolver stored.
- provider selection filtered by vessel/checkout host rather than repository key.

Note: whichever it is, add the missing regression test with a remote-vessel convoy — #827's tests presumably only cover local vessels, same class of gap as #837.

Acceptance:
- [ ] Remote-vessel convoy with an open PR shows the ref, same latency as local
- [ ] Regression test with a remote-placed convoy in the fixture

rjwittams:
## What changed

- Resolve convoy change-request refs from cached merged provider data when the repo is virtual or peer-backed, while preserving live provider lookup for local tracked roots.
- Retry convoy PR resolution when repo snapshots or deltas arrive, not only on `RepoRefreshCompleted`.
- Add regression coverage for a virtual remote repo with peer PR data and for the aggregator snapshot refresh trigger.

## Why

Remote-placed convoys can be persisted on the target daemon where the matching repository is represented as a virtual/peer-backed repo. The previous resolver only looked at the preferred local root's change-request providers, so those convoys could resolve to `None` even though peer provider data already contained the open PR.

Closes #846

## Validation

### DIFF
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -115,6 +115,10 @@ fn host_name_for_provider_environment(providers: &ProviderData, environment_id: 
     candidates.into_iter().next()
 }
 
+fn cached_change_request_by_branch(providers: &ProviderData, branch: &str) -> Option<(String, flotilla_protocol::ChangeRequest)> {
+    providers.change_requests.iter().find(|(_, request)| request.branch == branch).map(|(id, request)| (id.clone(), request.clone()))
+}
+
 fn static_ssh_environment_id(config_key: &str) -> EnvironmentId {
     let mut encoded = String::with_capacity(config_key.len() * 2);
     for byte in config_key.as_bytes() {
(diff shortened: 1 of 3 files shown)


rjwittams:
Fixed the hot-loop concern in the follow-up commit. The aggregator now fingerprints each repo's change-request provider data and only re-resolves convoy PR refs when a RepoSnapshot changes that fingerprint or a RepoDelta carries Change::ChangeRequest entries. Empty and unchanged snapshots no longer wake refresh_all_change_requests(), and there is a regression test covering that unchanged-snapshot path.

I left the peer-cache terminal-state behavior as an explicit trade-off of the cached provider data path: live local trackers still win first and continue to query state=all, while peer/virtual rows can only reflect the mirrored provider cache currently available from the remote vessel.

rjwittams:
Confirmed intentional for this patch. The cached fallback is meant to keep convoy rows decorated from the best mirrored data available when the live path cannot produce a match. The refresh cadence should correct stale cache hits, and surfacing provider errors from that mixed live/cache path would need a separate UI/error-state decision. I would handle that as follow-up rather than expanding this fix.

PR #856, opened 2026-07-22
fix: retire crew sessions during convoy teardown

### LINKED ISSUE #852: convoy teardown leaves crew terminal sessions running
## What happened

After the first gated convoy deletes on the new #839 teardown path (feta, 2026-07-22, convoys `multi-select-dispatch` and `flotilla-resource-inspect`): both checkout worktrees were correctly verified (Clean ∧ Pushed ∧ Landed) and removed, but both crew cleat sessions survived the teardown — `terminal-multi-select-dispatch-work-coder` was left as a detached codex process whose cwd is a deleted directory.

## What to build

Vessel/convoy teardown should retire the crew's terminal sessions as part of the same settlement:

- On convoy delete (gated or `--force`), after checkout removal succeeds, `cleat kill` each crew session owned by the convoy's vessels (recordings survive kill by default — do not `--purge`).
- An **attached** session should not be killed silently out from under a viewer: detach-then-kill or surface it in the refusal/summary — decide which during implementation and record the choice on this issue.
- Adopted-checkout release (which never deletes the checkout) should likewise leave sessions alone.

## Acceptance criteria

- [ ] Deleting a completed convoy removes its worktrees AND its crew cleat sessions; `cleat list` on the placement host shows no `terminal-<convoy>-*` sessions afterwards.
- [ ] Session recordings are preserved (kill, not purge).
- [ ] Attached-session behaviour is decided, implemented, and noted here.
- [ ] Covered by a behavior test through injected collaborators (no real cleat needed).

## Context

ADR 0017 teardown contract (docs/adr/0017-convoy-completion-claims-conditions-attention.md); #839 implemented the checkout gates; #805 covered worktree residue. Session residue is the remaining leak.

rjwittams:
## What changed

- resolve terminal teardown from the persisted `TerminalSession` pool and environment instead of daemon-local runtime state
- retire the Cleat session through `cleat kill`, preserving its recording
- warn with convoy/vessel context before retiring an attached agent session
- cover both finalizer delegation and post-restart teardown with injected collaborators

## Why

After a daemon restart, the previous in-memory session registry was empty, causing terminal finalization to skip `cleat kill` and leave crew processes in deleted worktrees.

Closes #852.

## Validation

### DIFF
--- a/crates/flotilla-controllers/src/reconcilers/terminal_session.rs
+++ b/crates/flotilla-controllers/src/reconcilers/terminal_session.rs
@@ -32,7 +32,7 @@ pub trait TerminalRuntime: Send + Sync {
     ) -> Result<(), String> {
         Err("terminal runtime does not support crew message delivery".to_string())
     }
-    async fn kill_session(&self, session_id: &str) -> Result<(), String>;
+    async fn kill_session(&self, session_id: &str, spec: &flotilla_resources::TerminalSessionSpec) -> Result<(), String>;
 }
 
 pub struct TerminalSessionReconciler<R> {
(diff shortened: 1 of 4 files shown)


rjwittams:
Fixed in c0a9ef8d. Liveness-tracking pools now skip `kill_session` when the named session is already absent, preserving finalizer idempotency and avoiding a controller crash loop. A listing failure is logged but does not prevent the kill attempt. Added a regression test for the already-gone session path.

PR #867, opened 2026-07-22
fix: state the assignment in crew briefs instead of disclaiming it

rjwittams:
Observed live (wave 4, convoy-attention-signal): a crew brief for an issue-carrying convoy said `## Assignment` → "No additional assignment was provided." followed by the issue under `## Issue snapshot` — and the codex crew rationally read the disclaimer as overriding the contract below it, deciding it had nothing to do until cajoled.

- `build_crew_brief` now takes a `CrewAssignment` enum (`Prompt` / `CarriedIssue` / `Unassigned`) instead of `Option<&str>`: 'no ad-hoc prompt' splits into two very different situations and the old signature conflated them.
- `CarriedIssue` briefs state: your assignment is the issue below, its body is the contract.
- `Unassigned` briefs say so and forbid inventing work.
- `## Issue snapshot` renamed `## Assigned issue` (snapshot-as-of line retained inside) — the old header read as informational context.

Tests cover all three variants. Note: `server::tests::dispatch_add_list_remove_repo_round_trip` fails on macOS locally (`/private/var` vs `/var` canonicalization) on clean main too — pre-existing, unrelated, filed separately.

### DIFF
--- a/crates/flotilla-controllers/src/reconcilers/vessel.rs
+++ b/crates/flotilla-controllers/src/reconcilers/vessel.rs
@@ -1,7 +1,7 @@
 use std::{collections::BTreeMap, marker::PhantomData};
 
 use chrono::{DateTime, Utc};
-use flotilla_core::agent_adapter::{build_crew_brief, CrewBriefMember};
+use flotilla_core::agent_adapter::{build_crew_brief, CrewAssignment, CrewBriefMember};
 use flotilla_resources::{
     clone_key,
     controller::{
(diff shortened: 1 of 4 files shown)


rjwittams:
Both taken: (1) handoff_crew_brief now receives the convoy, computes the same three-way CrewAssignment match, and — the deeper half of your finding — appends the convoy work context via the now-shared append_convoy_work_context (moved from vessel.rs into core), so a CarriedIssue handoff brief actually contains the `## Assigned issue` section it points at, and `## Human instruction` is no longer dead guidance in that path. (2) Added issue_carrying_convoy_without_prompt_assigns_the_issue_in_the_brief at the reconciler level, pinning the CarriedIssue selection end-to-end.

rjwittams:
Residual gap filed as #869 (vessel-scoped repos in handoff briefs) — convoy cargo, not blocking here.

Issue #650, opened 2026-07-07
PersistentAgent kind + Governor v1 (island steward)


rjwittams:
## Parent
#604 · ADR 0009

## What to build
(Brainstorm → implementation when sequenced.) The `PersistentAgent` kind (role as label), its reconciler (free-floating-vessel residency via VesselRequirement, bounded-restart supervision), and the first **Governor** for one island (plausibly dinghy or flotilla itself): charter-driven, fast-and-loose creds (ADR 0009 v1), memory ref backed by fs or a hub git repo, wakes itself for duties like upstream watching.

Blocked conceptually by requirements/resolver v1 (ADR 0007) for the residency requirement — or can start with the seeded-policy equivalent if pulled forward by dogfooding.

## Acceptance criteria
- [ ] PersistentAgent kind + reconciler; agent survives daemon/host restart (re-materialised, memory intact)
- [ ] Governor for one island: maintains issues/docs, watches an upstream, launches a convoy in its own namespace
- [ ] Never edits source / raises PRs (charter v1; credential scoping tracked separately)

## Blocked by
Sequencing decision — after observer reshape unless dogfooding pulls it forward.

rjwittams:
Charter input from fleet operations (Robert, 2026-07-22): **crew follow-up filing should eventually be mediated through the Governor, not sprayed at the tracker.** Observed: convoy agents going wild filing follow-up issues directly. Direction:

- A convoy's findings-beyond-scope return as **a report — 'cargo' — likely an object-store artifact** (#748's briefs/artifacts store is the natural home) delivered to the steward on settlement, rather than as direct `gh issue create` calls from crew.
- The **Governor triages its island's cargo into its own backlog** — dedup against existing issues, merge related findings, decide what deserves a ticket at all. Backlog grooming is already in its charter; this makes convoy exhaust an *input* to grooming instead of a bypass around it.
- Trajectory: **a majority of convoys/crew may not touch the issue tracker directly at all** — tracker writes become a steward capability, crew get report-writing. (Same shape as the completion lesson: crew testimony in, authority mediates out.)
- pr-shepherd changes were already anticipated; some of its behavior likely moves in-convoy as this lands.

Not for v1 necessarily — but the Governor's interface should leave room for 'receives settlement cargo' from day one.

rjwittams:
[...]

PR #872, opened 2026-07-22
docs: the Governor's Charter

rjwittams:
The working profile of the Governor role (ADR 0009), distilled from the dogfood governor sessions: the seven-verb loop, inline-vs-convoy-cargo boundary, dispatch discipline (body-is-contract, placement economics), monitoring + the stuttering ladder, ADR 0017 settlement (claims/conditions/attention kept separate), recording discipline, and the human-escalation boundary.

Written so a manually-started governor agent gets the role and the command vocabulary immediately; destined to be carried verbatim as a PersistentAgent charter when #650 lands.

Ref #650.

### DIFF
--- a/docs/charters/governor.md
+++ b/docs/charters/governor.md
@@ -26,6 +26,13 @@ Everything you do is one of those seven verbs.
   If a "small fix" starts growing mid-flight, finish only if it's already done
   and tested; otherwise write the contract and dispatch.
 
+**Authority note.** The inline tier above reflects a *human-supervised*
+governor session, acting with the human's authority. An autonomous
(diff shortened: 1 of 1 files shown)


rjwittams:
Took the finding via the authority distinction rather than either proposed edit alone: the inline tier reflects a *human-supervised* governor session (which is what the dogfood practice was — the human's authority covers string-scale fixes); an autonomous PersistentAgent Governor keeps ADR 0009's read-only-git boundary and dispatches even trivial fixes, until ADR 0009 is deliberately amended. Default-to-tighter when ambiguous. On `cleat capture`: it's a real cleat verb (used extensively from governor sessions — it's cleat CLI surface, outside this workspace's provider shell-outs).

PR #874, opened 2026-07-22
fix: scope handoff brief repositories

### LINKED ISSUE #869: fix: scope handoff-brief work context to the vessel's repositories
## What to build

From PR #867 review (second round): `handoff_crew_brief` (crates/flotilla-core/src/in_process.rs) builds its work-context repository list from **all** of `convoy.spec.repositories`, while the initial-dispatch path (vessel reconciler) scopes it to the vessel's `requirement.repository_refs` subset. On a multi-repo convoy with a repo-scoped vessel, initial and handoff briefs for the same vessel disagree about which repositories are in play.

Thread the per-vessel repository scope into the handoff path: `crew_handoff_internal` already has the workflow-snapshot `task` in scope, so `requirement.repository_refs` (with the same fall-back-to-all-when-unspecified rule as the reconciler) can be passed to `handoff_crew_brief` without touching `ResolvedCrewContext`.

## Acceptance criteria

- [ ] A handoff brief for a repo-scoped vessel lists the same repositories as its initial-dispatch brief
- [ ] Unscoped vessels keep the all-repositories behavior
- [ ] Behavior test covers the scoped case through the handoff path

## Blocked by

None — can start immediately.

## Context

PR #867 (which first wired `append_convoy_work_context` into the handoff path); the scoping rule lives in the vessel reconciler around its `repository_refs` resolution.

rjwittams:
## What changed

- Threaded the workflow vessel repository scope into handoff brief generation.
- Kept the existing fallback-to-all behavior when a vessel has no explicit repository scope.
- Added handoff-path behavior coverage for scoped and unscoped multi-repo convoy briefs.

## Why

Initial dispatch already scopes repository work context to `requirement.repository_refs`, but handoff briefs rebuilt that section from every snapshotted convoy repository. That made briefs for the same repo-scoped vessel disagree after a handoff.

## Validation

Closes #869

### DIFF
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -4536,13 +4536,13 @@ impl InProcessDaemon {
             .enumerate()
             .find(|(_, process)| process.role == target && target != context.caller_role)
             .ok_or_else(|| crew_handoff_address_error(target, &context.vessel))?;
+        let CrewSource::Agent { selector, prompt } = &process.source else {
+            return Err(format!("crew target `{target}` is a tool process and cannot receive a handoff"));
+        };
         let repository_refs = task
             .repository_refs
             .clone()
             .unwrap_or_else(|| convoy.spec.repositories.iter().map(|repository| repository.repo_ref.clone()).collect());
-        let CrewSource::Agent { selector, prompt } = &process.source else {
-            return Err(format!("crew target `{target}` is a tool process and cannot receive a handoff"));
-        };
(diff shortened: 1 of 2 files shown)


rjwittams:
Applied the optional tidiness note in `1fb9550`: `repository_refs` is now resolved after the `CrewSource::Agent` check, so the tool-process error path returns before allocating the scoped repo list.

Rechecked locally:
- `cargo test -p flotilla-core --locked handoff_brief`

PR #876, opened 2026-07-22
feat: add PrincipalAttention resources

### LINKED ISSUE #860: feat: PrincipalAttention resource — regards and demands v1
## What to build

The mesh-resident, per-principal attention state from ADR 0018 (docs/adr/0018): a resource holding the principal's **regard set** and **demand queue**.

- **Regard**: target (any resource ref), source (expressed vs implicit + which policy), created/refreshed timestamps, expiry policy (decaying vs pin). Coverage (ownership subtree, downward) is computed at query time, not stored.
- **Demand**: originating work ref, kind (permission / human-gate / review), addressee (principal ref; default = dispatching principal; pool when unroutable), raised/satisfied/acknowledged lifecycle with as_of.
- Store + registry + `flotilla resource list/get/watch` support (the #851 CLI should show regards and demands with zero extra work).
- Single-principal v1: one implicit principal is fine; the shapes must not assume it (addressee and principal refs are real fields).

## Acceptance criteria

- [ ] Regards and demands persist, patch, and watch like other resources (contract tests against in-memory + sqlite per testing philosophy)
- [ ] `flotilla resource list regards` / `demands` (or equivalent kind names) works
- [ ] Demand lifecycle transitions are status patches with authority stamped, ADR 0017-style
- [ ] No emission wiring in this ticket — creation via CLI/API only (emission is the follow-up)

## Blocked by

None — can start immediately.

## Context

ADR 0018 §Principals; grill rulings 1–3 on #813.

rjwittams:
## Summary

- add mesh-resident `Regard` and `Demand` resources for principal attention state
- register the resources for dynamic list/get/watch support and add CRD projections
- add raw `flotilla resource apply` support so attention resources can be created or updated via CLI/API
- cover backend contracts, status patch transitions, registry apply, and CLI parsing

Closes #860.

## Validation

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1207,6 +1207,7 @@ dependencies = [
  "libc",
  "ratatui",
  "serde_json",
+ "serde_yml",
  "tokio",
  "tracing",
  "uuid",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,5 +51,6 @@ color-eyre = { workspace = true }
 clap = { version = "4", features = ["derive"] }
 tracing = { workspace = true }
 serde_json = { workspace = true }
+serde_yml = "0.0.12"
 libc = "0.2"
 uuid = { version = "1", features = ["v4"] }
--- a/crates/flotilla-core/src/executor.rs
+++ b/crates/flotilla-core/src/executor.rs
@@ -167,7 +167,10 @@ pub async fn build_plan(
     let issue_source = forge_issue_source(&repo.identity);
 
     match action {
-        CommandAction::QueryResourceList { .. } | CommandAction::QueryResourceGet { .. } | CommandAction::ResourceWatch { .. } => {
+        CommandAction::QueryResourceList { .. }
+        | CommandAction::QueryResourceGet { .. }
+        | CommandAction::ResourceApply { .. }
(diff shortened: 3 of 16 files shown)


rjwittams:
Fixed in `2a52022`. `apply_resource_document` now distinguishes omitted metadata fields from explicitly supplied empty values, preserves existing labels, annotations, ownerReferences, finalizers, and deletionTimestamp when those fields are omitted on update, and only replaces them when the document provides the field. I also added an apply-then-reapply regression test for a `Demand` with ownerReferences, labels, and finalizers set so a bare reapply updates spec without stripping controller metadata.

rjwittams:
Fixed in `24ab18e`. I removed `deletionTimestamp` from the apply-settable metadata surface and now preserve the existing server-owned deletion timestamp on update. The reapply regression also covers a pending-delete `Demand`: after delete sets `deletion_timestamp`, a bare `resource apply` leaves it intact while preserving the controller metadata.

rjwittams:
Evaluated. The remaining notes are non-blocking: the create/update race is acceptable for this CLI-driven apply path, and `apiVersion` validation can be tightened when the registry needs to distinguish multiple groups or versions. No further code change needed for #860.

PR #882, opened 2026-07-22
Open convoy rows in presentation manager

### LINKED ISSUE #878: feat: open-in-PM action on convoy and vessel rows
## What to build

A TUI action (action menu + keybinding) on convoy and vessel rows: **open a PM tab for this** — distinct from the existing `attach` intent (which attaches one session into the *current* terminal). Under ADR 0018 this gesture is *expressing a Regard*; v0 implements the effect directly without waiting for the regard machinery:

- On a convoy/vessel row, the action asks the connector/PM path (`pm connect` realization machinery) to materialize the tab for that convoy in the connected presentation manager, and focuses it if it already exists (View identity: focus-not-duplicate).
- Local-homed convoys only in v0. For remote-homed convoys, surface an honest 'not reachable from this PM yet' message — do not fake a tab.
- If no PM is connected, say so.

## Acceptance criteria

- [ ] Action menu on convoy and vessel rows offers 'open in PM'; a keybinding exists
- [ ] Invoking it on a local convoy makes/focuses its tab in the connected zellij (andamento)
- [ ] Remote-homed convoy and no-PM-connected cases produce clear status messages, not silence
- [ ] Behavior test through injected collaborators for the intent→connector dispatch

## Blocked by

None — uses the existing Presentation/connector machinery (v0 predates #862/#835 convergence; it will later become one emitter of Regards).

## Context

ADR 0018 (opening a latent IS expressing a regard — this is the row-level equivalent); requested from live dogfood: wave-5 feta convoys are invisible in andamento and there is currently no gesture to summon a tab.

rjwittams:
## Summary
- add an Open in PM row intent, action-menu entry, and `o` keybinding for convoys and vessels
- expose terminal materialization separately from an existing PM attachment
- focus or create local Zellij tabs using stable semantic identity, with honest remote-host and no-PM feedback

## Testing
- `TMPDIR="$PWD/.codex-tmp" cargo test --workspace --locked --features flotilla-daemon/skip-no-sandbox-tests --quiet`

Closes #878

### DIFF
--- a/crates/flotilla-daemon/src/aggregator.rs
+++ b/crates/flotilla-daemon/src/aggregator.rs
@@ -119,6 +119,7 @@ pub struct Aggregator {
     sessions_by_source: HashMap<LocalSource, HashMap<SessionKey, ResourceObject<TerminalSession>>>,
     presentation_workspaces: HashMap<PresentationKey, String>,
     terminal_sessions: HashMap<SessionKey, ResourceObject<TerminalSession>>,
+    attachable_references: HashSet<String>,
     projects: HashMap<(String, String), ResourceObject<Project>>,
     repositories: HashMap<RepositoryKey, ResourceObject<Repository>>,
     observed_checkouts: HashMap<ResourceRef, ResourceObject<Checkout>>,
@@ -149,6 +150,7 @@ impl Aggregator {
             sessions_by_source: HashMap::new(),
             presentation_workspaces: HashMap::new(),
             terminal_sessions: HashMap::new(),
+            attachable_references: HashSet::new(),
             projects: HashMap::new(),
             repositories: HashMap::new(),
             observed_checkouts: HashMap::new(),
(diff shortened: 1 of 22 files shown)


rjwittams:
Fixed in db7e61f8. Environment events now rebuild both the independent and local convoy projections after attach capabilities are recomputed, so a convoy vessel's materialization reference cannot remain stale.

Added a regression test that starts with an unresolvable convoy-bound session, flips the resolver result on an Environment event, and verifies the convoy vessel gains the expected materialization reference. The focused test, strict workspace Clippy, and the full sandbox-safe workspace suite all pass locally.

rjwittams:
Thanks for the re-review. I evaluated the remaining notes and am leaving them unchanged for this PR:

- Offering Open in PM on every convoy/vessel row is intentional. The issue requires the action on those rows and explicit feedback for remote, unknown-locality, and no-PM cases; hiding the action would remove that feedback path.
- The connector-level mutex intentionally spans discovery and creation so focus-or-create is atomic. This user action is low-frequency, and the regression test demonstrates the duplicate-prevention invariant it protects.
- Injecting the Zellij environment lookup would improve isolated detection coverage, but the current behavior is exercised at the app boundary and the note identifies no correctness defect. It can be revisited if connector detection grows more complex.

No additional code change is needed for these non-blocking observations.

PR #886, opened 2026-07-22
fix: align convoy routing with live peer routes

### LINKED ISSUE #883: fix: remote convoy ops report home host unreachable while peer data flows fresh
## What happened

First live use of #858's op routing (2026-07-22, fleet on 42ff7b12): `flotilla convoy delete awareness-band` from kiwi failed with `convoy awareness-band is homed on feta, which is unreachable` — while, simultaneously, `flotilla ls` on kiwi showed the same feta convoys with staleness `fresh (19:14:58)` and feta was up (ssh fine, daemon running). The routing layer correctly resolved the convoy's home (#858 works that far) but its reachability determination disagrees with the live peer feed.

## What to build

- Diagnose why command routing considers a peer unreachable while its snapshot/peer stream is fresh — separate connection path, stale reachability cache, address resolution difference, or a health-check with different criteria than the feed.
- Command routing should share (or at least agree with) the peer-liveness the aggregation feed demonstrates; when it genuinely cannot connect, the error should say what it tried ('connect to feta at <addr>: <cause>'), not a bare 'unreachable'.

## Acceptance criteria

- [ ] With a live peer feed from host H, `convoy delete <convoy-homed-on-H>` from another host routes and executes
- [ ] Reachability failures name the transport attempt and cause
- [ ] Behavior test at InProcessDaemon/contract level for routing-vs-feed liveness agreement

## Context

#858 (routing implementation), #850 (original op-routing issue). Workaround: run ops on the home host via ssh.

rjwittams:
## Summary

- route existing-convoy commands using the peer manager's live sender/route instead of the presentation-level connection-status cache
- preserve home-host context and report the configured transport address plus the underlying routing or send failure
- exercise routing-vs-feed agreement through an aggregation replica feed and an in-memory two-daemon request session

## Root cause

Home resolution used the aggregator projection correctly, but then rejected the command from `HostRegistry::peer_connection_status`. Actual peer delivery uses `PeerManager::resolve_sender`, so a stale presentation status could say disconnected while the live peer stream and routed sender were healthy.

## Validation

- `cargo test --locked -p flotilla-daemon --test request_session_pair`
- `RUSTC_WRAPPER= TMPDIR="$PWD/.codex-tmp" cargo test --workspace --locked --features flotilla-daemon/skip-no-sandbox-tests`

Closes #883

### DIFF
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -1211,6 +1211,12 @@ fn checkout_integration_summary(checkout: &ResourceObject<ResourceCheckout>, int
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ExistingConvoyTarget {
+    pub home: HostName,
+    pub node_id: NodeId,
+}
+
 pub struct InProcessDaemon {
     repos: RwLock<HashMap<flotilla_protocol::RepoIdentity, RepoState>>,
     repo_order: RwLock<Vec<flotilla_protocol::RepoIdentity>>,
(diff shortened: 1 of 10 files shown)


rjwittams:
Thanks — agreed that placement admission is a distinct path and needs the same liveness contract. I split that later incident scope into #887 with daemon-level acceptance criteria, so it remains tracked without broadening this convoy-operation fix after its implementation and validation pass.

On the optional address lookup point: the active configured connection label is preferred first, then an expected node ID; host-name matching is only the diagnostic fallback for a configured target that never established enough identity to populate those stronger associations.

PR #888, opened 2026-07-22
fix: reconcile repository identity changes

### LINKED ISSUE #881: fix: repo add does not reconcile project/repository identity when a remote appears
## What happened

Dogfood sequence (2026-07-22, adding andamento): `flotilla repo add ~/dev/andamento` on a repo with **no remote** derived a local identity and a project pinning that key. After adding a GitHub remote and re-running remove/add (and even a daemon bounce), the `andamento` project kept referencing the stale local-identity repository key — convoy admission failed with 'has no transport remote' despite the remote existing. Root cause: `ensure_whole_repository_project` only searches for an existing project containing the **new** key; the stale project (persisted in sqlite) squats on the natural name, so re-add silently created a disambiguated twin (`github-com-flotilla-org-andamento`) while the broken `andamento` project remained the one a user would type. Three generations of Repository resources for the same working copy now coexist (local, stale-forgejo-remote, github-remote).

## What to build

- On `repo add` (and identity-affecting refresh), when a tracked path's derived identity differs from an existing project/repository resource for the same path: **migrate** — update the project's repository entry to the new key (or replace the project spec), rather than creating a name-disambiguated twin and stranding the original.
- Garbage-collect superseded Repository resources for the same working copy (or mark them superseded) so raw listings don't accumulate identity generations.
- Surface the identity change in the command output ('repository identity changed: local → https://github.com/...').

## Acceptance criteria

- [ ] Track a remote-less repo, add a remote, re-add → the SAME project name now references the remote-keyed repository and convoy admission passes
- [ ] No orphaned twin project; superseded repository resources cleaned or marked
- [ ] Behavior test through injected collaborators covering the local→remote identity transition

## Context

Found making andamento dispatchable (#870 fixture 1). Related friction, deliberately separate: admission rejects issues whose tracker isn't part of the project — the cross-repo tracker-binding question belongs to #870's grill.

rjwittams:
## Summary

- reconcile a tracked repository when its identity changes from local-only to a canonical remote
- migrate the existing whole-repository Project without creating a twin, while preserving identities still tracked elsewhere
- remove superseded Repository resources when safe and annotate retained resources that still have durable checkouts
- surface identity changes from both repo add and refresh commands
- cover the behavior through injected repository inspection and in-memory resource backends

## Validation

Closes #881

### DIFF
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -1008,6 +1008,8 @@ struct FleetReplicaCacheEntry {
     last_error: Option<String>,
 }
 
+const SUPERSEDED_BY_ANNOTATION: &str = "flotilla.work/superseded-by";
+
 #[derive(bon::Builder)]
 struct ResolvedCrewContext {
     namespace: String,
@@ -2945,8 +2947,11 @@ fn local_repository_matches_checkout(spec: &RepositorySpec, checkout: &crate::re
     }
 }
 
-struct RepositoryProjectReconciliation {
-    identity_change: Option<RepositoryIdentityChange>,
+#[derive(Debug)]
+pub struct AddRepoOutcome {
+    pub tracked_path: PathBuf,
+    pub resolved_from: Option<PathBuf>,
+    pub identity_change: Option<RepositoryIdentityChange>,
 }
 
 fn normalize_workspace_slug(candidate: &str) -> String {
@@ -3677,7 +3682,7 @@ impl InProcessDaemon {
     async fn reconcile_whole_repository_project(
         &self,
         inspection: &crate::repository_inspection::RepositoryInspection,
-    ) -> Result<RepositoryProjectReconciliation, String> {
+    ) -> Result<Option<RepositoryIdentityChange>, String> {
         let namespace = self.provisioning_namespace().await;
         let repository_spec = &inspection.spec;
         let repository_key = repository_spec.key();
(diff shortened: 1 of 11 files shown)


rjwittams:
Fixed in 0d243c98. Reconciliation now recognizes any Project with a whole-repository entry for the migrated key, and prefers a custom-named Project as the survivor when cleaning generated duplicates. Added `tracking_after_custom_project_identity_change_does_not_create_generated_twin`, which creates a custom Project under the local identity, changes the injected inspection to the remote identity, tracks the path, and asserts that only the custom Project remains with its display name preserved and repository key migrated.

rjwittams:
Fixed in 3c84b230. Reconciliation now records which Projects were actually migrated and prefers one of those as the survivor, so name ordering cannot select a pre-existing remote twin. Generated-twin matching permits the generated display names from either identity while still requiring the canonical workflow, issue source, and repository definition, preserving user-evolved Projects. Added `identity_change_preserves_migrated_project_when_local_and_remote_names_differ`, with a lexically later local directory and a pre-existing remote Project, and asserted that the migrated Project identity and display name survive as the sole Project.

rjwittams:
Confirmed the low-confidence note is possible if the earlier Repository snapshot races with deletion or begins inconsistent. It is outside #881's normal migration path and non-blocking for this PR, so I filed #893 to make the annotation path tolerate an already-missing Repository and cover it with a regression test.

PR #889, opened 2026-07-22
feat: compute central awareness salience

### LINKED ISSUE #864: feat: central salience computation
## What to build

The centrally-computed salience hint from ADR 0018: every projection node/entry carries `none/info/attention/urgent`, computed mesh-side by the demand-join (e.g. NeedsInput ∧ unacknowledged ∧ addressee-relevant ⇒ urgent) so surfaces never reimplement routing judgment. Surfaces own rendering only.

## Acceptance criteria

- [ ] Salience appears on projection nodes with as_of; recomputed on demand/attention/regard changes
- [ ] The join precedence (demand state × regard coverage × ADR 0017 attention observations) is table-tested
- [ ] At least one surface (TUI) renders salience distinctly; no surface computes it

## Blocked by

#860 (demands), #862 (the projection it decorates).

## Context

ADR 0018 §vocabulary; grill ruling 6 on #813.

rjwittams:
Closes #864.

## Summary

- add centrally computed `none` / `info` / `attention` / `urgent` salience and `as_of` metadata to every awareness node and entry
- join demand lifecycle, principal regard coverage, and ADR 0017 terminal attention in the mesh-side awareness projection
- model explicit Project → Convoy → Vessel coverage, exact demand/attention targets, salience-only sequence advancement, and decaying-regard expiry wakeups
- render the central salience enum distinctly in the TUI without surface-side policy computation

## Validation

- independent Standards and Spec review passes

### DIFF
--- a/crates/flotilla-core/src/aggregator_projection.rs
+++ b/crates/flotilla-core/src/aggregator_projection.rs
@@ -22,6 +22,7 @@ use uuid::Uuid;
 use crate::{
     awareness_projection::{project_awareness, AwarenessInput, ScopedIssueRow},
     query_registry::QueryRegistry,
+    salience::SalienceFacts,
     scoped_store::{ScopedCheckoutProjection, ScopedIndependentProjection},
 };
 
@@ -108,6 +109,12 @@ impl<R: QueryRow + PartialEq> QueryProjection<R> {
     }
 }
 
+#[derive(Debug, Default)]
+struct SalienceProjection {
+    facts: SalienceFacts,
+    revision: u64,
+}
+
 #[derive(Debug, Default, Clone, bon::Builder)]
 pub struct AggregatorProjectionState {
     convoys: Arc<RwLock<QueryProjection<ConvoyRow>>>,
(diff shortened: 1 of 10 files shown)


rjwittams:
Fixed in 6bef94bb. The central awareness projection now emits typed per-family salience/as-of summaries (with the Convoys family intentionally including vessel descendants), and the TUI only looks up and renders the matching family summary. The project panel regression test now gives all four families distinct levels, proving an urgent convoy is not broadcast to unrelated checkout, issue, or independent panels. Pinned formatting, workspace Clippy, and the sandbox-safe full workspace tests all pass locally.

rjwittams:
Verified both non-blocking observations; no further code change is warranted here. `vessel_ref` is produced by `crates/flotilla-resources/src/convoy/reconcile.rs` when placement fields are built, so the concrete Vessel identity lookup matches its source. Pool-addressed demands intentionally remain outside the principal-specific urgent join: the current domain has no pool-membership-to-principal resolution, and treating an arbitrary principal regard as covering a pool demand would invent routing semantics. Such raised demands still surface as `Attention`; they can become principal-routed/urgent once a concrete addressee exists.

PR #894, opened 2026-07-22
fix: make convoy table row lifecycle prompt

### LINKED ISSUE #890: fix: convoy table is slow to reflect row creation and removal
## What happened

Live TUI observation (2026-07-22, fleet on 42ff7b12+): the convoy table takes long enough to show new convoys / drop deleted ones that it reads as 'not updating' (a restart was used to see new rows; a deleted convoy's row lingered well past the delete). Field *updates* on existing rows propagate fine.

## What we know (governor pre-investigation)

- The resource layer is watch-driven, not polled: the aggregator runs `list_and_watch_convoys` and applies convoy events from the durable stream as they arrive.
- `rebuild_local_projection` runs on those events and computes changed/removed row deltas against the shared view.
- Therefore the latency lives downstream of the rebuild — candidates: how/when deltas are published to subscribed clients, the TUI's application/render path for row inserts/removals vs field updates (updates being fast suggests an asymmetry here or in delta vocabulary), or the convoy table being served by a demand-backed query with its own refresh cadence rather than the delta stream.

## What to build

- Instrument or trace one create and one delete end-to-end (resource event → projection rebuild → client delivery → TUI state) and find the slow hop.
- Fix it: row lifecycle should propagate with the same promptness as field updates — seconds, not restart-scale.
- Behavior test at whichever layer the gap is (delta emission or TUI application) covering insert and removal.

## Acceptance criteria

- [ ] `convoy start`/`delete` visible in a connected TUI within a few seconds, no restart
- [ ] The found root cause is written up on this issue (it informs #862-family query design)
- [ ] Regression test for row insert + removal propagation

rjwittams:
Closes #890

## What changed

- publish convoy row insert/remove projection changes before resolving change-request metadata
- resolve change-request enrichment asynchronously, cancelling stale lookups and ignoring stale completions
- treat convoys pending finalization as removed from the table, so slow worktree teardown does not keep a deleted row visible
- add a regression covering insert and pending-finalization removal while enrichment never completes

## Root cause

The aggregator's single watch loop awaited live change-request resolution before rebuilding the convoy projection. A slow provider lookup withheld the initial insert delta and queued later events behind it. Deletion had a second lifecycle-specific delay: resource deletion first marks the convoy for finalization, and the actual `Deleted` event arrives only after presentation, vessel, checkout, and worktree cleanup. Large target directories can make that teardown take about a minute.

The client relay, client sequencing, and PM connector apply/rebuild paths are immediate once the daemon emits a row delta; the waits were upstream in aggregation and finalization.

## Validation

- `cargo test -p flotilla-daemon --locked --lib aggregator::tests:: --no-default-features` (28 passed)
- `cargo test --workspace --locked` reached an unrelated existing macOS path assertion in `server::tests::dispatch_add_list_remove_repo_round_trip` (`/private/var/...` vs `/var/...`); the changed-area tests pass


rjwittams:
Addressed both review points in 3953d532.

- Refresh generations now use unique UUID tokens and are pruned when a convoy leaves the projected set. The unique token prevents an old queued completion from matching a later convoy with the same resource key; the new regression exercises that stale-result ABA case.
- Pending-finalization filtering is now scoped to `projected_convoy`/`effective_convoys`. `effective_convoy` retains the raw last-known resource for salience joins, with a regression proving a terminating completed convoy is absent from table rows while its settled work remains visible to salience.

All 34 aggregator tests, formatting, and workspace clippy pass locally.

rjwittams:
Fixed both findings in 3953d532 and added regressions for generation pruning/stale-token rejection and the table-versus-salience projection split. All CI checks pass on the updated commit.

rjwittams:
Confirmed. Independent rebuilds are the intentional correctness-first trade-off that keeps provider I/O off the watch loop. The observation is non-blocking and no current convoy-count measurement shows a problem, so I am not adding a speculative debounce or issue; coalescing the completion queue remains a measurable optimization if this becomes hot.

PR #899, opened 2026-07-22
feat: shimmer pending table rows

### LINKED ISSUE #891: feat: row shimmer for long-running actions in convoy and project tables
## What to build

Port the Plane-A checkouts-table affordance: while a long-running action is in flight against a row (convoy delete/abandon, checkout ops), the row renders a **shimmer** (animated pending treatment) until the action resolves and the row updates or disappears. Plane A's repo/checkouts table had this; Plane-B tables lost it.

- Applies at least to the convoy table and project-page tables; implement as a shared row-state in the view-model/rendering layer (surface-agnostic per the factored-TUI direction), not per-widget.
- Shimmer starts when the intent's command is dispatched, ends on the row's next authoritative update or removal; failures clear the shimmer and surface the error status.

## Acceptance criteria

- [ ] Deleting a convoy shows the row shimmering until it actually disappears (pairs with the latency fix — but is correct feedback at ANY latency)
- [ ] Shared mechanism, adoptable by other tables without copy-paste
- [ ] Snapshot/behavior test coverage for the pending state

## Context

Requested from live dogfood alongside the row-latency bug (see sibling issue); the shimmer is correct UX regardless of how fast propagation gets.

rjwittams:
## What changed

- add shared query-qualified row lifecycle state to the reusable table view model
- shimmer convoy and project-table rows while commands are submitting or awaiting authoritative updates
- retain successful rows until their next authoritative update/removal, and stop animation with visible errors on failure or cancellation
- route convoy deletion and project issue starts through the shared mechanism, including command-finish/acknowledgement races
- add behavioral coverage for rendering, authoritative cleanup, project-panel routing, failures, cancellation ordering, and retries

## Why

Long-running Plane-A table actions previously had no row-local feedback, so convoy deletion could appear unresponsive until the row disappeared.

## Impact

Pending state is surface-agnostic and owned by `TableState`; all reusable table surfaces can adopt it without duplicating rendering or lifecycle logic. Existing repo-page pending actions retain their current behavior.

## Validation

Closes #891


rjwittams:
Thanks for calling this out. `ResultSet` is intentionally treated as authoritative for every row in its query: it is the next authoritative update when delta history is unavailable, and retaining pending state across it could leave a removed row shimmering indefinitely after gap recovery. The row-level marker may therefore yield to the full snapshot; a later command failure still surfaces through the app-wide status message.

I added `full_authoritative_snapshot_clears_pending_rows_for_its_query` in 78d6265e to pin both parts of that contract: the matching query is cleared, while pending state for another query is retained.

rjwittams:
Agreed—the current guards make the transition ordering reliable, but removing the row before validating its state was an unnecessary footgun. Commit 8f491584 now clones only a `Submitting` state before replacing it, so out-of-order calls are true no-ops. `out_of_order_mark_pending_preserves_existing_row_state` covers both existing `Pending` and `Failed` states.

rjwittams:
Thanks for the final verification. Confirmed: there are no remaining actionable findings; the two unchanged observations are non-blocking and outside this issue's required behavior.

PR #901, opened 2026-07-22
fix: tear down nested repositories completely

### LINKED ISSUE #884: fix: teardown leaves nested repositories in convoy worktrees; gate misses them as dirt
## What happened

First andamento convoy (andamento-total-fallback, 2026-07-22): the crew cloned ~/dev/zellij *inside its worktree* (reasonable — conformance testing against real zellij). Gated delete reported ok and removed the worktree's own content, but left the nested clone behind (`…/andamento-total-fallback/zellij` survived, on a checked-out feature branch). Two gaps:

1. **The Clean gate apparently didn't treat the nested repo as dirt** — `?? zellij/` should have shown in porcelain and produced a refusal naming it (unless something filtered it). If the crew gitignored it, the gate is technically consistent but the teardown contract still silently strands a repo containing potentially-unique commits.
2. **Teardown's removal skips/fails on embedded repositories** rather than either removing them (when the gate verified them safe) or refusing loudly.

The stranded clone contained commits verified (by hand) to exist in the source repo before manual removal — but that verification is exactly what the machinery should do or demand, not the operator.

## What to build

- `inspect_clean` (or its successor) must surface nested repositories explicitly ('embedded repository zellij/ (branch feat/…, N local commits)') — as a refusal by default.
- Teardown either removes what the gate verified, completely, or refuses — no partial removal that strands an embedded repo.
- Decide + record whether crews cloning scratch repos into worktrees should instead be steered (brief guidance or a scratch path outside the checkout) — cheap prevention.

## Acceptance criteria

- [ ] A worktree containing an embedded repo with local-only commits refuses gated deletion, naming the repo and its dirt
- [ ] After a passing gate, teardown leaves nothing at the worktree path
- [ ] Behavior test with a nested-repo fixture

## Context

ADR 0017 teardown contract ('refusals name the dirt per checkout'); #839/#856 implemented the gates + session retirement; found on the first non-core-project convoy.

rjwittams:
Closes #884

- surface embedded repositories in clean-gate inspection with branch, local commit, and dirt details
- ensure teardown removes the complete managed worktree path after a passing gate
- add behavior coverage for ignored nested repositories with local-only commits and successful teardown removal

Validation: cargo +nightly-2026-03-12 fmt --check; cargo clippy --workspace --all-targets --locked -- -D warnings; sandbox-safe workspace cargo test


rjwittams:
Thanks for the review. I verified the two points as non-blocking: the local-commit count is intentionally conservative across all local refs, and cleanup verification is covered by the end-to-end teardown assertions. The PR keeps the current concise refusal wording and does not add a separate mock-only branch test.

PR #904, opened 2026-07-23
fix: recover peer session establishment

### LINKED ISSUE #902: fix: peer sessions never establish after cd876ab0 — placement blocked with 'no peer summary'
## What happened

After the fleet bump to cd876ab0 (2026-07-23 ~01:00Z, all three hosts same sha — no version skew), kiwi→feta placement fails with `no peer summary for host environment d783f8a7…` and does not recover. Peer sync should take milliseconds; it never completes at all.

## Evidence (all collected live)

- kiwi daemon has both ssh tunnels up (`ssh -N -L …peers/feta.sock:…flotilla.sock`, fresh socket files); feta's daemon + socket confirmed alive.
- `flotilla topology` on kiwi: both peers listed, Direct=yes, **Connected=no** — and it stays no for minutes (no successful retry).
- `flotilla host list` on kiwi: **only the local host row**; feta/udder absent despite `~/.config/flotilla/hosts.toml` configuring both. Local row shows Configured=no, Repos=0.
- **Asymmetry**: feta's `topology`/`host list` show kiwi **connected with a summary** — feta accepted an inbound session (possibly from an earlier, now-dead kiwi daemon; kiwi daemons were bounced repeatedly incl. an accidental double-daemon).
- Ruled out: version skew (same sha), tunnel transport (up), feta readiness (socket+daemon fresh), kiwi daemon restart with feta already up (bounced twice — still Connected=no seconds-to-minutes later).

## Suspects

#886 (align convoy routing with live peer routes) and #887 (align placement admission with live peer liveness) landed in this bump and changed exactly this plumbing — 'no peer summary' is their vocabulary. Candidate mechanisms: peer session handshake failing silently with no retry; hosts.toml no longer loaded into configured hosts; summary subscription happening only on a path the new code no longer triggers.

## What to build

- Find and fix the establishment failure; peer summary present within ~1s of both daemons being up.
- **Retry with backoff** on peer-session failure — a failed dial must never be terminal.
- `topology` should show last-attempt/last-error per peer so this class diagnoses itself.
- Reproduce at InProcessDaemon/two-local-daemons level; regression test for establish + re-establish after either side bounces.

## Workaround

Dispatch on the placement host itself via ssh (local placement needs no peer summary).

rjwittams:
Closes #902

Peer dial attempts now retain diagnostic state in topology, failed subscriptions are fully disconnected so the reconnect runtime owns and retries them, and the CLI surfaces the last dial attempt/error.

The root cause was a successful transport connect followed by subscription failure leaving an activated session without an inbound forwarder or retry owner.

### DIFF
--- a/crates/flotilla-core/src/host_registry.rs
+++ b/crates/flotilla-core/src/host_registry.rs
@@ -760,7 +760,15 @@ fn build_topology(local_node: &NodeInfo, routes: &[TopologyRoute], configured_pe
         }
         if !routes.iter().any(|r| r.target.node_id == *node_id) {
             let node = NodeInfo::new(node_id.clone(), display_name.clone());
-            all_routes.push(TopologyRoute { target: node.clone(), next_hop: node, direct: true, connected: false, fallbacks: vec![] });
+            all_routes.push(TopologyRoute {
+                target: node.clone(),
+                next_hop: node,
+                direct: true,
+                connected: false,
+                fallbacks: vec![],
+                last_attempt: None,
+                last_error: None,
+            });
         }
     }
 
--- a/crates/flotilla-core/src/in_process/tests.rs
+++ b/crates/flotilla-core/src/in_process/tests.rs
@@ -1931,6 +1931,8 @@ async fn attach_query_uses_topology_next_hop_for_multi_hop_route_shape() {
             direct: false,
             connected: true,
             fallbacks: vec![],
+            last_attempt: None,
+            last_error: None,
         }])
         .await;
 
@@ -2174,6 +2176,8 @@ async fn attach_query_reports_route_that_points_back_to_local_host() {
             direct: false,
             connected: true,
             fallbacks: vec![],
+            last_attempt: None,
+            last_error: None,
         }])
         .await;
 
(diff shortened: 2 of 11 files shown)


rjwittams:
Fixed in 4ca47f97. The manager now retains the last learned node identity for each configured transport independently of active connectivity, and topology also falls back to configured expected node IDs. Dial diagnostics therefore remain attributable after disconnect. The regression test now drives connect success plus subscribe failure, asserts sender teardown and disconnected diagnostics, then retries successfully and verifies the error clears.

rjwittams:
Fixed in 4ca47f97. Added an injectable one-shot subscription failure to MockTransport and replaced the shallow diagnostic test with an end-to-end PeerManager lifecycle test: connect succeeds, subscribe fails, the activated sender is torn down, the disconnected route retains its attempt/error, and reconnect succeeds and clears the error. Learned configured-target identity now survives disconnect so the diagnostic lookup remains valid.

rjwittams:
Addressed the non-blocking diagnostic items in 5edee573: the CLI test now covers populated attempt/error cells, timestamps include the UTC date and `Z` suffix, and duplicate-arbitration errors use consistent wording. The focused CLI test and exact workspace Clippy gate both pass locally.

rjwittams:
Correction: the diagnostic follow-up commit referenced above is `b2f045bf` (the review reply contained a stale prospective hash). Its description and verification results are unchanged.

rjwittams:
Acknowledged. No blocking issues remain; all four checks pass on `b2f045bf`, and the worktree is clean.

PR #906, opened 2026-07-23
feat: resolve convoy issue snapshots store-first

### LINKED ISSUE #897: feat: convoy admission resolves issue snapshots store-first
## What to build

Per ADR 0019 ruling 3 (docs/adr/0019): convoy admission currently does a raw forge fetch for the issue snapshot on every `convoy start --issue`. Change to **store-first**: if the resource store holds the issue (from the daemon's incremental polling) with as_of within a hard freshness bound (suggest minutes-scale; make it a constant, not config, until someone needs otherwise), use it; otherwise fetch. On a 403 rate-limit during fetch, fail fast with an error naming the reset time. Never silently queue the dispatch; never proceed on a stale snapshot (body-is-contract — a stale body risks a crew working a superseded contract).

## Acceptance criteria

- [ ] Dispatching an issue the daemon has freshly polled makes zero forge calls
- [ ] Stale/missing → fetch; 403 → error includes the reset time
- [ ] No dispatch proceeds with a snapshot older than the bound
- [ ] Behavior tests: warm-store dispatch, stale-store fetch, 403 path

## Blocked by

None — can start immediately.

rjwittams:
Closes #897\n\nUses fresh daemon-cached issue snapshots during convoy admission, fetches when missing/stale, and surfaces GitHub rate-limit reset times on 403 responses.

### DIFF
--- a/crates/flotilla-core/src/aggregator_projection.rs
+++ b/crates/flotilla-core/src/aggregator_projection.rs
@@ -415,6 +415,7 @@ mod tests {
                 state: IssueState::Open,
                 labels: vec![],
                 as_of: Utc::now(),
+                observed_at: None,
                 association_keys: vec![],
                 provider_name: "github".into(),
                 provider_display_name: "GitHub".into(),
--- a/crates/flotilla-core/src/awareness_projection.rs
+++ b/crates/flotilla-core/src/awareness_projection.rs
@@ -425,6 +425,7 @@ mod tests {
                 state: IssueState::Open,
                 labels: vec![],
                 as_of: Utc::now(),
+                observed_at: None,
                 association_keys: vec![],
                 provider_name: "github".into(),
                 provider_display_name: "GitHub".into(),
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -15,7 +15,7 @@ use std::{
 };
 
 use async_trait::async_trait;
-use chrono::{DateTime, Utc};
+use chrono::{DateTime, Duration as ChronoDuration, Utc};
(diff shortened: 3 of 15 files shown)


rjwittams:
Fixed in `3aaf94be`. `ReplayGhApi` now recognizes a case-insensitive `X-RateLimit-Reset` fixture header on 403 and returns the same reset-time error as the live client. Recording also preserves this failure as a 403 interaction instead of flattening it to a generic 500. A fixture-driven replay test covers the complete 403-to-human-reset-time behavior.

rjwittams:
Fixed in `3aaf94be`. `Issue.as_of` again carries GitHub's `updated_at`, preserving the UPDATED column, salience ordering, and provider ordering contract. A separate backward-compatible `Issue.observed_at` records when Flotilla fetched the exact snapshot; admission uses only that field for the five-minute freshness bound. The GitHub provider test now verifies both timestamps independently.

rjwittams:
Both consolidated findings are fixed in `3aaf94be`: issue modification time and local observation time are now separate fields, and replay fixtures reproduce the GitHub rate-limit reset error with a fixture-driven regression test. Pinned formatting, workspace clippy with warnings denied, and the full locked workspace test suite all pass locally.

rjwittams:
Thanks for the re-check. No further code change is needed for the non-blocking observations. The post-fetch freshness check intentionally enforces the admission invariant at the `IssueProvider` boundary rather than relying on GitHub's current timestamping implementation; it also rejects a future or alternate provider response with a missing/stale observation time. `Retry-After` support is outside #897's reset-timestamp contract and the reset-based polling protocol already on `main`, so it should be handled separately if that response mode is observed in practice.

Issue #913, opened 2026-07-23
Overlay replication slice 1: home-bound runtime kinds + store-level include-replicas read view


rjwittams:
## Parent

#742 (umbrella; ADR 0016 is the mechanism ruling). First slice, scoped by the openable-latent grill 2026-07-23 (record on that date's session; consumer chain: #913-follow-on aggregator slice → andamento latent tabs).

## What to build

The **generic replicator mechanism** with only the home-bound runtime kinds enabled — deliberately the fat cut (Convoy + Vessel + TerminalSession together) so a remote-homed convoy is fully fleet-visible in one landing, never a rows-without-sessions split-brain:

- **Replicator per peer** (per ADR 0016): list + watch the enabled kinds via the store's existing k8s-style HTTP API over the peer's forwarded UDS (existing ssh tunnel machinery suffices interim; Tender inherits the channel later), applying events into a **durable replica area keyed by origin root**, resuming from stored version on reconnect.
- **Kinds enabled in this slice: Convoy, Vessel, TerminalSession** — home-bound runtime class only: the replica *is* the view — read-only, staleness-stamped, never merged. Mechanism stays kind-generic with per-kind opt-in declared where the kind's class is declared; definitions/convergent-facts classes (merge blocks, MV-registers, conflict UX) are explicitly OUT of this slice and stay on #742.
- **ADR 0004 generation semantics** for generational observed kinds: replica survives the holder's restarts; an origin generation change forces discard + re-list; origin *absence* is not a generation change — rows persist as last-known, staleness-stamped (the no-regress property).
- **Store-level read view** (grill ruling): the store's list/watch API grows an **include-replicas** option; returned replica rows are origin-tagged and staleness-stamped. **Local-only remains the default**, and the include-replicas view is **read-only by construction** — reconcilers never see replicas, preserving reconciled-only-at-home by construction rather than convention.
- Deletion: origin-authored deletes propagate through the watch stream; replica rows removed accordingly.

## Acceptance criteria

- [ ] With both daemons up, feta's Convoys/Vessels/TerminalSessions are listable/watchable on kiwi via include-replicas, origin-tagged and staleness-stamped, within seconds (live watch — convergence latency is connection latency)
- [ ] Default (no include-replicas) reads return local-only rows — no consumer changes behaviour unasked
- [ ] Peer disconnect: replica rows persist, staleness-stamped; reconnect resumes from stored version without full re-list
[...]

PR #916, opened 2026-07-23
fix: report vessel placement host

### LINKED ISSUE #912: Vessel rows stamp the local host unconditionally — remote-placed vessels lie about their host
## What to fix

`summarize_vessel` (`crates/flotilla-daemon/src/aggregator.rs`) stamps `.host(self.local_host.clone())` on every VesselRow (~:1539) and `.on_host(self.local_host.clone())` on the vessel resource ref (~:1489), regardless of placement. A feta-placed vessel of a kiwi-homed convoy reports `host: kiwi`. Downstream, `KEY_VESSEL_HOST` in the PM catalog carries the same lie, and the remote-materialize work (openable-latent chain, grill 2026-07-23) needs the honest host to mint cross-host attach recipes.

The placement state already carries the truth (`WorkState.placement` fields — `vessel_ref` is read from there today). Carry the placement's actual host into both the row's `host` field and the vessel resource ref.

## Acceptance criteria

- [ ] A vessel placed on a remote host reports that host in its VesselRow and vessel resource ref
- [ ] Local placements unchanged
- [ ] Regression test covering a remote placement

## Blocked by

None — can start immediately.

rjwittams:
## Summary

- derive each vessel row's host from `WorkState.placement` when available
- apply the same placement host to the concrete Vessel resource reference
- preserve placement hosts when convoy rows pass through fleet replica ingestion
- retain the local host fallback for unplaced and legacy local rows

## Root cause

The aggregator unconditionally stamped the convoy daemon's local host onto both vessel summaries and concrete Vessel references. Replica ingestion repeated the overwrite using the convoy snapshot origin, so a remotely placed vessel could be reported as living on its convoy's home host.

## Impact

Downstream presentation metadata now receives an honest `flotilla.vessel.host`, enabling later cross-host materialization work to route to the vessel's actual host.

## Validation

Closes #912

### DIFF
--- a/crates/flotilla-daemon/src/aggregator.rs
+++ b/crates/flotilla-daemon/src/aggregator.rs
@@ -1480,16 +1480,18 @@ impl Aggregator {
 
     fn summarize_vessel(&self, convoy_ref: &ResourceRef, definition: &VesselRequirement, state: Option<&WorkState>) -> VesselRow {
         let requested_stance = definition.stance.to_string();
-        let vessel_resource = state
-            .and_then(|state| state.placement.as_ref())
-            .and_then(|placement| placement.fields.get("vessel_ref"))
+        let placement = state.and_then(|state| state.placement.as_ref());
+        let vessel_host = placement
+            .and_then(|placement| placement.fields.get("host"))
             .and_then(serde_json::Value::as_str)
-            .map(|name| {
+            .map(HostName::new)
+            .unwrap_or_else(|| self.local_host.clone());
+        let vessel_resource =
+            placement.and_then(|placement| placement.fields.get("vessel_ref")).and_then(serde_json::Value::as_str).map(|name| {
                 ResourceRef::new(api_version(Vessel::API_PATHS), Vessel::API_PATHS.kind, &convoy_ref.namespace, name)
-                    .on_host(self.local_host.clone())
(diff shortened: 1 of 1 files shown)


rjwittams:
Thanks — the grep result is correct for the current local-only `placement_status()` producer, but adding the suggested producer change there would encode the wrong identity. `PlacementPolicySpec.host_direct.host_ref` and environment `host_ref` are stable `HostId` values; `VesselRow.host` and `ResourceRef.host` require the user-facing `HostName`. The resources reconciler has no host-registry/display-name collaborator from which to resolve that mapping.

Issue #912 explicitly defines `WorkState.placement` as the source of placement-host truth and scopes this slice to consuming it in `summarize_vessel`; it is a dependency of #913/#914's replicated remote-placement path, where the origin-tagged placement data supplies that truth. The tests therefore exercise the contract boundary intentionally: one verifies a placement carrying `host: feta` projects both host-bearing outputs, and the second verifies fleet replica ingestion does not replace that truth with the kiwi convoy origin.

I’m retaining `Closes #912` and not adding a `placement_status()` derivation because deriving `HostName` from policy/environment `HostId` would make the data look populated while still lying about its type and value. The local fallback remains unchanged until a placement actually carries an origin-resolved host, as required by the issue contract.

PR #918, opened 2026-07-23
feat: replicate home-bound runtime resources

### LINKED ISSUE #913: Overlay replication slice 1: home-bound runtime kinds + store-level include-replicas read view

rjwittams:
## Summary

- add durable per-origin replica partitions and cursors to the in-memory and SQLite resource stores
- replicate Convoy, Vessel, and TerminalSession resources over the SSH-forwarded Unix socket using the k8s-style HTTP list/watch API
- add read-only `include-replicas` list/watch views with origin and per-object last-sync provenance while preserving local-only defaults
- resume watches across reconnects, preserve last-known rows while peers are absent, and atomically relist when origin generations change

## Impact

Fleet roots can now read live, durable remote home-bound runtime state without exposing replica writes to reconcilers. Existing consumers remain local-only unless they explicitly request replicas.

The resource HTTP surface deliberately shares the primary daemon Unix socket. The connection preface multiplexes read-only HTTP and the existing framed control protocol, so SSH forwarding reuses the daemon's existing peer trust boundary rather than opening a second listener.

## Validation

## Follow-up

- #923 adds generation-aware retry supervision for transient kind-local watch failures without risking duplicate replicators across reconnects.

Closes #913

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1196,6 +1196,7 @@ dependencies = [
 name = "flotilla"
 version = "0.1.0"
 dependencies = [
+ "bon",
  "clap",
  "color-eyre",
  "flotilla-commands",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,6 +49,7 @@ ratatui = "0.30"
 tokio = { workspace = true }
 color-eyre = { workspace = true }
 clap = { version = "4", features = ["derive"] }
+bon = { workspace = true }
 tracing = { workspace = true }
 serde_json = { workspace = true }
 serde_yml = "0.0.12"
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -32,7 +32,8 @@ use flotilla_protocol::{
 use flotilla_resources::{
     api_version, apply_resource_document, apply_status_patch as apply_resource_status_patch,
     apply_status_patch_checked as apply_resource_status_patch_checked, external_patches as convoy_external_patches, get_resource_kind,
-    list_resource_kind, normalize_project_spec, resolve_project_issue_sources, terminal_session_attach_target, watch_resource_kind,
+    list_resource_kind, list_resource_kind_including_replicas, normalize_project_spec, resolve_project_issue_sources,
+    terminal_session_attach_target, watch_resource_kind, watch_resource_kind_from, watch_resource_kind_including_replicas,
     Checkout as ResourceCheckout, CheckoutIntegrationStatus, CheckoutPhase as ResourceCheckoutPhase, CheckoutSpec as ResourceCheckoutSpec,
     CheckoutStatus as ResourceCheckoutStatus, ConditionValue, Convoy as ResourceConvoy, ConvoyIssue, ConvoyRepositorySpec, ConvoySpec,
     ConvoyStatusPatch, CrewSource, Environment as ResourceEnvironment, EnvironmentPhase, Host as ResourceHost,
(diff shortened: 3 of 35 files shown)


rjwittams:
Addressed the review:

- Fixed SQLite provenance to persist `last_synced_at` on each replica object rather than reading the partition cursor timestamp. Incremental updates now restamp only the touched row.
- Extended the shared in-memory/SQLite contract with two rows in one origin partition, updates only one, and asserts the untouched row keeps its original timestamp.
- Filed #923 for generation-aware per-kind retry supervision. That needs connection-generation cancellation to avoid duplicate replicators after an SSH reconnect, so it is intentionally a follow-up rather than a naive loop here.
- Kept the string-based routed fallback limited to `test-support`; production uses typed HTTP errors.
- Documented in the PR body that HTTP deliberately shares the existing daemon socket and peer trust boundary.

The exact format, clippy, and full locked workspace test commands pass after the fix.

rjwittams:
Thanks — agreed that the duplicated derivation is a real maintenance risk even though the formulas match today. A shared helper would require crossing the explicitly frozen `crates/flotilla-daemon/src/peer/**` boundary for this slice, so I kept #918 scoped and filed #924 to centralize or contract-test the path derivation, including label encoding behavior.

PR #927, opened 2026-07-23
fix: label repositories with readable slugs

### LINKED ISSUE #919: Awareness/group labels leak hashed Repository keys — label by forge slug, never the key
(A of the 2026-07-23 latent-tabs shakedown.) The PM sidebar shows groups like `repo/umfdl0jvpivapi195j3h2n74gh69gbfub13294vogopginvd1m8g` — that's the Repository resource's hashed natural-key *name*; its spec carries `forge.repository: changedirection/reticulate`. Not fake data — a display leak: the projection labels repo groups (and anywhere else Repository names surface) with the identity hash instead of a display slug. Same identity-vs-display split as #907/#645: keys stay canonical, labels come from the forge slug (or a short name on collision).

**AFK-dispatchable** once #913 lands quiets the area: audit the projection + row surfaces for raw RepositoryKey display and substitute display labels.

- [ ] No surface renders a hashed repository key as a label
- [ ] Collision case yields a qualified readable label

rjwittams:
## Summary

- keep `RepositoryKey` as canonical identity while projecting forge-derived display labels
- carry readable repository labels through checkout, independent-session, project-list, awareness, and TUI surfaces
- preserve globally qualified labels across fleet replication and disambiguate same-slug repositories across forge services
- remove opaque keys from human-facing fallback text and view-address labels

## Verification

Closes #919

### DIFF
--- a/crates/flotilla-core/src/aggregator_projection.rs
+++ b/crates/flotilla-core/src/aggregator_projection.rs
@@ -165,7 +165,7 @@ impl AggregatorProjectionState {
 
     pub async fn replace_store_catalog(
         &self,
-        repositories: HashSet<RepositoryKey>,
+        repositories: HashMap<RepositoryKey, String>,
         projects: HashMap<QueryScope, Vec<RepositoryKey>>,
     ) -> Vec<ResultDelta> {
         let mut deltas = self.independents.write().await.replace_catalog(repositories.clone(), projects.clone());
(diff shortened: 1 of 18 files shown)


rjwittams:
Verified the three non-blocking observations. No code change is needed for #919: the Convoy fallback is a readable filesystem-safe workspace slug rather than a `RepositoryKey`; the unavailable-repository condition intentionally avoids surfacing opaque identity keys in user-facing text while the keys remain in projection state for diagnostics; and `host:path` is the deliberate readable qualifier for otherwise-colliding local repositories. The issue's hash-leak and collision acceptance cases are covered directly, including fleet collisions across forge services.

PR #937, opened 2026-07-23
chore: manage target cache growth

### LINKED ISSUE #935: Target storage as managed cache state: incremental GC policy
#928/#929 report rec 5: a long-lived target/ was 32 GiB of which **19 GiB is accumulated incremental generations**; one dev+test build reaches 11 GiB. This is also the hull-pool constant: clonefile is O(inodes) (~1.4k/s measured), so pruning shrinks warm-hull forks and disk together.

Scope (keep simple):
1. Add a documented cleanup mechanism (cargo-sweep or equivalent age/size-based pruning of incremental + stale deps) runnable per-checkout; wire a convenience invocation (script or cargo alias).
2. CI: evaluate non-incremental or short-lived targets where cache upload/disk cost outweighs reuse (report notes CI already caches target/).
3. Document the policy in CLAUDE.md-adjacent dev docs: developer incrementals kept, aged generations pruned.

- [ ] Pruning tool + invocation documented and working (measured before/after on a real target)
- [ ] CI decision made and applied (or explicitly declined with reason)
- [ ] No cargo clean anywhere

rjwittams:
## Summary

- add a per-checkout `cargo-sweep`-backed policy that prunes aged incremental generations and stale dependency families, then enforces incremental and whole-target size ceilings
- provide a faithful `--dry-run` preview and injected behavioral coverage for pruning, safety, cleanup, and relative target handling
- keep developer incrementals enabled while disabling and stripping incremental state from CI caches
- document installation, thresholds, workflow, CI rationale, and a real long-lived target measurement

## Validation

- real scratch target: 68 GiB / 180,542 files before; 20 GiB / 52,150 files after (48 GiB and 71% reclaimed)
- `scripts/test-prune-target.sh`

Closes #935

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,8 +25,10 @@ jobs:
         with:
           toolchain: nightly-2026-03-12
           components: rustfmt
-      - run: scripts/test-prune-target.sh
-      - run: cargo +nightly-2026-03-12 fmt --check
+      - name: Test target pruning policy
+        run: scripts/test-prune-target.sh
+      - name: Check Rust formatting
+        run: cargo +nightly-2026-03-12 fmt --check
 
   clippy:
     name: Clippy
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,6 +10,11 @@ on:
 permissions:
   contents: read
 
+env:
+  # Cached dependency artifacts pay off across runs; short-lived incremental
+  # generations cost more to upload and store than they save in CI.
+  CARGO_INCREMENTAL: "0"
+
 jobs:
   fmt:
     name: Format
@@ -20,6 +25,7 @@ jobs:
         with:
           toolchain: nightly-2026-03-12
           components: rustfmt
+      - run: scripts/test-prune-target.sh
       - run: cargo +nightly-2026-03-12 fmt --check
 
   clippy:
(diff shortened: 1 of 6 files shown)


rjwittams:
Addressed the actionable review notes in `9d4be046`:

- named the pruning-policy and Rust-format steps so the lightweight behavior test is not mistaken for formatting;
- clarified that the default target is checkout-local, while an explicitly shared `CARGO_TARGET_DIR` is intentionally honored and should be unset or redirected before pruning;
- added rejection tests for invalid age, incremental-cap, and full-target-cap settings.

I kept the four pre-cache cleanup steps local to their jobs because each must run immediately before its corresponding cache post-action, including on failure. I did not add a standalone lock: it would not protect an active Cargo build unless Cargo participated in the same lock, so the existing explicit “no build or test running” precondition remains the accurate safety contract.

rjwittams:
Addressed in `9d4be046`; the full review-cycle summary is in the follow-up comment above.

rjwittams:
Documented in `e28c8357`: relative `CARGO_TARGET_DIR` values are anchored to the checkout root, and developers who normally invoke Cargo elsewhere should use an absolute value.

rjwittams:
Evaluated: this final review confirms all prior items are resolved and raises no new action, so no further code change is needed.

Issue #928, opened 2026-07-23
Convoy cycle time: measured decomposition and remediation threads


rjwittams:
Robert's ask 2026-07-23: convoys take an hour+; analyse and remediate. Measured (same-day PRs, dispatch times from session log):

| Convoy | →1st commit | commit→PR | PR→merge | total | dominated by |
|---|---|---|---|---|---|
| #893 | ~10m | <1m | 15m | ~25m | nothing (clean baseline) |
| andamento#6 | ~13m | 4m | 8m | ~25m | nothing |
| #912 | ~12m | 4m | 68m | ~85m | review rounds + CI + shepherd sleeps |
| #913 | ~20m | 71m | 40m | ~2h10 | legitimate size + 5 review rounds |
| andamento#8 | — | — | 30m | ~2h10 | ~90m stalled awaiting steer |
| #919 retry | ~55m | 1m | 18m | ~75m | xhigh test iteration + wasted first dispatch |

Clean baseline is ~25 min; the hours come from waits between actions, not agent speed. Threads (Robert's numbering, enriched with measured facts):

1. **Build reuse within vessel + CI**: fresh worktree per convoy pays full link + disk (#914's worktree target/ = 31G; main = 49G) despite sccache. CI caches target/ but Test is still 5m/round (restore+incremental+link+run). Levers: warm hulls, shared-target strategies, link time, CI job profile.
2. **Stall latency**: 90m idle on andamento#8; done-in-draft PRs sit until a human notices. Levers: Bosun stutter-ping (#650, `Idle ∧ unsettled` already detectable); crews never open drafts / `gh pr ready` on green themselves (skill-line change, ready to enact).
3. **Effort/model schedule per crew — dynamic workflow**: xhigh everywhere costs 10-20m orientation on trivial tickets. Other half: **splitting phases to different crews** (different harness/model family, fresh-or-richer context, different authority per phase). The capability→(adapter,model) table is the plumbing; scheduling is the design.
4. **sccache**: NOT off — active on kiwi, 81.7% Rust hit rate; what's parked is in-image enforcement (#812: advisory infra gets bypassed). Research: shared backend (S3/Redis/WebDAV on lab infra) to share hits across hosts + fresh worktrees; sccache-dist macOS support (comte as second M4 builder).
5. **Testing strategy**: too many nitty-gritty repetitive tests vs behavior tests — costs code, compile time, and run time every round.
[...]

PR #938, opened 2026-07-23
Move default Rustls provider to Ring

### LINKED ISSUE #934: Get the aws-lc-sys native build off the clean critical path
#928/#929 report rec 1: `aws-lc-sys` consumed **189.11s** (182.91s C/CMake build script) = **43.7% of the 418s clean build**, and it gates the chain into resources→core→daemon. It arrives as the rustls provider via reqwest.

Investigate in order:
1. Can reqwest/rustls use the **ring provider** (or platform-native TLS) explicitly, without changing security or platform requirements? If yes, switch — largest plausible clean-build wall win.
2. If not: make sccache/CI caching retain the native build output reliably (sccache does not cache build-script *runs* — document what does: cache the target build dir portion, or prebuilt aws-lc).

Report findings in the PR description with before/after `--timings` numbers for the clean build.

- [ ] Clean-build wall measured before/after
- [ ] Security posture unchanged and stated (what provider, why acceptable)
- [ ] Full CI green

rjwittams:
## Summary

- switch Reqwest to `rustls-no-provider` and explicitly install Rustls's Ring provider for normal Flotilla builds
- centralize provider-aware Reqwest construction so binaries, resource HTTP clients, request factories, and tests cannot race provider selection
- preserve Reqwest's prior `charset`, HTTP/2, JSON, system-proxy, query, and stream behavior
- add `--features aws-lc-provider` as an opt-in production path when deployment-specific profiling shows a runtime benefit or the broader AWS-LC algorithm set is required

## Clean-build timings

Measured on the same Apple M4 host (10 logical CPUs), with Cargo/rustc 1.94.1, an empty isolated target directory, `RUSTC_WRAPPER=` to bypass sccache, and:

```console
cargo build --workspace --locked --timings
```

The baseline is `f4333f23`, as recorded by the build profile behind #928/#929. The accepted Ring sample was taken after competing Cargo processes had cleared.

| Metric | AWS-LC baseline | Ring default | Change |
| --- | ---: | ---: | ---: |
| `/usr/bin/time` wall | 418.57s | 349.67s | **-68.90s (-16.5%)** |
| Cargo total | 418.3s | 349.5s | -68.8s (-16.4%) |
| Provider package units | `aws-lc-sys` 189.11s | `ring` 21.75s | -167.36s (-88.5%) |
| Native build-script run | 182.91s | 6.69s | -176.22s (-96.3%) |
| Average active units | 5.68 | 6.90 | +1.22 |
| All 10 slots active | 46.4% | 60.1% | +13.7pp |

Before, `aws-lc-sys`'s build-script run occupied 26.88–209.79s and gated Rustls/Reqwest into resources → core → daemon. With Ring, its build-script run is 18.21–24.90s and the Ring library finishes at 37.35s; the critical gate moves back to ordinary Rust crates. The wall reduction is smaller than the provider-unit reduction because the old native build overlapped other dependency work.

## TLS/security posture

This keeps the Rustls protocol stack, platform certificate verification, TLS 1.2/1.3 support, and Flotilla's standard default cipher-suite policy. Ring is one of Rustls's built-in crypto providers, and Flotilla installs and verifies that exact built-in provider at process startup and at every owned Reqwest construction seam.

This is equivalent for Flotilla's documented security requirements, but the provider capabilities are not byte-for-byte identical:

- the previous AWS-LC configuration was not FIPS mode, and Flotilla has no documented FIPS requirement
- AWS-LC's default provider additionally exposes the hybrid `X25519MLKEM768` group and some less-common P-521/signature-verification combinations; Ring does not
- Flotilla does not require those capabilities; ordinary X25519/P-256/P-384 key exchange and RSA/ECDSA/Ed25519 verification remain available

Platform-native TLS was rejected because it would introduce platform-divergent OpenSSL/Security Framework/Schannel behavior and change the current Rustls/platform-verifier and kubeconfig PEM assumptions.

For production deployments that demonstrate an AWS-LC runtime advantage or need its broader algorithm catalogue, `cargo build --release --features aws-lc-provider` restores AWS-LC as the explicitly selected runtime provider. Cargo features are additive, so this ergonomic opt-in compiles Ring too, but provider selection remains deterministic; `aws-lc-sys` is intentionally absent from the default clean-build graph.

## Validation

- default `cargo tree --workspace --locked -i aws-lc-sys -e features` finds no active `aws-lc-sys` package
- `cargo tree --workspace --locked --features aws-lc-provider -i aws-lc-sys -e features` confirms the opt-in graph
- provider-selection tests pass with both Ring and `aws-lc-provider`

Closes #934

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1184,6 +1184,7 @@ dependencies = [
  "reqwest",
  "rstest",
  "rusqlite",
+ "rustls",
  "serde",
  "serde_json",
  "serde_yml",
@@ -1411,10 +1412,8 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ff2abc00be7fca6ebc474524697ae276ad847ad0a6b3faa4bcb027e9a4614ad0"
 dependencies = [
  "cfg-if",
- "js-sys",
  "libc",
  "wasi",
- "wasm-bindgen",
 ]
 
 [[package]]
@@ -1424,11 +1423,9 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "899def5c37c4fd7b2664648c28120ecec138e4d395b459e5ca34f9cce2dd77fd"
 dependencies = [
  "cfg-if",
- "js-sys",
  "libc",
  "r-efi 5.3.0",
  "wasip2",
- "wasm-bindgen",
 ]
 
 [[package]]
@@ -2073,12 +2070,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "lru-slab"
-version = "0.1.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "112b39cec0b298b6c1999fee3e31427f74f676e4cb9879ed1a121b43661a4154"
-
-[[package]]
 name = "mac_address"
 version = "1.1.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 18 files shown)


rjwittams:
Addressed the review findings in `fbf232a2`:

- `HttpBackend::from_unix_socket` now routes through `crate::tls::client_builder()`. Both affected focused tests pass, as does the complete locked workspace suite.
- The AWS-LC opt-in intentionally uses Cargo's additive feature model: it compiles Ring too but deterministically selects AWS-LC at runtime. The PR security-posture section now states this explicitly. The default graph remains Ring-only and contains no `aws-lc-sys`.

PR #947, opened 2026-07-23
fix: skip splash for scoped views

### LINKED ISSUE #941: Scoped (non-full) TUI views skip the splash screen entirely
Robert, 2026-07-23. Focused `flotilla view <address>` scoped-mode TUIs — the panes that per-level PM layouts and convoy-open materialization spawn — should never show the splash screen: they open into a specific view; the splash is a full-app greeting.

Bonus, measured same day: the splash's kitty image placements are exactly the trigger surface of the screen-thread saturation pathology (fork-issues/zellij#10 — O(scrollback) anchor resolution on erase-to-EOL for panes holding placements). Scoped TUI panes are the ones that multiply per-tab in the PM, so skipping splash there removes N placement-holding panes from the blast radius as a side effect.

- [ ] `flotilla view <address>` renders the target view immediately, no splash
- [ ] Full `flotilla` (unscoped) keeps the splash unchanged

rjwittams:
## Summary

- skip the splash operation when the TUI starts with a scoped view address
- preserve the existing splash path for full, unscoped startup
- regression-test splash invocation for both startup modes

Closes #941

## Testing

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,4 +1,4 @@
-use std::{path::PathBuf, process::ExitStatus, sync::Arc};
+use std::{future::Future, path::PathBuf, process::ExitStatus, sync::Arc};
 
 use clap::Parser;
 use color_eyre::Result;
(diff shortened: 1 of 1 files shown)


rjwittams:
Keeping the typed `Option<&ViewAddress>` here: it passes the startup scope state already held by `run_tui` directly into the behavior seam and avoids introducing a separate boolean that could drift from that state. This also follows the repository's preference for typed state over booleans. The suggestion is non-blocking and does not warrant a code change before merge.

PR #951, opened 2026-07-23
fix: use readable convoy checkout directories

### LINKED ISSUE #907: Convoy checkout directories get unreadable names (repo-key hash / verbose remote slug)
## What

Convoy checkout directories under the checkout root are named from the repository key, which produces unreadable paths in practice:

- The andamento convoy's checkout landed under a verbose remote-derived slug (`github-com-flotilla-org-andamento...`-style), flagged during dogfood 2026-07-22.
- Worse specimen found 2026-07-23 during wave-7 settlement: an orphaned andamento checkout directory named `18bgq9qrinlfemovt24qr753ipaqh21avh9rgl4vr1hkm0ianq7g` — a bare hash, presumably a repo-key fallback when the remote could not be normalized to a readable slug.

Humans navigate these paths constantly (attach lands there, editors open there, `ls ~/dev/flotilla-repos` is a working surface). The convoy-name level of the hierarchy is readable; the repo level inside it should be too.

## Direction

- Directory name should prefer a short readable repo slug (e.g. `cleat`, `andamento`, `flotilla`), qualified only on collision.
- The repo *key* (identity) and the *directory name* (presentation) need not be the same string — identity can stay canonical/hashed while the path stays human. Relates to #645 (machine-independent repo identity) — that ticket is about identity semantics; this one is only about the on-disk name.
- Hash-fallback should never surface as a path name; if normalization fails, fall back to something derived from the URL basename.

## Acceptance criteria

- [ ] Fresh convoy on cleat/flotilla/andamento produces checkout dirs whose repo segment is a short readable slug
- [ ] Collision between two repos with the same basename yields a qualified-but-readable name
- [ ] No code path can produce a bare-hash directory name

rjwittams:
## What changed

- centralize short, path-safe repository workspace slug allocation
- use collision-qualified readable slugs for shared clone directories
- use the convoy's snapshotted workspace slug for single-repository checkout paths
- cover cleat, flotilla, andamento, basename collisions, clone paths, and worktree paths with regressions

## Why

Convoy paths exposed opaque repository-key hashes for shared clones and verbose remote-derived slugs for single-repository worktrees. Repository identity remains canonical and opaque, while the on-disk presentation is now human-readable.

## Validation

Closes #907

### DIFF
--- a/crates/flotilla-controllers/src/reconcilers/vessel.rs
+++ b/crates/flotilla-controllers/src/reconcilers/vessel.rs
@@ -7,12 +7,12 @@ use flotilla_resources::{
     controller::{
         delete_lifecycle_owned_matching, Actuation, LabelJoinWatch, LabelMappedWatch, ReconcileOutcome, Reconciler, SecondaryWatch,
     },
-    Checkout, CheckoutPhase, CheckoutSpec, CheckoutWorktreeSpec, Clone, ClonePhase, CloneSpec, Convoy, CrewSource, DockerCheckoutStrategy,
-    DockerEnvironmentSpec, Environment, EnvironmentMount, EnvironmentMountMode, EnvironmentPhase, EnvironmentSpec, FreshCloneCheckoutSpec,
-    HostDirectPlacementPolicyCheckout, HostDirectPlacementPolicySpec, InputMeta, LifecycleAuthority, OwnerReference, PlacementPolicy,
-    PlacementPolicySpec, Repository, RepositoryIdentity, RepositoryKey, Resource, ResourceBackend, ResourceError, ResourceObject, Stance,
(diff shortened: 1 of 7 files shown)


rjwittams:
Addressed both points in f014d1e6. The repository catalog is now listed and slug allocation is performed only in the missing-clone branch. That allocation explicitly includes the already-fetched current repository, so a point-in-time catalog omission cannot make the current repository lookup panic. Existing clones no longer incur a catalog list or slug computation on reconcile.

Validation after the change: pinned formatting, workspace clippy with warnings denied, the controller regression suite, and the full workspace test suite all pass.

rjwittams:
Confirmed. The remaining repeated list is confined to the one-time missing-clone path and bounded by the vessel's repository scope, so I am leaving it local to the creation branch rather than adding caching or a broader prefetch seam to this fix. No further code change is needed for this non-blocking observation.

PR #956, opened 2026-07-24
feat: split project and repository fact segments

### LINKED ISSUE #948: Fact dialect v1: split the project/repo segment overload, add source fact, align the watcher
## Parent

#943 (fact semantics) — enactment slice 1(i) from the 2026-07-23 grill (record on #921). Deliberately scoped WITHIN today's path-shipping connector: dialect coherence now, the flat-facts restructure is a separate gated contract.

## What to build

- Kill the `project_ref.or(repo)` overload in `project_segment` (flotilla-manifest): the project segment key becomes `flotilla.project` with value = Project resource name, ONLY minted when a project_ref exists. A **separate repo-level segment** (`vcs.repo`) carries value = canonical forge slug (host:path fallback when slugless), minted from Repository knowledge. Entities with no project are truthfully unparented at project level (repo segment only) — never squeezed into the project key.
- Every patch grows a **producer/source fact** (e.g. `source=flotilla`) so surfaces can badge provenance.
- **Align the git-watcher's stamp** (andamento repo, python watcher): same repo fact key + forge-slug dialect, `source=git-watcher`. One small change; the watcher keeps no other role in this slice.
- Group paths keep being shipped as today (spine project → repo → convoy → vessel becomes what the connector emits) — the template-derived-hierarchy step is NOT this ticket.

## Acceptance criteria

- [ ] No segment/fact key carries either-of-two-meanings values; a semantics note per emitted key lands in the wire module docs
- [ ] A watcher-born pane and a flotilla convoy for the same repo join at the repo level (live check in andamento)
- [ ] Latent and live presences of one dispatched convoy land in the SAME group (the dual-presence specimen dies)
- [ ] Source fact present on all patches; existing tests updated

ADR carry: ride the PR with the ADR 0018 refinements-block amendment for rulings 1-2 (facts one-dialect; source facts).

rjwittams:
## Summary

- split Project resource grouping (`flotilla.project`) from canonical repository grouping (`vcs.repo`)
- preserve canonical repo facts separately from mutable display labels across convoy, checkout, independent, and awareness projections
- give independent sessions their own single-meaning grouping vocabulary
- add `source=flotilla` provenance to every catalog, pane, and tab patch, including retractions
- make latent catalog paths and live workspace stamps share the canonical project → repo → convoy → vessel spine
- document the v1 fact semantics and amend ADR 0018

Companion watcher alignment: flotilla-org/andamento#19

## Validation

Closes #948

### DIFF
--- a/crates/flotilla-controllers/src/reconcilers/presentation.rs
+++ b/crates/flotilla-controllers/src/reconcilers/presentation.rs
@@ -27,7 +27,7 @@ use flotilla_protocol::{arg, EnvironmentId};
 use flotilla_resources::{
     controller::{LabelJoinWatch, ReconcileOutcome, Reconciler, SecondaryWatch},
     Convoy, Environment, Host, Presentation, PresentationStatus, PresentationStatusPatch, Repository, ResourceBackend, ResourceError,
-    ResourceObject, TerminalSession, TerminalSessionPhase, TypedResolver, CONVOY_LABEL,
+    ResourceObject, TerminalSession, TerminalSessionPhase, TypedResolver, CONVOY_LABEL, VESSEL_LABEL,
 };
 use sha2::{Digest, Sha256};
 use tracing::warn;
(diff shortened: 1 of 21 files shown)


rjwittams:
Addressed both correctness findings in f280f6c4:

- awareness independents now emit `flotilla.independent`, matching the direct projection; checkout awareness entries also received their own `flotilla.checkout` segment so `flotilla.vessel` remains vessel-only
- `repo_segment` now derives a useful local-repository label from `host:path/.git` (`flotilla`, not `.git`)

Both paths have manifest regression coverage.

rjwittams:
Fixed the repo-only awareness parent mismatch in the latest commit: projected awareness nodes now carry a canonical `vcs.repo` annotation when their entries agree, and the manifest parent spine reads that fact rather than the display label. Regression coverage uses `github.com/flotilla-org/flotilla` as the label and `flotilla-org/flotilla` as the fact, and asserts no duplicate repo segment.

PR #963, opened 2026-07-24
feat: template crew briefs

### LINKED ISSUE #960: Crew briefs as overridable minijinja templates (fleet/project/repo hierarchy), not hardcoded format! strings
Robert, 2026-07-24: hoping the prompting isn't hardcoded too deeply. Current state: `agent_adapter.rs` builds crew briefs by `format!` concatenation — hardcoded, but contained in one module, so the lift is cheap. minijinja is already a workspace dependency (the worktree provider uses it).

## Direction

- Extract the brief bodies into minijinja templates shipped as defaults.
- Django-esque override resolution: fleet (installation) → project → repo, most specific wins; an override can extend/replace named blocks rather than forking the whole template.
- Briefs should be SHORT and lean on skills, tools, and the environment — the template's real job is setting **interaction style**, **"what are we waiting for"** (the settlement gate the crew works toward), and **how aggressively to complete without feedback**. Everything mechanical (shepherd loop, never-merge, never-draft) stays, but as template blocks an installation can tune.
- Workflow variants select templates (the interactive-session workflow in #959 is the first second-consumer).
- Later: an agent that helps author/maintain the overrides (the same instinct as the skill adaptor, #955).

## Acceptance (first slice)

- [ ] Default briefs render byte-identical to today's output from templates
- [ ] A repo-level override file demonstrably changes one block for one repo's convoys
- [ ] Brief text differences between workflows expressed as template selection, not rust branches

rjwittams:
## Summary

- Move generated crew brief bodies into bundled minijinja templates while preserving the default `crew.md` output byte-for-byte.
- Add fleet/project/repo override resolution with block-only override layering, including repo-level overrides from `.flotilla/brief-templates/<template>`.
- Add `brief_template` selection to agent workflow crew entries and wire initial provisioning plus handoff-created sessions through the shared renderer.

Closes #960.

## Validation

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1066,6 +1066,7 @@ dependencies = [
  "flotilla-resources",
  "rstest",
  "sha2",
+ "tempfile",
  "tokio",
  "tracing",
 ]
--- a/crates/flotilla-controllers/Cargo.toml
+++ b/crates/flotilla-controllers/Cargo.toml
@@ -20,3 +20,4 @@ sha2 = "0.10"
 [dev-dependencies]
 tokio = { workspace = true, features = ["macros", "rt-multi-thread"] }
 rstest = { workspace = true }
+tempfile = "3"
--- a/crates/flotilla-controllers/src/reconcilers/vessel.rs
+++ b/crates/flotilla-controllers/src/reconcilers/vessel.rs
@@ -1,7 +1,9 @@
-use std::{collections::BTreeMap, marker::PhantomData};
+use std::{collections::BTreeMap, marker::PhantomData, path::PathBuf};
 
 use chrono::{DateTime, Utc};
-use flotilla_core::agent_adapter::{append_convoy_work_context, build_crew_brief, CrewAssignment, CrewBriefMember};
+use flotilla_core::agent_adapter::{
+    append_convoy_work_context, build_crew_brief_with_options, CrewAssignment, CrewBriefMember, CrewBriefTemplateResolver,
+};
 use flotilla_resources::{
     clone_key,
     controller::{
@@ -29,6 +31,7 @@ pub struct VesselReconciler {
     checkouts: TypedResolver<Checkout>,
     terminal_sessions: TypedResolver<TerminalSession>,
     namespace: String,
+    brief_templates: CrewBriefTemplateResolver,
 }
 
 impl VesselReconciler {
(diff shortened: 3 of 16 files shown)


rjwittams:
Fixed both points.

- Custom non-builtin `brief_template` names now treat a first block-only override as a variant of the bundled default crew template, so a `pairing.md` override containing only `{% block delivery %}...{% endblock %}` keeps the normal header, crew list, and assignment sections. Added `custom_block_only_template_inherits_the_default_brief_shape` for that path.
- The controller repo-override test now uses `tempfile::tempdir()` for RAII cleanup, with `tempfile` added as a dev-dependency for `flotilla-controllers`.

Re-ran:

rjwittams:
Addressed both non-blocking observations:
- Added an inline note beside `is_block_only_override` explaining that block-declaring standalone templates need explicit `{% extends %}` or no block declarations.
- `push_template_override` now ignores missing override files but emits a warning for other read errors, so permission/type/UTF-8 issues are visible instead of silently falling back.

Re-ran:

rjwittams:
Thanks, I checked both points against the current code.

On the controller I/O point: the premise here is off for the vessel reconciler path. The template resolution/rendering is already performed inside `VesselReconciler::fetch_dependencies` while building the `CreateTerminalSession` actuation, not inside `reconcile()`. `reconcile()` only consumes `VesselDeps` and maps the precomputed patch/actuations into a `ReconcileOutcome`, so this PR does not add filesystem reads to the synchronous reconcile method.

There are still small synchronous `read_to_string` calls inside the async dependency phase. I’m leaving those as-is for this PR because the resolver checks at most the bounded fleet/project/repo template files for the selected name, and switching this small shared helper to async/template preloading would widen the change without changing the behavior requested by #960. Non-`NotFound` read failures are now warned so bad override files are visible.

On multi-repo repo-level precedence: agreed that conflicting repo-local overrides are deterministic by `RepositoryKey`/`BTreeMap` order rather than a separate primary-repo concept. That mirrors the existing multi-repo workspace ordering in this reconciler and keeps the explicit hierarchy from the issue as fleet -> project -> repo; within the repo tier, multiple repo roots are layered in the order the reconciler supplies them. I’m not adding a primary-repo rule in this PR because the issue does not define one and it would introduce product semantics beyond templating support.

No code changes from this comment; current checks are green:
- Format
- Clippy
- Test
- claude-review

PR #968, opened 2026-07-24
fix: filter awareness issues by ready label

### LINKED ISSUE #966: Awareness band lists only ready issues (ADR 0018 alignment) — label filter as a query parameter
Robert, 2026-07-24: issues are dominating the sidebar — lots of them, not all ready. ADR 0018 already rules this: the awareness band contains "ready issues" (browse-scope = everything a principal COULD regard; for issues that is the earned `ready` label, per the #921 ready-label ruling). The aggregator currently ships every issue row into the awareness projection with no filter.

## What to build

- Add a **label filter parameter** to the scoped issues query family. The awareness assembly (aggregator input for `AwarenessKind::Issue` entries) requests issues with the `ready` label. This MUST be a query parameter, not a family-wide cut: the TUI project-page issue panel renders the same family unfiltered (browsing all open issues there stays correct).
- The label name is centralized (one constant / config point), not scattered as string literals — per-project configurability is the known future, don't build it, just don't preclude it.
- No display/template changes in this ticket — compact rendering and class toggles are andamento-side and gated on the template-model design (andamento#26). Volume reduction only.

## Acceptance

- [ ] Sidebar issue latents show only `ready`-labeled issues (live check in andamento against flotilla-org repos)
- [ ] Project-page issue panel still lists open issues regardless of label
- [ ] Filter expressed as a query parameter with tests at the query layer, not post-filtering in the projection

rjwittams:
## Summary

- add an optional label parameter to scoped issue query identities and result/delta rows
- request the centralized `ready` label for awareness-band issue materialization
- keep project-page issue panels and source searches unfiltered
- pass label filters through the GitHub provider and preserve them during incremental refreshes

## Why

Awareness projections subscribed to the same unfiltered open-issue windows used by project browsing, so every open issue could appear in the sidebar. ADR 0018 defines awareness issues as those that have earned the `ready` label, while project pages must continue to browse all open issues.

## Validation

Closes #966

### DIFF
--- a/crates/flotilla-commands/src/commands/issue.rs
+++ b/crates/flotilla-commands/src/commands/issue.rs
@@ -64,7 +64,7 @@ impl IssueNoun {
                     // SENTINEL: repo is empty — dispatch must fill it from --repo or FLOTILLA_REPO.
                     action: CommandAction::QueryIssues {
                         repo: RepoSelector::Query("".into()),
-                        params: IssueQuery { search: Some(query.join(" ")) },
+                        params: IssueQuery { search: Some(query.join(" ")), label: None },
                         page: 1,
                         count: 50,
                     },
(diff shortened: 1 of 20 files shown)


rjwittams:
Fixed in 9fa2852a. Added an incremental materializer regression test with a `ready`-filtered query that verifies an issue losing the label is evicted and an issue gaining it is admitted into the live window. The targeted test and daemon clippy both pass.

rjwittams:
Acknowledged. The additional provider window is the intended consequence of keeping the ready-only awareness query distinct from the unfiltered project-page query, as required by #966. No code change is needed for this non-blocking rate-budget observation.

PR #976, opened 2026-07-24
fix: bound host readiness by heartbeat freshness

### LINKED ISSUE #962: Host resource stays ready=true with hours-stale heartbeat after peer sleeps
Live specimen 2026-07-24: feta went to sleep overnight. ~10.5 hours later, `flotilla resource list host` on kiwi still shows the feta host environment `ready: true` with `heartbeat_at: 01:22` (stale by the full sleep duration). Convoy dispatch to feta failed correctly ("no peer summary for host environment …"), so the placement layer knows the peer is gone — but the host resource lies.

- Readiness should be derived from heartbeat freshness (or flipped by the peer-connection teardown that the placement layer evidently observed), not left at the last written value.
- This is the same family as ADR 0018's "disconnected becomes a visible state" refinement (`status.connectivity=disconnected` as an annotation-tier fact) — the host-resource side of that twin. Surfaces reading the host list get affirmative staleness today.
- Error-message nit while here: the dispatch failure surfaced as a color-eyre backtrace dump from src/main.rs:622 rather than a one-line "peer host feta is not connected" — placement resolution failure is an expected operational condition, not a crash.

## Acceptance

- [ ] A slept/disconnected peer's host resource stops reporting ready (bounded staleness, e.g. missed-heartbeat threshold, or teardown-driven)
- [ ] Dispatch to an unavailable host prints a one-line diagnosis naming the host, no backtrace

rjwittams:
Closes #962.

## What changed
- Host resources now opt into home-bound runtime replication so peer host rows can be listed through the resource overlay.
- Host readiness exposed through resource list/get/watch is derived from heartbeat freshness with a 60s TTL, so stale heartbeats no longer report `ready: true`.
- Host-direct placement admission applies the same heartbeat freshness rule before trusting `Host.status.ready`.
- Unavailable host-direct dispatch now reports a one-line `peer host <name> is not connected` diagnostic instead of bubbling a color-eyre report.

## Validation

### DIFF
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -3169,9 +3169,10 @@ async fn validate_workflow_agent_adapters(
             let host = backend.clone().using::<ResourceHost>(namespace).get(&host_direct.host_ref).await.map_err(|error| {
                 format!("placement `{}` host `{}` is not ready: {error}", placement.metadata.name, host_direct.host_ref)
             })?;
-            let status = host
+            let mut status = host
                 .status
                 .ok_or_else(|| format!("placement `{}` host `{}` has no observed status", placement.metadata.name, host_direct.host_ref))?;
+            status.apply_heartbeat_readiness(Utc::now());
             if !status.ready {
                 return Err(format!("placement `{}` host `{}` is not ready", placement.metadata.name, host_direct.host_ref));
             }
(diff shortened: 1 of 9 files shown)


rjwittams:
Fixed. `run_query_command` now matches the non-query path by returning `Ok(CommandValue::Error)` after formatting the result, so `main.rs` can exit nonzero without printing a second JSON document. Added `query_command_error_returns_result_after_formatting` to cover the behavior.

rjwittams:
Reviewed. The remaining notes are non-blocking: the boundary behavior is already covered through the resource projection and placement-admission tests added in this PR, and the decode/mutate/encode path is isolated to the Host read projection with tiny host counts. No further code change needed for this issue.

PR #987, opened 2026-07-24
fix: gossip learned peer display names

### LINKED ISSUE #908: Route table shows learned-but-unconfigured peers as raw node-id hashes
Seen 2026-07-23 on feta after the f4333f23 bump: `flotilla topology` lists a route learned via kiwi as `1c8df992acde8c3e863343c568a95be7` (udder's node id) because feta has no configured name for it. The route is correct; the display is illegible, same class as #903's legibility recast and #907's hash paths.

Direction: peers should gossip their self-declared node name alongside the id, and the table should render `udder (learned via kiwi)` — raw ids only as a last-resort fallback, never the default rendering for a healthy route. Host list omitting unnamed learned peers is arguably honest; the route table showing a bare hash is not informative enough to act on.

rjwittams:
Closes #908.

Peers now include their self-declared display name with both ordinary snapshot gossip and routed resync snapshots. The peer manager retains those learned names and uses them when constructing topology routes, falling back to raw node IDs only when no name is known.

A regression test covers feta learning `udder-id` via `kiwi-id` and verifies that the route target is rendered with the learned `udder` name.

## Verification

### DIFF
--- a/crates/flotilla-daemon/src/peer/channel_tests.rs
+++ b/crates/flotilla-daemon/src/peer/channel_tests.rs
@@ -23,6 +23,7 @@ fn snapshot_msg(origin: &str, repo: &RepoIdentity, seq: u64) -> PeerDataMessage 
     }
     PeerDataMessage {
         origin_node_id: node(origin),
+        origin_display_name: origin.to_string(),
         repo_identity: repo.clone(),
         repository_key: None,
         host_repo_root: Some(PathBuf::from("/repo")),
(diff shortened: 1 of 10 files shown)


rjwittams:
Thanks — verified both non-blocking observations against the project guidance and existing peer-manager behavior. Required fields are intentional during the repository's explicit no-backwards-compatibility phase, and retaining learned names across disconnects matches the existing diagnostic identity retention policy. No code changes needed.

PR #989, opened 2026-07-24
fix: supervise overlay replication by generation

### LINKED ISSUE #923: Add generation-aware retry supervision for overlay replicators
From PR #918 review by @claude: https://github.com/flotilla-org/flotilla/pull/918#issuecomment-5059931266

Per-kind overlay replicator tasks currently stop after a non-connection error such as a malformed watch event. Add an independent retry/backoff supervisor so transient kind-local failures resume without waiting for a full SSH reconnect.

The supervisor must be generation-aware and cancel or supersede the prior task when a new `PeerConnectedNotice` arrives. A naive retry loop can survive a tunnel reconnect and race a newly spawned replicator, causing duplicate watches and writes.

Acceptance:

- transient per-kind HTTP/watch failures retry with bounded backoff
- a new peer connection generation cancels/supersedes older retrying tasks
- only one active replicator exists per `(peer, kind, connection generation)`
- tests cover malformed events, tunnel reconnect during backoff, and no duplicate application

rjwittams:
## Summary

- supervise each home-bound runtime resource replicator with capped exponential retry backoff
- track the active connection generation per peer and cancel older retry loops on reconnect
- ignore duplicate and stale connection notices so each peer/kind/generation has one application stream
- cover malformed watch events, reconnect cancellation during backoff, and duplicate application suppression

## Validation

Closes #923

### DIFF
--- a/crates/flotilla-daemon/src/server/peer_runtime.rs
+++ b/crates/flotilla-daemon/src/server/peer_runtime.rs
@@ -17,7 +17,7 @@ use tokio::sync::{mpsc, Mutex};
 use tracing::{debug, info, warn};
 
 use super::{
-    remote_commands::RemoteCommandRouter, replicator::spawn_peer_replicators, shared::sync_peer_query_state, PeerConnectedNotice,
+    remote_commands::RemoteCommandRouter, replicator::PeerReplicatorSupervisors, shared::sync_peer_query_state, PeerConnectedNotice,
     SshTransport,
 };
 use crate::peer::{dispatch_pending_sends, peer_resource_socket_path, HandleResult, InboundPeerEnvelope, PeerManager, PeerSender};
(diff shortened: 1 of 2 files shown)


rjwittams:
Fixed both correctness issues in `fda225ae`.

- Backoff now resets to its initial delay after an attempt remains healthy for 60 seconds, with a paused-time regression test. Clean watch endings log at debug while failures remain warnings.
- The newer generation still cancels the old generation immediately, because allowing an old tunnel generation to survive would reintroduce the duplicate-writer race. Instead, Unix-socket HTTP backend construction now occurs inside each kind's generation-scoped supervised attempt, so construction failures receive the same capped retry behavior without reviving stale work.

The pinned format check, workspace clippy with warnings denied, and full locked workspace tests all pass locally.

rjwittams:
Confirmed. The re-review found no new issues, so no further code changes are needed for this cycle.

PR #990, opened 2026-07-24
fix: polish daemon reconnect experience

### LINKED ISSUE #909: Polish TUI daemon-reconnect: narrate, preserve UI state, self-re-exec across wire bumps
The TUI survives daemon bounces (`src/main.rs` outer loop: `EventLoopExit::DaemonDisconnected` → `connect_or_spawn` with backoff → new `App`). First live exercise 2026-07-23 (fleet bump to f4333f23, same wire v17): it worked, and Robert's verdict was "decent feature, could use some polish — doesn't quite save enough of its state or say exactly what it is doing". Three gaps:

## 1. Narrate the transition
During the backoff loop the terminal is restored and silent (`info!` logs only). Render a minimal 'daemon disconnected — reconnecting (attempt N)…' screen while retrying, and a transient status-line note after reconnect. The user should never have to infer from a flicker.

## 2. Preserve UI state across the reconnect
The reconnect constructs a brand-new `App`, so all in-memory state dies: selected view, navigation/drill stack, cursors and scroll positions, multi-selections, search input, event-log contents. Only the persisted tab set (`open-views.toml`) survives. Either reuse the `App` and reset just the daemon-derived caches (result sets resync anyway via `resync_subscriptions`), or snapshot/restore `UiState` + per-view navigation around the rebuild. Full rebuild is the blunt instrument.

## 3. Self-re-exec across incompatible wire bumps
On `is_incompatible_daemon_error` the TUI hard-exits with an error. The binary on disk at that moment is almost always the *new* build (that's why the daemon is newer) — so instead: re-exec `current_exe` (carrying the state snapshot from gap 2 via a handoff file/env), and the fleet-bump ritual's final step ('tell the human to restart their TUI panes') disappears. Until then, note the subtle trap observed live: a same-wire reconnect makes a **stale TUI look fresh** — the pre-bump binary reconnects happily and keeps running old TUI code against the new daemon, silently. At minimum surface a daemon-build vs TUI-build mismatch hint in the status/help area.

rjwittams:
Closes #909

## Summary

- narrate TUI reconnect attempts in-terminal and show a post-reconnect status note
- retain the live `App` across daemon bounces while clearing and replaying daemon-derived caches
- share reconnect/backoff handling across the TUI, `watch`, and `pm connect`
- advertise build identity through the compatible Hello display field and surface same-wire stale-client mismatches
- self-re-exec once on wire incompatibility, carrying view/navigation/table/repo/search/event-log state through a JSON handoff

## Validation

### DIFF
--- a/crates/flotilla-client/build.rs
+++ b/crates/flotilla-client/build.rs
@@ -27,6 +27,5 @@ fn main() {
         .filter(|value| !value.is_empty())
         .or_else(|| git_output(&["rev-parse", "--short=12", "HEAD"]))
         .unwrap_or_else(|| "unknown".to_string());
-
     println!("cargo::rustc-env=FLOTILLA_BUILD_ID={build_id}");
 }
--- a/crates/flotilla-client/src/lib.rs
+++ b/crates/flotilla-client/src/lib.rs
@@ -18,6 +18,8 @@ use flotilla_transport::message::{connect_unix_message_session, MessageSession};
 use tokio::sync::{broadcast, oneshot, Mutex};
 use tracing::{debug, error, warn};
 
+pub mod reconnect;
+
 /// Std RwLock for local seq tracking — the critical sections are single HashMap
 /// operations (no async work while holding the lock), and using a sync lock
 /// avoids the race where a spawned seq update hasn't run before the next delta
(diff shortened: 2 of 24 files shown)


rjwittams:
Fixed the two substantive findings in 0b5500d1.

- `reconnect_daemon` now clears `project_issue_start_batches`, `command_project_issue_starts`, and `pending_cancel`, with the reconnect regression test seeding and asserting all three states.
- Build-ID generation moved out of the protocol crate and into the client and daemon leaf crates. The protocol crate now only owns the compatible display-name encoding, so a commit no longer invalidates the protocol crate and its nearly workspace-wide reverse dependency graph.

I retained the successful-connect guard cleanup in the shared reconnect helper intentionally: every long-lived mode must clear that inherited guard after a successful post-exec connection, otherwise a later fleet bump in the same process would be incorrectly suppressed. For callers that did not re-exec it remains a no-op.

rjwittams:
Confirmed. No further code change is needed from this re-review; the duplicated leaf build scripts are an intentional small trade-off to avoid introducing build-time workspace coupling while keeping build identity accurate.

PR #991, opened 2026-07-24
fix: assert canonical repo path round trip

### LINKED ISSUE #868: test: dispatch_add_list_remove_repo_round_trip fails on macOS (TMPDIR canonicalization)
## What happened

`server::tests::dispatch_add_list_remove_repo_round_trip` (crates/flotilla-daemon/src/server/tests.rs:277) fails on macOS on clean main:

```
left:  Some("/private/var/folders/.../T/.tmpUB2EiY/repo-a")
right: Some("/var/folders/.../T/.tmpUB2EiY/repo-a")
```

One side of the round trip canonicalizes the temp path (macOS `/var` → `/private/var` symlink), the other doesn't. CI (Linux) never sees it, so it silently taxes every macOS contributor's local `cargo test --workspace` run.

## What to build

Make the comparison canonicalization-consistent: either canonicalize the expected path in the test the same way the code under test does, or (better) pin down which layer canonicalizes repo paths and assert on that layer's output. Do not weaken to a suffix match.

## Acceptance criteria

- [ ] `cargo test --workspace --locked` passes on macOS
- [ ] The test still catches an actual path round-trip regression

rjwittams:
## Summary

- assert that the daemon lists the canonical form of a tracked repository path
- keep the complete path round-trip assertion so unrelated path regressions still fail
- pin canonicalization to the daemon layer instead of canonicalizing the daemon's output inside the assertion

## Root cause

On macOS, `TMPDIR` may use the `/var/...` spelling while repository-path normalization resolves that symlink to `/private/var/...`. The original test compared the daemon's canonical path with the lexical tempfile path. A later incidental change canonicalized both sides during comparison, which avoided the failure but no longer asserted that the daemon itself owns canonicalization.

This change computes only the expected canonical path and compares the listed path to it exactly.

## Validation

- `cargo test --workspace --locked` (macOS)

Closes #868

### DIFF
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -4163,7 +4163,8 @@ impl InProcessDaemon {
     ///
     /// If `path` is a git worktree, the main repo root is resolved via
     /// `git rev-parse --path-format=absolute --git-common-dir` and tracked
-    /// instead. `resolved_from` is `Some(original_path)` in that case.
+    /// instead. The returned `tracked_path` is canonicalized, and
+    /// `resolved_from` is `Some(original_path)` when the repo root changes.
     pub async fn add_repo(&self, path: &Path) -> Result<AddRepoOutcome, String> {
         let (path, resolved_from) = self.normalize_repo_path(path).await;
 
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -4163,8 +4163,9 @@ impl InProcessDaemon {
     ///
     /// If `path` is a git worktree, the main repo root is resolved via
     /// `git rev-parse --path-format=absolute --git-common-dir` and tracked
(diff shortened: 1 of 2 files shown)


rjwittams:
Addressed the non-blocking documentation note. The public `add_repo` contract now states that `tracked_path` is canonicalized, matching both `normalize_repo_path` and the strengthened round-trip assertion. The exact format, clippy, and workspace test gates remain green locally on macOS.

rjwittams:
Fixed the overstatement. The `add_repo` documentation now scopes the canonical `tracked_path` guarantee to inputs that resolve to a Git repository root, matching the `Some(repo_root)` branch in `normalize_repo_path`. Pinned format, clippy, and the full workspace suite are green locally on macOS after the tweak.

rjwittams:
Confirmed. No further changes are needed: the full-path regression assertion and the scoped `add_repo` documentation now match the `normalize_repo_path` contract. All final checks pass.

PR #993, opened 2026-07-24
fix: make mixed-format issue ID sorting transitive

### LINKED ISSUE #848: fix: make mixed-format issue ID sorting transitive
Follow-up from PR #842: https://github.com/flotilla-org/flotilla/pull/842#issuecomment-5044462310

`IssueRef::cmp_id_desc` currently uses numeric descending comparison only when both IDs parse as `u64`, otherwise lexical comparison. For a panel aggregating numeric and opaque IDs, this can create a comparison cycle (for example `"10"`, `"9"`, and `"3abc"`), which violates the transitivity required by `sort_by`.

Define a total mixed-format ordering by bucketing numeric and opaque IDs before applying the per-bucket descending order, with the source as the final deterministic tie-breaker. Add coverage for numeric/opaque mixtures, opaque pairs, and same-ID cross-source ties.

Out of scope for #842: GitHub is currently the only registered issue provider, so its IDs are numeric and the project-panel ordering fix is correct.

rjwittams:
## Summary

- bucket numeric and opaque issue IDs before comparing them
- sort numeric IDs descending numerically and opaque IDs descending lexically
- preserve source-qualified deterministic tie-breaking
- cover mixed-format transitivity, numeric/opaque precedence, opaque pairs, and same-ID cross-source ties

## Root cause

`IssueRef::cmp_id_desc` selected numeric or lexical comparison based on each pair. A set containing both numeric and opaque IDs could therefore produce a comparison cycle, violating the transitivity required by `sort_by`.

## Validation

Closes #848

### DIFF
--- a/crates/flotilla-protocol/src/provider_data.rs
+++ b/crates/flotilla-protocol/src/provider_data.rs
@@ -1,4 +1,4 @@
-use std::path::PathBuf;
+use std::{cmp::Ordering, path::PathBuf};
 
 use chrono::{DateTime, Utc};
 use indexmap::IndexMap;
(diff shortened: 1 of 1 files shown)


rjwittams:
Reviewed. No changes needed: the comment confirms the comparator is transitive, the requested cases are covered, and both observations are explicitly non-blocking pre-existing behavior.

PR #995, opened 2026-07-24
fix: select a viable local default convoy placement

### LINKED ISSUE #840: fix: default convoy placement picks an arbitrary (adapter-less) host now that multiple host-direct policies exist
Since #826 materialized host-direct placement policies for peer hosts, `default_convoy_placement_policy` selects among several — and picks one arbitrarily (observed: alphabetically-first `host-direct-b2aac222…`, a host advertising **zero** agent adapters), so every default-placement `convoy start` fails admission with 'workflow requires agent adapter codex, which is not available in placement …'. Observed 2026-07-22 dispatching a batch; explicit `--placement-policy` is the workaround.

Default selection should:
1. **Filter by admission-viability**: only policies whose target advertises the workflow's required adapters (the same capability check admission runs — do it during selection, not after).
2. **Prefer the local host** among viable candidates (least surprise; remote is an explicit choice or a future Quartermaster decision — see the #812 thread's allocation ladder; do NOT build load-awareness here).
3. Fail with the candidate list when nothing is viable ('no placement policy satisfies adapter codex; candidates: …').

Acceptance:
- [ ] Bare `convoy start` on a host with a viable local policy places locally
- [ ] Hosts advertising no required adapters are never default-selected
- [ ] No-viable-candidate error names the requirement and the candidates

rjwittams:
Closes #840

## What changed

- evaluate every stance-compatible placement policy with the same agent-adapter admission check used after selection
- prefer viable policies targeting the local host, while retaining deterministic host-direct/name ordering
- report the required adapter(s) and the complete candidate list when no default policy is viable
- cover selection behavior and a bare `convoy start` command that persists the viable local placement

## Root cause

Default selection sorted the materialized placement policies and chose the first host-direct policy before checking workflow adapter requirements. With peer host-direct policies present, an alphabetically earlier host advertising no adapters could be selected and then rejected by admission.

## Validation

### DIFF
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -753,25 +753,59 @@ async fn persist_adopted_checkout(
     }
 }
 
-async fn default_convoy_placement_policy(backend: &ResourceBackend, namespace: &str, contained: bool) -> Option<String> {
+async fn default_convoy_placement_policy(
+    backend: &ResourceBackend,
+    namespace: &str,
+    workflow: &WorkflowTemplateSpec,
+    local_host_ref: Option<&str>,
+) -> Result<Option<ResourceObject<PlacementPolicy>>, String> {
+    let contained = workflow.vessels.iter().any(|vessel| vessel.stance == flotilla_resources::Stance::Contained);
     let mut policies = match backend.clone().using::<PlacementPolicy>(namespace).list().await {
         Ok(list) => list.items,
         Err(err) => {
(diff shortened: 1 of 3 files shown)


rjwittams:
Review-cycle summary: verified the observations against the implementation. They are explicitly non-blocking and do not identify a correctness issue within #840's acceptance scope, so no code change is needed. The candidate count is expected to remain small; the required-adapter and stance scans are cheap; and the requested no-viable-candidate contract intentionally prioritizes the requirement plus deterministic candidate list. Per-candidate rejection diagnostics can be considered separately if operational evidence warrants it.

PR #996, opened 2026-07-24
feat: expose awareness projection truncation

### LINKED ISSUE #880: Surface awareness projection truncation state
From PR #877 review by @claude: https://github.com/flotilla-org/flotilla/pull/877#issuecomment-5048112607

`project_awareness` applies `AwarenessLimit` to groups and entries after ranking, but the result set does not expose whether awareness rows were truncated separately from the underlying issue query's own `has_more` state.

Add an explicit truncation signal so UI surfaces can distinguish a complete awareness tree from a limited one.

rjwittams:
Closes #880.

## Summary
- add projection-level `truncated` state independently of demand-backed `has_more`
- mark awareness results truncated when ranked group or entry limits discard rows
- preserve projection truncation through result-state composition and wire serialization
- cover group limits, entry limits, exact-limit completeness, and protocol round-tripping

## Verification

### DIFF
--- a/crates/flotilla-core/src/aggregator_projection.rs
+++ b/crates/flotilla-core/src/aggregator_projection.rs
@@ -365,6 +365,7 @@ fn merged_issue_state(issue_sets: &[(QueryScope, ResultSet)]) -> ResultSetState 
     let mut state = ResultSetState::default();
     for (_, set) in issue_sets {
         state.conditions.extend(set.state.conditions.clone());
+        state.truncated |= set.state.truncated;
         if let Some(demand) = &set.state.demand {
             state.demand = Some(match state.demand {
                 Some(existing) => flotilla_protocol::DemandBackedMetadata {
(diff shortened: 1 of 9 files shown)


rjwittams:
The result-set signal is the complete scope of #880, so I am not adding a TUI rendering policy in this PR.

The issue identifies the contract defect specifically: `project_awareness` applies `AwarenessLimit`, but its result set does not expose that fact separately from demand-backed `has_more`; it asks for an explicit signal *so* UI surfaces can distinguish those states. `ResultSetState::truncated` now provides that consumer-facing distinction through the protocol and is retained in the TUI's awareness query state.

Mapping it through `result_set_meta` would not actually complete that contract: awareness has no direct `TableView`, as the review notes. More importantly, the project panels render their own complete query families; awareness family counts and summaries are computed before entry truncation, so marking those tables `limited` would conflate truncation of an enrichment tree with truncation of the panel rows. Choosing where and how a surface presents awareness-tree completeness is a separate surface policy and was not requested by #880.

No code change is needed for this comment. The PR closes the stated result-set exposure gap while leaving presentation consumers free to render the new state accurately for their own awareness-tree affordances.

PR #1008, opened 2026-07-25
fix: re-resolve peer socket paths between retries

### LINKED ISSUE #998: Replicators bind the forwarded socket path at spawn — transport-level reconnects strand them on a dead socket
Second wedge class, observed live 2026-07-25 right after verifying #992 (which fixed the first): a one-sided feta restart where the ssh transport performs its OWN reconnect ("peer connection established" → "remote daemon restarted (session_id changed), clearing stale data") reuses the existing connection generation — this path never calls activate_connection, so no new generation is minted (correct per the transport's view: same logical connection). The replicator supervisor therefore (correctly) refuses a duplicate spawn for the same generation — but the surviving replicator retry loops were handed the forwarded resource-socket PATH at spawn time, and that socket died with the old tunnel. They retry forever against it: `LIST resources: error sending request` every backoff, heartbeats frozen, host NOT-READY.

Evidence (kiwi daemon.log, 12:25:41): reconnect + session-changed + "sending local state … generation=2" with gen-2 replicators continuing to fail LIST/WATCH against the pre-reconnect socket path from 12:24:53 onward. The udder peer shows the same signature.

## Fix direction (preferred)

**Replicators must re-resolve the forwarded socket path per attempt**, not bind it at spawn: pass a socket-path source (lookup against the live peer runtime state) into the retry loop, so a transport reconnect heals on the next backoff tick without any generation ceremony. This also fixes the sibling hole where spawn finds "no forwarded resource socket" and returns without scheduling any retry.

Alternative (weaker): treat session_id change as a full re-activation (mint a new generation, cancel + respawn). More churn, and still leaves the no-socket-at-spawn hole.

## Acceptance

- [ ] One-sided peer daemon restart via transport auto-reconnect: replication and heartbeats resume within one backoff interval, no daemon restarts needed
- [ ] Spawn with no forwarded socket yet: replicators wait/retry instead of returning silently
- [ ] Test at the replicator level with a socket-path source that changes mid-retry

rjwittams:
## What changed

- keep a generation-scoped live source for each peer's forwarded resource socket path
- refresh that source on same-generation reconnect notices without spawning duplicate replicators
- resolve the current path before every replication attempt
- keep replicators retrying when a generation starts before a forwarded socket exists
- cover changed and initially-missing socket sources at the replicator retry seam

## Why

Transport-level SSH reconnects can preserve the logical peer connection generation while replacing the forwarded resource socket. Replicators previously retained the path supplied when they spawned, and a generation that started without a socket returned without scheduling retries. Either case could strand replication and heartbeat updates until a daemon restart.

## Impact

A one-sided peer daemon restart can heal replication on the next backoff attempt without minting a new connection generation or restarting either daemon. Peers whose outbound forwarded socket appears after replication supervision starts also recover automatically.

## Validation

Closes #998

### DIFF
--- a/crates/flotilla-daemon/src/server/replicator.rs
+++ b/crates/flotilla-daemon/src/server/replicator.rs
@@ -1,10 +1,4 @@
-use std::{
-    collections::HashMap,
-    future::Future,
-    path::PathBuf,
-    sync::{Arc, RwLock},
-    time::Duration,
-};
+use std::{collections::HashMap, future::Future, path::PathBuf, sync::Arc, time::Duration};
 
 use chrono::Utc;
 use flotilla_core::{daemon::DaemonHandle, in_process::InProcessDaemon};
@@ -13,6 +7,7 @@ use flotilla_resources::{
     HttpBackend, K8sWatchEvent, ReplicationClass, Resource, ResourceBackend, ResourceError, ResourceList, ResourceObject, WatchStart,
 };
 use futures::StreamExt;
+use tokio::sync::watch;
 use tokio_util::sync::CancellationToken;
 use tracing::{debug, warn};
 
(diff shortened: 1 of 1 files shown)


rjwittams:
Fixed in `a8a505fa`. An unresolved socket-path source now waits on a watch-channel update under the supervisor's cancellation select instead of returning an error into the backoff loop. Inbound-only peers therefore remain silent without polling, while a same-generation outbound socket update wakes the existing replicators immediately. The regression test advances virtual time while the source is empty and verifies there is no second resolution attempt before the path is published.

rjwittams:
Thanks for the re-review. No further change is needed for this PR: the parked source wait is required so a same-generation socket can appear later, and the production branch's core behavior is covered at the replicator seam. I filed #1014 for the valid pre-existing lifecycle gap around cancelling and pruning supervisors after a permanent peer disconnect. The clone/match nit is harmless and does not justify another code cycle.

PR #1015, opened 2026-07-25
fix: preserve whole-repo project during identity migration

### LINKED ISSUE #999: Project resource silently vanished (flotilla project gone after apply of overlapping two-repo project + daemon double-restart)
Observed 2026-07-25 ~12:55: after (a) `project apply presentation` creating a two-repo project whose repositories INCLUDE the flotilla repo, and (b) a double daemon restart, `flotilla convoy start --project flotilla` failed with `resource not found: flotilla` — the flotilla Project resource was gone from the store. Every other project survived, including `presentation` (both repos intact) and `andamento` (whose repo is ALSO in presentation, and which survived — so "repo claimed by another project" alone doesn't explain it).

Sequence candidates to investigate:
- `project apply` semantics: does applying a spec that references an already-projected repo trigger any reconciliation that removes/renames another project?
- Track-time materialization / teardown interplay with the Plane-A repo winnow (repos/*.toml reduced to flotilla-only two bounces earlier — but the flotilla project is the one that vanished, which is exactly backwards from a winnow-driven teardown).
- Restart-time project reconciliation racing the freshly-applied overlapping project.

Recovered by `project add ~/dev/flotilla` (new resource, same slug — note #979/rename-stable-identity implications if anything referenced the old resource UID).

No dogfood impact beyond a failed dispatch, but silent resource deletion is a store-trust issue — worth root-causing, not shrugging.

rjwittams:
## Summary

- restrict whole-repository Project reconciliation candidates to Projects with exactly one unsliced repository
- preserve distinct multi-repository Projects that overlap a tracked repository during identity migration
- add a restart regression covering `project apply` of an overlapping two-repository Project

## Root cause

Restart-time tracked-repository materialization can discover that a checkout's repository identity changed from its local key to its canonical remote key. Reconciliation migrated every referencing Project correctly, but then selected any Project containing the repository as the primary whole-repository Project. A distinct multi-repository Project could therefore become primary, causing duplicate cleanup to silently delete the generated-name whole-repository Project.

The primary candidate is now required to actually be a whole-repository definition: exactly one repository with no subpath.

## Impact

Durable Project identities no longer disappear when a separately applied multi-repository Project overlaps a tracked repository and repository identity is reconciled during daemon restart.

## Validation

- `cargo test -p flotilla-daemon --locked --test project_cli`

Closes #999

### DIFF
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -3979,7 +3979,12 @@ impl InProcessDaemon {
         let spec = whole_repository_project_spec(repository_key.clone(), display_name)?;
         let primary_name = project_objects
             .iter()
-            .filter(|project| project.spec.repositories.iter().any(|entry| entry.repo == repository_key && entry.subpath.is_none()))
+            .filter(|project| {
+                matches!(
+                    project.spec.repositories.as_slice(),
+                    [entry] if entry.repo == repository_key && entry.subpath.is_none()
+                )
+            })
             .min_by_key(|project| {
                 (
                     !migrated_project_names.contains(&project.metadata.name),
--- a/crates/flotilla-daemon/tests/project_cli.rs
+++ b/crates/flotilla-daemon/tests/project_cli.rs
@@ -538,6 +538,70 @@ async fn daemon_start_backfills_project_idempotently_and_preserves_edits() {
 }
 
 #[tokio::test]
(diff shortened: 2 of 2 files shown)


rjwittams:
Thanks — confirmed this is a valid adjacent case in the no-primary/name-collision fallback. I filed #1017 with a concrete regression scenario and the matching exact-whole-repository predicate; keeping it separate avoids expanding the restart-time deletion fix after its checks passed.

PR #1018, opened 2026-07-25
fix: materialize remote convoy repositories

### LINKED ISSUE #1001: Remote placement fails when the host lacks a Repository record — materialize from ConvoySpec.repositories
First multi-repo convoy dispatch to a peer host failed at provisioning (2026-07-25): `repository 18bgq9… not found` — feta's store had no Repository record for andamento (no andamento convoy had ever run there). The ConvoySpec carries everything needed (`repositories[].url`, base_ref, workspace_slug), and `ensure_repository` machinery exists — remote placement should materialize missing Repository records from the spec instead of failing on catalog lookup.

## Acceptance

- [ ] Dispatching a convoy to a host that has never seen one of its repos clones/registers it from the spec's URL and proceeds
- [ ] Multi-repo case covered (the failing specimen: two-repo project, one repo known to the host, one not)
- [ ] Honest failure remains for a URL that cannot be cloned (auth, nonexistent)

Specimen: convoy entities-only-wire-andamento-grouping on project presentation, feta placement. Relates #870 (multi-repo projects), #756 (multi-repo workspaces).

rjwittams:
## Summary

- materialize missing destination Repository records from ConvoySpec repository URLs
- verify the reconstructed repository identity against the convoy repository key before provisioning
- cover a two-repository remote placement where one repository is already known and one is new
- preserve clone failures for inaccessible or unauthorized remotes

## Validation

Closes #1001

### DIFF
--- a/crates/flotilla-controllers/src/reconcilers/vessel.rs
+++ b/crates/flotilla-controllers/src/reconcilers/vessel.rs
@@ -5,16 +5,16 @@ use flotilla_core::agent_adapter::{
     append_convoy_work_context, build_crew_brief_with_options, CrewAssignment, CrewBriefMember, CrewBriefTemplateResolver,
 };
 use flotilla_resources::{
-    clone_key,
+    canonicalize_repo_url, clone_key,
     controller::{
         delete_lifecycle_owned_matching, Actuation, LabelJoinWatch, LabelMappedWatch, ReconcileOutcome, Reconciler, SecondaryWatch,
     },
     repository_workspace_slugs, Checkout, CheckoutPhase, CheckoutSpec, CheckoutWorktreeSpec, Clone, ClonePhase, CloneSpec, Convoy,
     CrewSource, DockerCheckoutStrategy, DockerEnvironmentSpec, Environment, EnvironmentMount, EnvironmentMountMode, EnvironmentPhase,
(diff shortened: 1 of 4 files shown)


rjwittams:
Fixed the responsiveness concern in `525fac18`. A missing repository now contributes both `CreateRepository` and its downstream clone/checkout actuations in the same reconciliation pass; controller actuation ordering registers the repository before creating the clone, so progress no longer depends on a later Repository watch or periodic resync.

I also changed the existing full controller-loop test to start with a single missing repository while the Vessel loop retains its 60-second resync interval. It reaches Ready in the normal 3-second assertion window and verifies the Repository was materialized. The mixed known/missing two-repository regression remains in place and passes.

Added a comment explaining the deliberate pre-canonicalization of SCP-like clone transports. I did not add separate invalid-URL and key-mismatch tests: those branches directly compose the already-tested canonicalization/key-verification helpers, while the acceptance-level inaccessible/authenticated transport behavior is covered through the clone failure regression. All exact workspace format, Clippy, and test gates pass after the revision.

rjwittams:
Verified the re-review. The remaining duplicate `verify_key` call is harmless and explicitly non-blocking; I am leaving the already-green revision unchanged to avoid another CI cycle for a purely cosmetic cleanup. The unconditional verification continues to protect existing catalog records, while the pre-actuation check keeps the missing-record branch locally explicit.

PR #1019, opened 2026-07-25
fix: reply to peer keepalive pings

### LINKED ISSUE #1016: Peer keepalive watchdog fires on idle links — no ping sender, so quiet tunnels die every 90s
Third member of the wedge family (after #992 generations, #1008 socket re-resolution — both landed). With those fixed, the underlying churn is visible: `keepalive timeout — no messages received in 90s` kills the transport on IDLE links, then reconnect succeeds (gen+1, state re-push, re-list). On a quiet mesh this repeats every ~90s forever: readiness flickers through the 60s heartbeat TTL (observed: dispatch batch interrupted mid-loop by a flicker), every cycle pays a full reconnect + state push + re-list, and logs churn.

The watchdog assumes traffic; nothing guarantees traffic. Fix: send periodic keepalive PINGs (interval well under the 90s watchdog, e.g. 20–30s) so idle links stay alive; the watchdog then only fires on genuinely dead transports.

## Acceptance

- [ ] An idle peer link stays connected indefinitely (no reconnect cycles in a quiet hour)
- [ ] Host readiness stable on a quiet mesh (no TTL flicker)
- [ ] Genuinely dead transport still detected within the watchdog window

rjwittams:
## Summary

- reply to inbound peer `Ping` messages with a same-timestamp `Pong`
- keep replies generation-aware by handling them after the existing stale-connection guard
- cover responsive idle links and silent dead links with deterministic watchdog tests

## Root cause

The dialing peer already sent a keepalive `Ping` every 30 seconds, but the accepting peer ignored it without replying. The dialer therefore received no traffic and forced a reconnect after the 90-second watchdog window even though the transport remained healthy.

With `Pong` replies, quiet links produce bidirectional keepalive traffic, preserving peer connection state and host readiness. A genuinely dead transport still produces no reply and follows the existing watchdog timeout path.

## Validation

Closes #1016

### DIFF
--- a/crates/flotilla-daemon/src/peer/manager.rs
+++ b/crates/flotilla-daemon/src/peer/manager.rs
@@ -717,7 +717,11 @@ impl PeerManager {
                     HandleResult::ReconnectSuppressed { peer: env.connection_peer }
                 }
             },
-            PeerWireMessage::Ping { .. } | PeerWireMessage::Pong { .. } => HandleResult::Ignored,
+            PeerWireMessage::Ping { timestamp } => {
+                self.queue_send_to(&env.connection_peer, PeerWireMessage::Pong { timestamp });
+                HandleResult::Ignored
+            }
+            PeerWireMessage::Pong { .. } => HandleResult::Ignored,
         }
     }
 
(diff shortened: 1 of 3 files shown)


rjwittams:
Evaluated in full. The review found no bugs, security issues, missing coverage, or blocking nits, so no code changes are needed. The generation guard, queued send path, responsive-link test, and silent-link timeout test cover the intended keepalive contract.

PR #1021, opened 2026-07-25
fix: polish project page tables

### LINKED ISSUE #1013: Project-page table polish batch (from the #666 acceptance feedback)
Robert's live feedback using the composite project page, 2026-07-25 (the de facto #666 acceptance pass — "scoped tabs on project level seem to work well"). Five batchable polish items, one crew:

- [ ] **Terminal title**: scoped-view panes set the terminal title to the tab/view name — currently `sh -c $FLOTILLA_BIN …`.
- [ ] **Back affordance**: drilling (e.g. into a convoy) has tab-local history with Esc-pops per the recorded ruling — add a VISIBLE back control (breadcrumb or ← in the view header) so the path out is discoverable, mouse-friendly.
- [ ] **Checkouts table**: port the Plane-A checkouts view's line-dedup handling and home-dir/path shortening (`~/…` + middle-elision) — currently duplicated lines and long raw paths.
- [ ] **Section headers**: standardize and align across tables — timestamps right-aligned consistently; shared header component.
- [ ] **Scroll/fetch-more**: infinite scroll is annoying for non-last tables (issues aren't last) — make fetch-more an actively chosen affordance for non-terminal sections (button/key), keeping auto-extend only where a section is last. Collapsing already mitigates; this makes it deliberate.

Recorded for later, NOT this batch: collapsible + reorderable sections; the section list evolving toward a grafana-style dashboard of heterogeneous widgets (noted on #945).

Parent: #666 (this batch + the default-flip decision are what remain of its acceptance).

rjwittams:
Closes #1013

## What changed

- set scoped-view terminal titles from the active tab/view label, including drill navigation
- add a visible, clickable `← Back` breadcrumb header for tab-local history
- shorten checkout paths against each host home, middle-elide long paths, and suppress repeated host/repository cells
- share section-header rendering and right-align freshness timestamps consistently
- require explicit `f` fetch-more for non-terminal project sections while preserving automatic extension for a terminal section

## Why

The composite project page worked functionally, but its pane identity and drill escape path were hard to discover, checkout rows were noisy, table headers were inconsistent, and automatic issue pagination could trap navigation before later sections.

## Validation

### DIFF
--- a/crates/flotilla-tui/src/app/mod.rs
+++ b/crates/flotilla-tui/src/app/mod.rs
@@ -332,6 +332,7 @@ impl<R: Clone> QueryTableResult<R> {
 }
 
 pub fn table_rows<'a>(
+    model: &'a TuiModel,
     namespaces: &'a NamespaceMap,
     queries: &'a QueryTableCache,
     source_search: Option<&'a str>,
@@ -357,6 +358,11 @@ pub fn table_rows<'a>(
             .awareness
             .iter()
             .map(|(query, result)| crate::table_view::QueryRows { query, rows: &result.rows, state: &result.state })
+            .collect(),
+        host_home_dirs: model
+            .hosts
+            .values()
+            .filter_map(|host| host.summary.system.home_dir.as_deref().map(|home| (host.host_name.clone(), home)))
             .collect(),
         source_search,
     }
--- a/crates/flotilla-tui/src/app/mod.rs
+++ b/crates/flotilla-tui/src/app/mod.rs
@@ -362,7 +362,10 @@ pub fn table_rows<'a>(
         host_home_dirs: model
             .hosts
             .values()
-            .filter_map(|host| host.summary.system.home_dir.as_deref().map(|home| (host.host_name.clone(), home)))
+            .filter_map(|host| {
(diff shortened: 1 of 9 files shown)


rjwittams:
Fixed in 5360fd88.

- Checkout host/repository deduplication now happens at render time against the filtered `TableView`, so the first surviving row retains its grouping labels.
- Home-directory lookup now goes through `resolve_host`; ambiguous duplicate host names are omitted instead of selecting an arbitrary path.
- Reused `shorten_against_home` directly and added regressions for filtering and ambiguous hosts.

The exact formatting, clippy, and workspace test commands all pass locally.

rjwittams:
Fixed in d079d932. Deduplication now restarts at `first_visible`, so the top row of every scrolled window always renders its host and repository labels. Repository grouping also locates the host column by id instead of assuming column zero. Added `checkout_line_dedup_restarts_at_the_top_of_the_visible_window` to cover the reported case.

The exact formatting, clippy, and full workspace test commands pass locally after this change.

rjwittams:
Reviewed. This pass confirms the filtering, scrolling, and ambiguous-host fixes and raises no new actionable issue, so no further code change is needed.

PR #1020, opened 2026-07-25
fix: confirm issue convoy dispatches

rjwittams:
## Summary

- route every issue-table convoy start through one confirmation modal
- show issue details, target project, placement/workflow defaults, and a non-ready warning
- accept optional guidance and pass it through to the convoy instruction/crew brief
- preserve confirmation for fan-out and batch multi-issue dispatches

## Testing

- `cargo +nightly-2026-03-12 fmt --check`
- `cargo clippy --workspace --all-targets --locked -- -D warnings`
- `cargo test --workspace --locked`

Closes #1012

PR #1022, opened 2026-07-25
feat: make convoy auto-attach presence-aware

### LINKED ISSUE #1009: Dispatch stops auto-opening convoy tabs when a presentation surface is connected
Robert, 2026-07-25: with andamento giving decent convoy visibility, dispatch should NOT auto-open convoy tabs in zellij — the latent appearing in the rail is the notification, opening is on-demand. But a plain zellij (or any setup) WITHOUT a presentation surface still wants auto-open as its only visibility.

## What to build

- Auto-attach/auto-open on `convoy start` becomes **presence-aware**: default ON only when NO presentation surface is connected to the daemon (the manifest connector / `pm connect` session is the presence signal — the daemon already knows its connections); default OFF (latent-only) when one is.
- Explicit flags stay decisive both ways: `--attach` forces open, `--no-attach` forces latent-only, regardless of presence.
- Config override for the default (per-repo or global) for setups where the heuristic is wrong.
- Same presence-awareness for the TUI's start-convoy flow (Story D auto-attach).

## Acceptance

- [ ] With a PM connected: `convoy start` (no flags) creates the convoy latent-only; no tab opens
- [ ] Without any PM connection: unchanged auto-open behavior
- [ ] `--attach`/`--no-attach` override presence in both directions
- [ ] Presence check covers the manifest-connector session, not just TUI clients

rjwittams:
Closes #1009

## What changed

- represent convoy-start attachment as a tri-state default/always/never intent
- resolve the default on the dispatching daemon from connected ambient presentation surfaces
- keep `--attach` and `--no-attach` decisive regardless of connector presence
- make the TUI start-convoy flows use the same presence-aware default
- add a global `[convoy] auto_attach = true|false` override and document it

## Why

A connected `flotilla pm connect` session already provides convoy visibility, so opening a multiplexer tab automatically is disruptive. Without that connector, auto-attach remains the only immediate visibility and stays enabled.

## Validation

### DIFF
--- a/crates/flotilla-commands/src/commands/convoy.rs
+++ b/crates/flotilla-commands/src/commands/convoy.rs
@@ -1,7 +1,7 @@
 use std::path::PathBuf;
 
 use clap::{Parser, Subcommand};
-use flotilla_protocol::{Command, CommandAction, ConvoyStartIntent, IssueRef, IssueSelector, IssueSource};
+use flotilla_protocol::{Command, CommandAction, ConvoyAutoAttach, ConvoyStartIntent, IssueRef, IssueSelector, IssueSource};
 
 use crate::{
     quote::quote_value,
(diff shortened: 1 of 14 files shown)


rjwittams:
Thanks for the review. No change needed for the two non-blocking notes: ambient presence is intentionally daemon-wide because the daemon currently serves one provisioning namespace, and issue #1009 permits either a per-repo or global override, so this implementation deliberately chooses the documented global `[convoy]` setting.

rjwittams:
Thanks for the post-merge verification. No changes requested or needed; the conflict resolution intentionally preserves #1020's dispatch confirmation and instruction propagation alongside the presence-aware attach modes.

PR #1024, opened 2026-07-25
Publish presentation entities as flat facts

### LINKED ISSUE #982: Enactment: entities-only wire + andamento grouping engine (atomic cutover, one bump window)
Enactment umbrella for the #950 ruling (resolution comment there is the contract; map #969 is the index). Gated on #970 (template model core) for the remaining KDL schema detail, then dispatchable as coordinated work:

- **flotilla side**: connector emits entity patches (canonical ids per kind via shared flotilla-manifest constructors; flat facts incl. activatability/recipe, source provenance; per-key fold per the declared pipe contract). `MetadataTarget::Group` deleted from the wire.
- **andamento side**: grouping engine derives the tree from entities via named declarative KDL templates (ordered levels, per-level label-key/collapse-single/show-empty, filter clause; kind→presence-class mapping; truthful nearest-ancestor filing). Runtime template switching via the #971 variables mechanism when it lands (a fixed default template is acceptable for the cutover PR).
- **Atomic**: both PRs merge in one bump window, one daily-driver relaunch. Neither merges alone.
- ADR carry: ADR 0018 amendment (entities-only wire) rides the flotilla PR.

Acceptance = #950's checklist: template switch re-groups live with no producer change; spine is config not code; issues never tabs, single-member groups collapse per policy; dual presence unrepresentable (also closes #964's dedup question structurally — one convoy entity, one vessel entity, collapse policy renders one tab).

Blocked by: #970.

rjwittams:
## Summary

- replace presentation `Group` wire targets with canonical project, repo, convoy, vessel, issue, session, and checkout entities
- publish flat hierarchy, status, provenance, and primary-action facts with per-key fold semantics
- stamp panes and tabs with canonical entity identity and focus/materialization targets
- amend ADR 0018 for the entities-only transport contract and regenerate the Andamento-owned serde fixtures

## Why

Andamento must own hierarchy and presence policy. Producers now describe semantic entities once; named presentation templates can regroup those same facts without restamping or creating dual presence.

## Validation

Closes flotilla-org/flotilla#982

## Atomic cutover

Companion Andamento PR: https://github.com/flotilla-org/andamento/pull/35

**Do not merge this PR without https://github.com/flotilla-org/andamento/pull/35. Both halves must land in one bump window and be followed by one daily-driver relaunch.**


rjwittams:
Fixed in 6548f867. Repository-only awareness nodes now derive parent facts through the same canonical entity resolver as node publication, so a `repo/*` node cannot fabricate a Project entity or stamp fake project ancestry. I also restored an entities-only regression covering the repo and independent child facts while asserting that no Project entity is emitted.

I checked the vessel note as well: `AttachBinding::resource_ref` already defines vessel identity as convoy-qualified and treats a vessel without a convoy as non-identity data, so retaining a raw vessel string would violate the canonical entity-ID contract introduced by this cutover.

rjwittams:
Confirmed and retaining this intentionally. Although Flotilla no longer constructs `MetadataValue::GroupPath` during entity publication, the variant remains part of the cross-repository wire schema and must stay serde-compatible with Andamento. The paired consumer still uses `GroupPath` extensively for its derived internal presentation model, so keeping the public conversion helper alongside the wire type avoids asymmetric schema tooling during this atomic cutover.

PR #1025, opened 2026-07-25
fix: surface codex workspace trust prompts

### LINKED ISSUE #1002: Multi-repo workspace roots aren't pre-trusted — codex crews stall silently at the directory trust prompt
Live specimen 2026-07-25: the first multi-repo convoy (entities cutover on project presentation) showed crew phase Working for hours while the codex session sat at "Do you trust the contents of this directory?" — the multi-repo workspace parent (`~/dev/flotilla-repos/<convoy>/<project>-<convoy>/`) is a new directory shape outside codex's accumulated trust, unlike single-repo checkout paths. Unstuck manually with one Enter.

Two defects:

1. **Provisioning**: vessel provisioning should pre-seed trust for the workspace root before launching the agent CLI (codex trust config entry, or launch flag if one exists) — for ALL layouts, so new workspace shapes never regress this.
2. **Observability**: a crew parked at an interactive prompt for hours reported `Working`. The attention plane (ADR 0017: NeedsInput) exists for exactly this — the adapter/session watcher should classify a codex trust/approval prompt as NeedsInput so salience goes to attention and the governor/principal sees it instead of a healthy-looking phase. (The cleat screen had the answer all along — the signal just never left the pane.)

## Acceptance

- [ ] A fresh multi-repo convoy's crew reaches its brief without human input
- [ ] A crew waiting at any interactive prompt surfaces NeedsInput/attention within the watch latency, not Working

rjwittams:
## Summary

- pre-seed the canonical workspace path as trusted in Codex config before launching crew agents
- capture stable cleat screens and classify current Codex trust, approval, permission, edit, network, and question prompts as NeedsInput
- preserve existing Codex config formatting/comments and cover live attention observation with regression tests

## Testing

Closes #1002

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1103,6 +1103,7 @@ dependencies = [
  "tokio",
  "tokio-util",
  "toml",
+ "toml_edit 0.22.27",
  "tracing",
  "urlencoding",
  "uuid",
--- a/crates/flotilla-core/Cargo.toml
+++ b/crates/flotilla-core/Cargo.toml
@@ -24,6 +24,7 @@ indexmap = { workspace = true }
 color-eyre = { workspace = true }
 dirs = "6"
 toml = "0.8"
+toml_edit = "0.22"
 reqwest = { workspace = true, features = ["charset", "http2", "json", "system-proxy"] }
 http = "1"
 bytes = "1"
--- a/crates/flotilla-core/src/agent_adapter.rs
+++ b/crates/flotilla-core/src/agent_adapter.rs
@@ -6,8 +6,10 @@ use std::{
 
 use async_trait::async_trait;
 use flotilla_protocol::arg::{flatten, Arg};
-use flotilla_resources::TerminalBrief;
+use flotilla_resources::{TerminalAttentionState, TerminalBrief};
 use serde::Serialize;
+use tokio::sync::Mutex;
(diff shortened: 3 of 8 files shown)


rjwittams:
Addressed in 28102302. Default host discovery now publishes both HOME and CODEX_HOME into the environment bag, and the production-shaped detector regression now constructs the adapter from that real bag and calls prepare() successfully. Canonical workspace resolution also fails explicitly when pwd -P fails or returns an empty path instead of silently writing a potentially ineffective trust key. The shared discovery mock now models pwd -P for crew fixtures. The config update remains an atomic write serialized across Flotilla launches sharing the adapter; cross-process advisory locking against manual/user Codex edits is intentionally outside this fix. Final fmt, clippy -D warnings, and the full locked workspace suite are green.

rjwittams:
Thanks for the follow-up. I agree with both observations and am keeping the current behavior intentionally: capability/adapter lookup failures indicate a structurally invalid or unavailable terminal spec and should remain visible to reconciliation, while transient screen-capture failure alone degrades to activity-based attention. The substring classifier is deliberately limited to stable Codex agent screens and current upstream prompt vocabulary; if real false positives appear, tightening it with additional prompt structure belongs in a focused follow-up. No further code change needed for these non-blocking notes.

PR #1027, opened 2026-07-25
fix: emit empty projects in awareness

### LINKED ISSUE #1000: Awareness band omits projects with no children — empty projects should emit, templates decide visibility
Robert, 2026-07-25, at the restarted rail: only projects with children (a convoy, ready issues) appear; the other seven tracked projects are invisible.

ADR 0018's awareness band explicitly includes "projects and repos with no convoy activity" — browse-scope is everything a principal COULD regard — and the #921 open-semantics ruling gives project latents a real affordance (open the scoped TUI, minted by #953). The control point is also already ruled: `show-empty` is a per-level grouping-template option (template model core resolution on #970).

## What to build

- The aggregator's awareness projection emits a project node for every Project resource in scope, children or none (state: idle; counts: zeroes; scoped-view recipe as today for project nodes).
- Surface-side visibility is the template's call: the bundled default template can hide empties initially or show them dimmed — but the DATA must be there so it is a template toggle, not a producer change (the #950 principle).
- Windowing/limits per ADR 0014 still apply — this is about eligibility, not unbounded emission.

## Acceptance

- [ ] All tracked projects appear in awareness output (verifiable via the daemon query, independent of rail rendering)
- [ ] An empty project's latent opens its scoped TUI
- [ ] Rail behavior controlled by the grouping template's show-empty (or equivalent existing toggle), not hardcoded

rjwittams:
## What

- seed project-grouped awareness output from the complete Project catalog, so projects remain eligible even when they have no convoys, ready issues, checkouts, or independent sessions
- represent empty project nodes with explicit `idle` state, zero counts, their Project resource ref, and the existing scoped-view action recipe
- retain awareness group windowing, ranking active projects ahead of empty projects when the group limit truncates output
- advance awareness sequence numbers when the Project catalog changes
- canonicalize unqualified convoy project refs with the convoy namespace so child rows join the catalogued Project node

## Why

The aggregator previously created awareness groups only while processing child rows. Tracked Projects with no children therefore disappeared before presentation templates could apply their `show-empty` policy. The producer now emits the complete eligible data set, while surface grouping templates remain responsible for visibility.

## Impact

Daemon awareness queries include all catalogued Projects within the configured awareness limit. Empty Projects are idle latents and open their scoped TUI with `flotilla view project/<namespace>/<name>`.

## Validation

Closes #1000

### DIFF
--- a/crates/flotilla-core/src/aggregator_projection.rs
+++ b/crates/flotilla-core/src/aggregator_projection.rs
@@ -116,6 +116,12 @@ struct SalienceProjection {
     revision: u64,
 }
 
+#[derive(Debug, Default)]
+struct ProjectCatalogProjection {
+    projects: HashMap<QueryScope, Vec<RepositoryKey>>,
+    revision: u64,
+}
+
 #[derive(Debug, Default, Clone, bon::Builder)]
 pub struct AggregatorProjectionState {
     convoys: Arc<RwLock<QueryProjection<ConvoyRow>>>,
@@ -125,6 +131,8 @@ pub struct AggregatorProjectionState {
     checkouts: Arc<RwLock<ScopedCheckoutProjection>>,
     #[builder(skip)]
     salience: Arc<RwLock<SalienceProjection>>,
+    #[builder(skip)]
+    project_catalog: Arc<RwLock<ProjectCatalogProjection>>,
     /// Subscriber ownership and demand-backed materializations belong to the
     /// Aggregator state, shared with the daemon's subscription transport.
     demand_backed: QueryRegistry,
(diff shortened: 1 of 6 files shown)


rjwittams:
Reviewed the two non-blocking observations; no code change is needed. The inner scope filter keeps `project_awareness` correct as a pure public projection even when called directly, while the outer filter avoids carrying unrelated catalog entries across the async aggregation seam. Project refs are intentionally included before child refs so an empty node and its later non-empty form retain the authoritative Project identity; downstream identity remains based on `scope`/`id`, with no refs-order contract.

PR #1043, opened 2026-07-25
fix: inhibit host sleep while convoys are active

### LINKED ISSUE #1034: Placement hosts sleep mid-crew — daemon must inhibit system sleep while vessels are active
Live, twice today (2026-07-25): feta suspended ~1 minute after a convoy dispatch — a codex crew awaiting an API response presents as an idle system to sleep-on-idle policy. The suspend killed the crew's in-flight turn (session survived via cleat; the turn's work was lost; governor nudged it manually after WOL), froze heartbeats (host NOT-READY), and broke dispatch until a manual wake.

## What to build

- flotillad takes a **sleep inhibitor while any vessel/crew on the host is active** (Linux: `systemd-inhibit`/logind inhibitor fd; macOS: IOPMAssertion via `caffeinate`-equivalent) — released when the last crew completes, so idle hosts still sleep and WOL economics stay.
- On daemon start with active convoys (recovery), re-acquire the inhibitor.
- Wake-on-dispatch remains the front half (governor already WOLs manually; a dispatch-path auto-WOL using the fleet MAC registry is an optional stretch — note it, don't require it).

## Acceptance

- [ ] Host with an Active convoy does not auto-suspend (inhibitor visible in `systemd-inhibit --list`)
- [ ] Inhibitor released after last crew completes; host sleeps again on its normal policy
- [ ] Daemon restart mid-convoy re-acquires

rjwittams:
## What changed

- watch local durable convoy lifecycle state from `flotillad`
- hold a blocking logind sleep inhibitor on Linux with `systemd-inhibit`, and an idle-sleep IOPM assertion on macOS with `caffeinate`
- release inhibition only after the last non-terminal convoy completes, fails, is cancelled, is abandoned, or is deleted
- recover desired inhibitor state from the durable convoy list before watching changes, so daemon restart mid-convoy reacquires it
- periodically verify the helper process and reacquire if it exits unexpectedly

## Why

Placement hosts can appear idle while crew agents wait on external API responses. Their normal sleep-on-idle policy can then suspend the host, freeze heartbeats, and destroy the in-flight agent turn. The daemon previously did not connect durable active-convoy state to the host OS sleep policy.

## Impact

Hosts remain awake only while local vessel work exists. Once the final convoy reaches a terminal state, the inhibitor process exits and the host's normal sleep/WOL economics resume. Wake-on-dispatch remains a separate optional follow-up.

## Validation

Closes #1034

### DIFF
--- a/crates/flotilla-daemon/src/lib.rs
+++ b/crates/flotilla-daemon/src/lib.rs
@@ -1,5 +1,6 @@
 mod aggregator;
 mod issue_materializer;
+mod sleep_inhibitor;
 pub use aggregator::{Aggregator, AggregatorResolvers};
 
 pub mod cli;
--- a/crates/flotilla-daemon/src/runtime.rs
+++ b/crates/flotilla-daemon/src/runtime.rs
@@ -44,6 +44,7 @@ use tokio::{sync::Mutex, task::JoinHandle};
 use tracing::warn;
 
 use crate::{
+    sleep_inhibitor,
     supervisor::{supervise, ControllerSupervision},
     Aggregator, AggregatorResolvers,
 };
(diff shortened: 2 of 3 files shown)


rjwittams:
Addressed all three notes in 3ddb95a7. The Linux `--why` argument and acquire log now share `INHIBITOR_REASON`; Linux runs a `tail --pid=<daemon-pid>` payload and macOS uses `caffeinate -w <daemon-pid>`, so the OS inhibitor exits even when daemon drop glue cannot run; and repeated identical acquisition errors are logged once while the 30-second retry continues. The exact format, workspace clippy, and workspace test gates pass locally after the change.

PR #1053, opened 2026-07-25
fix: cover capped peer redial recovery

### LINKED ISSUE #1045: Peer redial backoff grows unbounded — a long host sleep permanently orphans the link until a daemon bounce
Fourth member of the reconnect family (#992 generations, #1008 socket re-resolution, #1016 keepalive pings — all landed). Remaining hole, hit twice today: when a peer host sleeps for long (feta's overnight + evening naps), the dialing side's reconnect backoff grows without cap; after wake, NO dial attempts appear in the log at all, and only a daemon bounce restores the link.

Fix: cap the redial backoff (≤60s) so recovery after any outage is bounded; optionally reset backoff on WOL-adjacent signals later, but the cap alone closes the operational hole.

Acceptance:
- [ ] Peer asleep 30+ min, then woken: link re-establishes within ≤2 backoff intervals, no bounce
- [ ] Compose-harness regression (extends #1029's suite: long transport outage then recovery)

rjwittams:
## Summary

- extend the compose wedge-regression suite with a prolonged peer transport outage
- assert every observed redial delay stays at or below 60 seconds and reaches the capped interval
- wake the peer without bouncing the dialing daemon and verify reconnection within two capped intervals

## Root cause

The SSH redial path already computes a capped 60-second exponential delay, but the compose suite only covered short transport failures. There was no real-boundary regression proving that the runtime continues dialing after a long outage and recovers when the peer returns.

## Validation

- `uv run --with pytest --with pytest-timeout pytest -v tests/integration/test_wedge_regressions.py::test_04_long_transport_outage_recovers_with_capped_backoff`

Closes #1045

### DIFF
--- a/tests/integration/test_wedge_regressions.py
+++ b/tests/integration/test_wedge_regressions.py
@@ -4,6 +4,8 @@ import json
 import re
 import time
 
+import pytest
+
 from conftest import (
     compose,
     daemon_log,
@@ -15,6 +17,7 @@ from conftest import (
 )
 
 GENERATION_RE = re.compile(r"\bgeneration=(\d+)\b")
+BACKOFF_RE = re.compile(r"\battempt=(\d+)\b.*\bdelay_secs=(\d+)\b")
 RECONNECT_MESSAGES = (
     "SSH connection dropped, will reconnect",
     "reconnecting after backoff",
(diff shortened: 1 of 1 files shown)


rjwittams:
Fixed both timing concerns in 29e7a5d0. The test now has a 400-second hard ceiling, leaving slack beyond its internal wait budgets, and the reconnect predicate checks elapsed time from `woken_at` on every poll so the two-interval deadline is load-bearing rather than a post-hoc assertion. The targeted compose regression and all three local CI gates pass after the change.

rjwittams:
Thanks for the re-review. I also cleaned up the non-blocking blank-line nit in a8c8cef2; test collection and the exact format, clippy, and workspace test gates remain green.

PR #1074, opened 2026-07-26
fix: isolate tab clicks from page mouse handling

### LINKED ISSUE #1068: tui: tab-bar click must not leak into the newly activated page
## What happened (corrected diagnosis)

Clicking a tab in the TUI's own tab bar switches tabs — and the same mouse-down then reaches the newly activated page widget (double-dispatch within the event cycle, or dispatch against a stale area/scroll mapping from before the switch). On a project page the first panel's header line is where a top-row click lands, and the first panel is Convoys — whose header-click handler drills instantly. Net effect: clicking the cleat TUI tab drills into the (unfiltered, #1069) convoys view with a context-less title (#1070), with zero intentional input.

## What to build

Diagnose precisely with a failing widget-level test (mouse event → SwitchToTab → assert no further state change from that event), then fix so a mouse event consumed by the tab bar is never also delivered to page widgets, and page mouse handling never runs against a stale layout. Additionally: right-click must never drill anywhere (the affordance contract reserves right-click for the action menu).

Note: the wider row-interaction contract (when rows drill vs act vs menu) is being ruled separately on the presentation map; this ticket is the event-leak fix + right-click-drill removal, both unambiguous now.

## Acceptance criteria

- [ ] Clicking a TUI tab changes the active tab and nothing else — locked by a regression test
- [ ] Section headers on the freshly activated page never receive the switching click
- [ ] Right-click never drills; covered by test

rjwittams:
## What changed

- invalidate page mouse hit-testing whenever the active View changes, and re-enable it only after that View renders its own layout
- keep tab-bar mouse-down handling isolated as a queued `SwitchToTab` action that is consumed before page dispatch
- prevent right-clicks on project section headers from drilling; right-click remains reserved for action menus
- add regressions for tab-switch isolation, stale project-page layouts, and right-clicked section headers

## Root cause

Project page widgets retain their last rendered hit-test area. A View can become active while input events are already queued and before the next frame renders, allowing page mouse handling to use stale coordinates. On a project page, the stale top content row maps to the Convoys section header and drills immediately.

The screen now records which View produced the current page layout. Active-view synchronization invalidates that record, so shell controls such as the tab bar remain usable while page input is consumed until the newly active View renders.

## Impact

Clicking a TUI tab changes only the active tab. The switching click cannot reach the newly activated page, section headers cannot receive input against an old layout, and right-click never drills through a project header.

## Validation

- `cargo test -p flotilla-tui --locked` (1,007 unit, 2 e2e, 1 high-fidelity, and 50 snapshot tests)

The exact workspace test command also ran; it reached an unrelated pre-existing `flotilla-core` failure in `in_process::tests::adopted_checkout_with_explicit_transport_applies_fork_stance_config`, which fails identically in isolation and is outside the unchanged TUI scope.

Closes #1068

### DIFF
--- a/crates/flotilla-tui/src/app/key_handlers/tests.rs
+++ b/crates/flotilla-tui/src/app/key_handlers/tests.rs
@@ -331,6 +331,25 @@ fn left_click(x: u16, y: u16) -> MouseEvent {
     MouseEvent { kind: MouseEventKind::Down(MouseButton::Left), column: x, row: y, modifiers: KeyModifiers::NONE }
 }
 
+fn render_screen(app: &mut App) {
+    let mut terminal = ratatui::Terminal::new(ratatui::backend::TestBackend::new(80, 30)).expect("terminal");
+    terminal
+        .draw(|frame| {
+            let mut ctx = crate::widgets::RenderContext {
+                model: &app.model,
+                views: &mut app.views,
+                ui: &mut app.ui,
+                theme: &app.theme,
+                keymap: &app.keymap,
+                in_flight: &app.in_flight,
+                namespaces: &app.namespaces,
+                query_tables: &app.query_tables,
+            };
+            app.screen.render(frame, frame.area(), &mut ctx);
+        })
+        .expect("draw");
+}
+
 // ── handle_key — top-level dispatch ──────────────────────────────
 
 #[test]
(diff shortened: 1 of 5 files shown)


rjwittams:
Verified; no code change needed. `OpenViews` enforces unique parsed View addresses, and duplicate broken targets dispatch to `ActivePage::Error`, whose mouse path is inert, so the rendered-target comparison cannot expose a stale interactive page through that edge case. The two small render helpers remain local to separate test modules for now; consolidating them would be unrelated test-harness cleanup and the review correctly marks it non-blocking.

PR #1076, opened 2026-07-26
tui: project-scope the convoys view address

### LINKED ISSUE #1069: tui: project-scope the convoys view address
## What happened

The project page's Convoys panel targets `convoys/flotilla` — `ViewAddress::Convoys { namespace }` has no project field in the protocol, so drilling the Convoys section from e.g. the cleat project page lands on the namespace-global convoys view, unlike Issues/Checkouts which carry `?project=`.

## What to build

Add an optional project scope to `ViewAddress::Convoys` (no-compat phase — change the address format freely), make the convoys view/query filter by it, and have the project page pass its project scope into the panel target. The unscoped global convoys view remains available as the scope-less address.

## Acceptance criteria

- [ ] Drilling Convoys from a project page shows only that project's convoys
- [ ] Address round-trips through open-views.toml persistence
- [ ] Global convoys view unchanged when unscoped

rjwittams:
## Summary

- add optional Project scope to Convoys view addresses and query identities
- filter scoped Convoys result sets in the Aggregator while preserving the global set
- make Project panels subscribe to and drill into their scoped Convoys view
- keep scoped and global Convoys caches isolated in the TUI

## Tests

- `TMPDIR="$PWD/.codex-tmp" cargo test --workspace --locked`

Closes #1069

### DIFF
--- a/crates/flotilla-client/src/lib/tests.rs
+++ b/crates/flotilla-client/src/lib/tests.rs
@@ -73,7 +73,7 @@ fn no_subscriptions() -> Arc<QuerySet> {
 }
 
 fn convoys_subscription() -> Arc<QuerySet> {
-    Arc::new(std::sync::RwLock::new(HashSet::from([QueryId::Convoys])))
+    Arc::new(std::sync::RwLock::new(HashSet::from([QueryId::Convoys { scope: None }])))
 }
 
 fn request_harness() -> RequestHarness {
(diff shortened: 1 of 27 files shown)


rjwittams:
Thanks for the review. I verified each point:

- The full scoped snapshot fan-out is correct but can do avoidable work for unrelated project changes. I kept the smallest correct delivery model in this PR and filed #1077 to measure and implement snapshot deduplication or scoped deltas with regression coverage.
- The table projection first filters by the address namespace, then by the scope's project reference; the server result is also selected by the exact scoped query. A hand-written address whose namespace disagrees with the scope therefore renders no rows rather than crossing scopes, so no correctness change is needed here.
- The `project_panels` fallback is retained for symmetry with the other composite panels and to keep the composition resilient if the Convoys projection gains a fallible lookup later.
- `QueryChanges::Convoys::scope` is intentionally `None` while global subscribers use deltas and scoped subscribers use snapshots; #1077 is the place to revisit scoped deltas.

No code changes were required from this review cycle.

PR #1084, opened 2026-07-26
fix: tolerate drifted replica snapshot records

### LINKED ISSUE #1079: replica: one unparseable record fails the whole snapshot
## What happened

udder's replica status on kiwi: `replica snapshot parse failed: missing field repo_label at line 43 column 13` — no sync at all. The drift itself is expected (no-compat phase; udder's daemon predates a schema change and gets bumped at the next fleet pass), but a single unparseable record rejecting the **entire** replica snapshot turns any schema skew into total replication blindness, with the error only visible in `flotilla ls` output.

## What to build

Per-record tolerance in replica snapshot ingestion: skip-and-count unparseable records, sync the rest, surface the skew clearly (count + first error + generation) in replica status. A fleet running mixed builds mid-rollout should degrade, not blank.

## Acceptance criteria

- [ ] A snapshot with one bad record syncs the remaining records
- [ ] Replica status shows skipped-count and first parse error
- [ ] Test with a deliberately drifted record

rjwittams:
## Summary

- decode fleet replica snapshots record by record, including rows nested inside typed result sets
- retain valid records while counting skipped records and preserving the first path-qualified parse error
- expose parse skew in replica status without marking a partially synced replica unreachable
- render skipped count, first parse error, and generation in `flotilla ls`

## Root cause

Replica snapshot ingestion deserialized the entire JSON document directly into `FleetReplicaSnapshot`. A schema mismatch in any typed row therefore failed the whole document and left the remote host completely absent from the replica cache.

## Impact

Mixed-build fleets now degrade during rollouts: incompatible records are omitted, compatible records continue syncing, and operators see the exact skew in replica status.

## Validation

- `cargo test -p flotilla-core --lib replica_refresh_ --locked --features test-support`
- `cargo test -p flotilla-tui --lib command_result_human::fleet_list --locked`
- `cargo test --workspace --locked` progressed through the workspace but is locally blocked by the unrelated existing `in_process::tests::adopted_checkout_with_explicit_transport_applies_fork_stance_config` assertion

Closes #1079

### DIFF
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -661,7 +661,11 @@ struct ReplicaParseDiagnostics {
 
 impl ReplicaParseDiagnostics {
     fn record_skip(&mut self, path: impl std::fmt::Display, error: impl std::fmt::Display) {
-        self.skipped_records += 1;
+        self.record_skips(1, path, error);
+    }
+
+    fn record_skips(&mut self, count: usize, path: impl std::fmt::Display, error: impl std::fmt::Display) {
+        self.skipped_records += count;
         self.first_error.get_or_insert_with(|| format!("{path}: {error}"));
     }
 }
@@ -675,6 +679,8 @@ struct ParsedFleetReplicaSnapshot {
 fn result_set_records_mut(result_set: &mut serde_json::Value) -> Option<&mut Vec<serde_json::Value>> {
     let content = result_set.get_mut("rows")?.get_mut("rows")?;
     if content.is_array() {
+        // Tuple-style variants such as Rows::Convoys serialize their content
+        // directly as the record array.
         content.as_array_mut()
     } else {
         content.get_mut("rows")?.as_array_mut()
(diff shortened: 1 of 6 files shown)


rjwittams:
Fixed in 7ca2de21. Envelope failures now add every known row in the result set to `skipped_records` (with a minimum of one for an unreadable empty envelope), and the end-to-end refresh test now drifts `scope` on a two-row checkout set and verifies both rows are counted.

I also documented that the bare-array branch is used by the tuple-style `Rows::Convoys` variant. The per-row deserialization remains intentionally linear in row count and keeps validation tied to the canonical `ResultSet` decoder; current periodic snapshot sizes make that tradeoff acceptable.

rjwittams:
Added the requested flat-row coverage in 89c88aa6. The same end-to-end refresh test now removes required `namespace` from one of two top-level `FleetListRow` values, verifies the valid sibling remains cached, and checks the path-qualified status diagnostic.

The enum-shape claim does not match this branch: `crates/flotilla-protocol/src/result_set.rs` defines `Rows::Convoys(Vec<ConvoyRow>)` at line 315, while the other variants are struct variants. Under the adjacent tag, Convoys therefore does serialize its content as the bare record array. The `content.is_array()` branch and its comment are both live and retained.

rjwittams:
You were right about the code that will ship. My checkout was six commits behind current `main`, where `Rows::Convoys` was still tuple-shaped; GitHub's merge result already contained the newer struct-shaped variant. I merged `origin/main` and removed the now-dead array branch/comment in e94f63d5. The focused drift test passes against the updated base.

PR #1085, opened 2026-07-26
fix: show configured hosts during redial

### LINKED ISSUE #1073: host list: configured host mid-redial vanishes instead of showing reconnecting
## What happened

While feta was asleep and kiwi's daemon was in capped redial backoff (attempt 592, 60s cap — the #1045 fix working as designed), `flotilla host list` showed **no row for feta at all** — not even configured/disconnected — while udder rendered as configured+connected. An operator reading the list concludes the host isn't configured, when actually the daemon knows it, is dialing it every 60s, and hosts.toml lists it.

Observed 2026-07-26 during the everything-placed-on-kiwi confusion: placement had (correctly) fallen back local, but the host list hid *why*.

## What to build

Configured hosts always render a row; status shows the live link state (connected / reconnecting (attempt N, next dial in Ns) / disconnected). The redial state is already tracked by peer_runtime — surface it.

## Acceptance criteria

- [ ] A configured host in redial backoff shows as its own row with reconnecting status
- [ ] Covered by a test at the host-list projection level

rjwittams:
## Summary

- project every configured peer target into `host list`, even before it has a canonical node or environment identity
- surface the peer runtime's current retry attempt and deadline as `reconnecting (attempt N, next dial in Ns)`
- keep canonical host snapshots unchanged and make config-only row identities explicitly optional
- add host-list projection and human-format coverage

## Root cause

`host list` was built only from canonical host snapshots. When a peer disconnected, its summary and canonical host row were removed even though the daemon's peer runtime still owned the configured target and was actively redialing it.

## Validation

- `cargo test -p flotilla-daemon --locked host_list_projection_includes_configured_target_in_redial_backoff`
- `cargo test -p flotilla-tui --locked host_list_shows_hosts_and_counts`
- `cargo check --workspace --all-targets --locked`
- `cargo test --workspace --locked` runs the changed coverage successfully, but the local macOS run has one unrelated reproducible failure in `in_process::tests::adopted_checkout_with_explicit_transport_applies_fork_stance_config`

Closes #1073

### DIFF
--- a/crates/flotilla-core/src/host_registry.rs
+++ b/crates/flotilla-core/src/host_registry.rs
@@ -105,7 +105,7 @@ impl HostRegistry {
         host_entries.sort_by(|a, b| {
             b.is_local
                 .cmp(&a.is_local)
-                .then_with(|| a.node.node_id.cmp(&b.node.node_id))
+                .then_with(|| a.node.as_ref().map(|node| &node.node_id).cmp(&b.node.as_ref().map(|node| &node.node_id)))
                 .then_with(|| a.environment_id.cmp(&b.environment_id))
         });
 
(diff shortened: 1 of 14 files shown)


rjwittams:
Fixed both consistency gaps in `b41cc083`.

- `project_host_list` now enriches every canonical environment row matching the configured node, with a regression test covering two environments and asserting no duplicate config-only row is added.
- local and forwarded query paths now share `execute_projected_query`, so every `QueryHostList` result produced by this daemon receives the configured-target projection.

The focused projection tests and daemon-wide clippy pass locally.

rjwittams:
Addressed the optional race in `3143855a`: the runtime now records the attempt/deadline before publishing `PeerConnectionState::Reconnecting`, so a concurrent host-list projection cannot transiently downgrade that state to disconnected. Format and daemon-wide clippy pass locally.

rjwittams:
Thanks for the final pass and confirmation. No further changes needed; all raised findings are addressed in the current head.

Issue #1090, opened 2026-07-26
Observed vs authored resources: does host mastering fit fleet-level intent?


rjwittams:
## Question

Host mastering (ADR 0016) assumes each host is the natural authority for what it reports — replicas are read-only copies, and conflicts are impossible because no two hosts describe the same thing. That fits **observed** resources exactly: checkouts, sessions, vessels, host status.

It fits **authored** resources poorly: Project, PlacementPolicy, WorkflowTemplate, annotations (#1089), and portfolio-level artifacts like the DAG board. These are fleet-level *intent*, not observations. Today they are mastered by whichever daemon the operator happened to type into — an accident of CLI invocation, not a design. Symptom: the DAG board is a **multi-repo portfolio artifact** spanning four trackers (GitHub × flotilla/andamento, the zellij fork, lab Forgejo) and belongs to no host at all; "just make it a resource" doesn't obviously work.

Rule:
- Is "authored vs observed" a real class distinction in the resource model, and if so what marks it?
- What is the mastering story for authored resources — designated fleet authority, git/object-store backing (#748), CRDT-ish merge, or the log model already sketched for replication?
- Does a portfolio-scoped artifact (spanning projects, repos, and *trackers*) fit the resource model at all, or is it a view over federated stores (→ #1035's federation direction)?
- Consequences for replica semantics: authored resources presumably must be writable from more than one host, or explicitly not.

Companion to #1089 (annotations are the extreme authored case) and dependent on #1039 (stable ids). HITL grill; likely amends ADR 0016.

rjwittams:
## Reframed after reading ADR 0016 and the code (2026-07-26)

**This ticket's premise was substantially wrong and I should correct it.** ADR 0016 (via the #688 scenario-first grill) already rules exactly what the ticket claimed was missing: **three resource classes** — convergent facts (natural key, union), **definitions** (human-named, *edit-anywhere-including-offline*, conflicts surfaced never auto-resolved), and home-bound runtime (reconciled only at home) — with a masterless overlay model: no root ever writes another root's log, durable materialized replicas per origin, relay, and a deterministic per-field merge where roots agree "by confluence, not consensus."

So "authored" is not a missing class. It is the **definitions** class, already ruled.

[...]

Issue #1039, opened 2026-07-25
Identity & lifecycle interface: rename-stable UIDs, canonical-key migration


rjwittams:
Part of #1035

## Question

Rule the identity substrate: what survives rename (the #999 recreate minted a NEW resource for the same slug — is continuity wanted, and keyed how?); canonical repository identity migration semantics (the #1015 root-cause area: local key → remote key, done safely); and the INTERFACE #979's extraction/rename transitions need from Project/Repository (provenance edges, continuity guarantees) — interface only, mechanics stay #979's.

rjwittams:
**Raised stakes (2026-07-26):** this ticket has become the keystone for two new directions, not just Project-internal tidiness. The annotation layer (#1089) targets foreign entities that may be renamed or moved — an annotation whose subject id is unstable is worse than no annotation. And the authored-vs-observed mastering question (#1090) needs ids that survive being written on one host and read on another. Worth grilling with those consumers in view rather than closing narrowly on Project rename-stability.

rjwittams:
## Resolution (grilled with Robert, 2026-07-26)

**1. No surrogate UIDs. Anywhere.** An opaque immutable id was proposed and **rejected**: for it to earn anything, consumers would have to stamp, store, and re-check it, and today none would — an unread id is machinery pretending to be safety. Reasoning that killed it: enumerate the resources we actually mint and rename essentially never happens. Convoys and vessels live minutes to days (recreate is meaningless); placement policies and workflow templates are configs referenced by name (renaming *should* mean editing references); repositories take identity from the canonical remote, which the forge itself maintains (GitHub's moved-records handle org moves). **Name is identity for our resources; foreign entities (GitHub/Forgejo issues and PRs) keep their own canonical ids and must never get surrogates** — a parallel id is a second truth that can drift.

**2. Extraction needs no identity machinery.** Correcting a repeated conflation: of the three live fixtures, **only ui-scratch→wheelhouse is a rename**; porthole→jackstay and katzensteg→luchs are **extractions**, and extraction happens at **repo level**, not top-level project. An extracted repo has a new canonical remote, so it is a genuinely new Repository — nothing to carry, no rename to model. What it wants is a **provenance edge** ("extracted from") for history/licensing: an annotation-shaped fact, not an identity mechanism. This is the interface #979's transitions need — provenance edges, not continuity guarantees.

[...]

Issue #1037, opened 2026-07-25
Overlap & primacy: shared repos, dual listings, one truth per question


rjwittams:
Part of #1035

## Question

When repos belong to multiple projects: which project claims what? Rule: primacy as explicit data vs derived (#1015 made 'primary whole-repo project' load-bearing implicitly); issue dedup across overlapping projects (the #982-in-both specimen: primary-only, show-in-both, or per-view); grouping claims (which project a watcher tab files under when two could take it); and admission implications (dispatch --project picks the lens — what else does it pick?).

rjwittams:
## Resolution (grilled with Robert, 2026-07-26)

**1. Primacy is explicit data.** A `primary` marker on exactly one member claim per repository, validated fleet-side (second primary rejected at admission/reconcile), never derived from name-matching or claim order. Primary gates the boring ownership questions only: exactly-one-controller of the Repository resource (lifecycle, shared infra) and the default catch-all home for unclaimed ambient entities (#1036). It is **not** exclusive dispatch — any member claim dispatches. Deliberately low-stakes: cheap insurance against the #1015 class, nothing more.

**2. Appearance is unified: association-with-context is the single mechanism, always query-governed.** An ambient-vs-owned type split (issues query-shaped, convoys provenance-shaped) was considered and **rejected**: issues *become* convoys and stay about the same thing; the appearance regime must not change mid-lifecycle. What differs is only how association is established:
- **Issues**: derived — labels/filters/triage per the #1036 sketch, catch-all for the unmatched; usually singular, occasionally plural.
- **Convoys**: inherited — dispatching from a context creates the convoy in that context, association carried as a plain fact (batched issues: the context you stood in when batching).
- Every project view is the same query shape — "entities associated with me" — so a foreign project's convoy never appears in a co-claimant's lens merely for sharing the repo. "Everything touching my repo regardless of context" is available as an explicit query, never a default.
- **Continuity constraint**: the convoy lands in the lens it was dispatched from, and its issues' derived association should agree — a mismatch is a triage smell, plausibly surfaced.

**3. The #982-in-both specimen**: one entity, one canonical id, visible through every lens whose query matches. No data-model dedup. A cross-project aggregation surface may declare "collapse duplicates to the primary claimant" as a per-view query parameter — presentation choice, never stamped.

[...]

PR #1093, opened 2026-07-26
docs: research annotation layer and materialized views

### LINKED ISSUE #1089: Research: annotation layer and materialized views over joined streams
## Question

Much of what we hand-cobble (the DAG board, triage state, "this PR needs a rebase", agent-authored notes) may be **materialized views over joined data streams** rather than bespoke artifacts. That suggests a first-class **annotation layer**: authored statements *about* an entity that don't live inside it, written by human operators and persistent agents alike.

Research brief:
- Prior art for annotation-over-foreign-entity: k8s labels/annotations, git notes, Fossil, code-review systems, JSON-LD/RDF (**explicitly hoping this is not the conclusion** — the failure mode we're avoiding is unbounded vocabulary with no readable artifact; contrast with a declared schema per the #1072 meta-model ruling), event-sourced projections, materialized-view engines.
- How annotation systems survive the target being renamed/moved (→ #1039 stable ids) and being unavailable/offline.
- How authored annotations merge across writers without a single master (→ the authored-vs-observed question in the companion ticket).
- Whether triage state, board status, and convoy↔issue reverse indexes are better modeled as views over facts than as stored state.

## Starting position (not a ruling)

**Constitutive vs commentary**: a fact that is *why the resource exists* lives in the resource (a convoy's issue association is provenance, stamped at admission — #1037); a statement *about* it that its owner wouldn't accept lives as an annotation. Reverse indexes are views, never stored.

Deliverable: a research doc in the repo comparing approaches against flotilla's model, with a recommendation on whether the annotation layer is a resource kind, a separate stream, or a view convention.

rjwittams:
## Summary

- compare Kubernetes metadata, Git notes, Fossil, code-review metadata, RDF/JSON-LD, and event-sourced materialized views
- recommend typed immutable `AnnotationStatement` resources on Flotilla's existing resource log, with typed disposable effective-state and reverse-index views
- define identity, offline/multi-writer merge semantics, semantic retraction, and a classification of triage, board, dependency, and provenance data

## Validation

- checked every document link resolves

Closes #1089

### DIFF
--- a/docs/superpowers/research/2026-07-26-annotation-layer-and-materialized-views.md
+++ b/docs/superpowers/research/2026-07-26-annotation-layer-and-materialized-views.md
@@ -1,0 +1,658 @@
+# Annotation layer and materialized views over joined streams
+
+**Date:** 2026-07-26
+
+**Issue:** [#1089](https://github.com/flotilla-org/flotilla/issues/1089)
+
+**Status:** Research recommendation, not an ADR
+
+## Recommendation
+
+Flotilla should model an authored annotation as a first-class, durable
+`AnnotationStatement` resource carried by the resource store's existing event
+log. It should **not** add an annotation map to every target, create a second
+annotation-specific transport, or treat authored facts as view configuration.
+
+The three layers have different jobs:
+
+1. **Statements are resources.** An annotation is an immutable, attributed
+   statement about a stable subject identity. A semantic retraction or
+   correction is another statement; administrative hard deletion/redaction is
+   a separate policy question. The statement remains present when its subject
+   is renamed, unavailable, or deleted.
+2. **The resource log is the stream.** Annotation statements use the same
+   origin-authored logs, durable replicas, relay, and list/watch API as authored
+   definitions in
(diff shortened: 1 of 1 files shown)


rjwittams:
Addressed in 1d5dc2dc: all prose references to #1037, #1039, #1089, and #1090 are now explicit links, so they remain navigable when the research file is rendered as repository Markdown. The `locator_hint` example intentionally remains literal data.

rjwittams:
Addressed in cc14237e. The identity model now follows #1039's completed ruling:

- subjects use source-native canonical keys; Flotilla does not mint surrogate UIDs
- internal names, canonical Repository identities, and provider-native IDs are distinct typed key encodings
- canonical-key changes create a successor plus provenance edge and do not silently retarget annotations
- project-scoped facts target constituent resources when those are the actual subject; genuinely Project-level annotations use the Project name, stay historical on rename by default, and may only follow provenance under explicit schema policy/restatement
- the former open #1039 dependency is replaced with a concrete wire-encoding design follow-up

The revised section was independently re-reviewed against the #1039 resolution with no blocker found.

PR #1105, opened 2026-07-26
fix: suppress dispatch regard for no-attach convoys

### LINKED ISSUE #1099: --no-attach does not suppress the implicit convoy-dispatch regard
## What happened

`flotilla convoy start` mints an implicit Regard unconditionally in `create_convoy_with_regard` (crates/flotilla-core/src/in_process.rs:3693-3705) — `source: {kind: implicit, policy: convoy-dispatch}`, decaying after 300s, target = the Convoy. The emission never consults the intent's `auto_attach`.

With a PM connected this materialises a zellij tab even when the caller passed `--no-attach`. Robert (2026-07-26), after a batch dispatch that all passed the flag: "we seem to be automatically attaching zellij tabs for new convoys - I thought the idea was for them to stay latent."

## Why this is more than a cosmetic annoyance

Regard→materialisation is **correct per ADR 0018** — "opening a latent is expressing a regard … discovery and attention are one mechanism joined at the regard". The defect is the *claim*, not the effect: ADR 0018 defines a regard as **an intention to be looking at a thing**, and the live regard set is the principal's **searchlight**, which the ADR then uses for demand routing, escalation, and attention telemetry (the Yeoman's charter as a data join).

So a dispatch-time implicit regard asserts the principal intends to be looking at something they explicitly said they did not — a **false attention claim that pollutes the searchlight**, not just an unwanted tab. ADR 0018 already rules the adjacent case in the same spirit: ambient surfaces "neither emit nor refresh regards — displaying everything is not regarding everything." Auto-regarding on dispatch is that category error one level up.

Note the documented `--attach` heuristic (attach only when no PM connector is present) is a different code path; the regard bypasses it entirely.

Related history: this mechanism is behind the earlier auto-materialised-pane seat confusion — each dispatch holds a cleat seat for 300s (#1011, closed as a misdiagnosis of this symptom).

## What to build

When `auto_attach` is `Never` (`--no-attach`), do not emit the implicit convoy-dispatch regard. Default and `--attach` paths unchanged. Consider whether the implicit policy should exist at all for non-interactive dispatch, or whether it should target the *dispatching* principal only when that principal has a live surface.

## Acceptance criteria

- [ ] `convoy start --no-attach` mints no implicit convoy-dispatch regard and materialises no tab with a PM connected
- [ ] Default and `--attach` behaviour unchanged
- [ ] The searchlight (live regard set) contains no entry for a convoy dispatched with `--no-attach`
- [ ] Covered at InProcessDaemon level

## Blocked by

None - can start immediately

rjwittams:
## Summary

- suppress the implicit `convoy-dispatch` regard when `auto_attach` is `Never`
- preserve the dispatch-regard decision across prepared remote convoy starts
- cover no-attach searchlight behavior with an ambient presentation surface while retaining default and explicit attach policy coverage

## Validation

Closes #1099

### DIFF
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -21,13 +21,14 @@ use flotilla_protocol::{
     commands::RepositoryIdentityChange,
     qualified_path::{HostId, QualifiedPath},
     result_set::{ConvoyChangeRequest, ConvoyRow, ResultSet, Rows},
-    AttachBinding, Command, CommandAction, CommandValue, CorrelationKey, CrewCommandContext, CrewListMember, CrewListResponse, DaemonEvent,
-    DeltaEntry, EnvironmentId, FleetListResponse, FleetListRow, FleetReplicaSnapshot, FleetReplicaStatus, FleetStaleness, HostListResponse,
-    HostName, HostProviderStatus, HostProvidersResponse, HostStatusResponse, HostSummary, NodeId, NodeInfo, PeerConnectionState,
-    PrincipalRef, ProjectListEntry, ProjectListRepository, ProjectListResponse, ProviderData, ProviderInfo, QueryCursor, RepoDelta,
-    RepoDetailResponse, RepoIdentity, RepoInfo, RepoProvidersResponse, RepoSnapshot, RepoSummary, RepoWorkResponse, ResolvedPaneCommand,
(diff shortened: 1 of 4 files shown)


rjwittams:
Addressed the clarity observation in 01f78c05. The default-path test now loads the `default-regard` convoy and compares the regard principal directly against that convoy's persisted `dispatching_principal_ref`.

rjwittams:
Thank you for the re-review. Confirmed there are no further action items from this review cycle.

PR #1106, opened 2026-07-26
fix: refresh connected peer host heartbeats

### LINKED ISSUE #1092: feta: peer connected but ResourceHost heartbeat goes stale, blocking placement
## What happened

After a kiwi daemon bounce (2026-07-26 ~18:10Z), `flotilla host list` shows feta **connected**, but the ResourceHost's `status.heartbeat_at` froze at 18:10:32Z and never advanced (checked 18:40Z, 30 min stale). `apply_heartbeat_readiness` therefore marks it not-ready, and placement refuses:

```
error: placement `host-direct-d783…` host `d783…` is not ready
```

So the peer link is healthy while the resource layer believes the host is down — two views of the same host disagreeing, the same family as #1073 (host vanishing from `host list` mid-redial) but in the opposite direction. Effect: feta is unusable for placement despite being awake and peered; everything falls back to kiwi.

Suspicion: the heartbeat publisher or its subscription doesn't re-establish after a local daemon restart, while the peer transport does. Worth checking whether the heartbeat is pushed by the remote or polled by the local daemon, and what re-arms it on reconnect.

## Acceptance criteria

- [ ] A connected peer's ResourceHost heartbeat advances after a local daemon restart
- [ ] Placement readiness and `host list` connectivity cannot disagree indefinitely (or the disagreement is surfaced)
- [ ] Regression coverage at the InProcessDaemon level

rjwittams:
## Summary

- refresh locally materialized peer `Host` status on every daemon heartbeat tick while the peer route is `Connected`
- stop refreshing disconnected peers so the existing heartbeat TTL continues to mark them unready
- add an `InProcessDaemon` regression covering a restarted heartbeat publisher with a connected peer whose `ResourceHost` status had gone stale

## Root cause

A peer's locally materialized `ResourceHost` heartbeat was stamped only when its `HostSummary` arrived. The periodic heartbeat publisher updated only the local host, so a healthy peer connection could remain connected while the peer host row aged past the 60-second placement-readiness TTL.

## Impact

After a local daemon restart, connected peers keep advancing their materialized host heartbeat and remain eligible for placement. Connectivity and placement readiness now derive from the same live peer-route signal instead of diverging indefinitely.

## Validation

- `TMPDIR="$PWD/.codex-tmp" RUSTC_WRAPPER= CARGO_BUILD_JOBS=2 cargo test --workspace --locked --features flotilla-daemon/skip-no-sandbox-tests`

Closes #1092

### DIFF
--- a/crates/flotilla-core/src/host_registry.rs
+++ b/crates/flotilla-core/src/host_registry.rs
@@ -854,6 +854,22 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn connected_peer_summaries_only_returns_live_routes() {
+        let registry = HostRegistry::new(local_node(), minimal_summary(&local_node()));
+        let peer = peer_node();
+        let summary = minimal_summary(&peer);
+        registry.publish_peer_summary(summary.clone(), &|_| {}).await;
+
+        assert!(registry.connected_peer_summaries().await.is_empty(), "a disconnected peer must not receive heartbeat refreshes");
+
+        registry.publish_peer_connection_status(&peer, PeerConnectionState::Connected, &HashMap::new(), &|_| {}).await;
+        assert_eq!(registry.connected_peer_summaries().await, vec![summary]);
+
+        registry.publish_peer_connection_status(&peer, PeerConnectionState::Disconnected, &HashMap::new(), &|_| {}).await;
+        assert!(registry.connected_peer_summaries().await.is_empty(), "a disconnected peer must stop receiving heartbeat refreshes");
+    }
+
+    #[tokio::test]
     async fn replay_uses_environment_id_stream_keys() {
         let registry = HostRegistry::new(local_node(), minimal_summary(&local_node()));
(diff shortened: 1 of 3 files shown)


rjwittams:
Fixed in ddbce0de. Connected peer heartbeat refresh now handles materialization errors per peer, logs the affected peer, and continues through the remaining live routes, so a resource-version race cannot starve later peers or misreport the already-successful local heartbeat. I also added a host-registry test proving disconnected peers are excluded and stop receiving refreshes.

Validated with the focused host-registry test and both daemon heartbeat tests.

rjwittams:
Acknowledged. The per-tick `HostStatus` write is intentional: `heartbeat_at` must advance within the 60-second readiness TTL, so a connected peer necessarily produces one status event per 30-second heartbeat interval. The `PlacementPolicy` path is a read on steady state and writes only if the advertised pool changes. At current mesh cardinality this is the bounded liveness signal the fix requires; a future move to direct replicated host status can remove the synthetic projection rather than optimizing away its heartbeat.

PR #1109, opened 2026-07-26
Wire definitions-class Project replication

### LINKED ISSUE #1098: Wire ReplicationClass::Definitions per ADR 0016, with Project as proving kind
## Parent

Enactment of the #1090 ruling. ADR 0016 already specifies the semantics; this is the missing implementation.

## What to build

`ReplicationClass` (crates/flotilla-resources/src/replica.rs) has only `None` and `HomeBoundRuntime`. ADR 0016's **definitions** class — human-named, edit-anywhere-including-offline, conflicts surfaced never auto-resolved — is unimplemented, so Project / PlacementPolicy / WorkflowTemplate / Repository fall through to `None` and are effectively owned by whichever daemon they were authored on.

Add the class and its read path per ADR 0016:
- durable materialized replicas per origin root (replicas area keyed by origin; **disk state, not cache** — a sleeping peer must not make the merged view regress);
- relay: any root may serve its replica of any origin's immutable log segments;
- **deterministic merge view** over {own log, replica per peer} — never written back as authority;
- **per-field causal merge, not last-writer-wins**: a write made having seen the current value supersedes silently however late it syncs; concurrent writes to the *same* field become causally-maximal siblings badged as a conflict; disjoint-field concurrent edits both stick; resolution is an ordinary write.

**Project is the proving kind.** Do not extend to annotation statements — separate immutable statements about one subject need a slot-level reducer that is a scoped ADR 0016 amendment, deliberately sequenced after this.

## Acceptance criteria

- [ ] A Project edited on kiwi is visible and editable on feta, and vice versa, with no designated master
- [ ] An edit made while a peer is asleep converges when it wakes, in both directions
- [ ] Concurrent same-field edits surface as a badged conflict rather than one silently winning; disjoint-field edits both apply
- [ ] Resolving a conflict is an ordinary write
- [ ] A peer going away (or a local restart while it is away) changes nothing observable in the merged view
- [ ] Contract tests run against both storage implementations per the testing philosophy

## Blocked by

None - can start immediately

rjwittams:
## Summary

- add `ReplicationClass::Definitions` and opt Project definitions into causal, field-level replication
- persist definition metadata and replica provenance across in-memory and SQLite backends
- relay definitions across direct and multi-hop connections without echo loops, and expose conflicts in Project CLI views
- cover cross-backend edits, conflict resolution, deletion semantics, restarts, reconnects, and multi-hop replication

Closes #1098

## Validation

### DIFF
--- a/crates/flotilla-core/src/agent_adapter.rs
+++ b/crates/flotilla-core/src/agent_adapter.rs
@@ -914,6 +914,7 @@ mod tests {
                 finalizers: Vec::new(),
                 deletion_timestamp: None,
                 creation_timestamp: Utc::now(),
+                merge: None,
             },
             spec: ConvoySpec {
                 workflow_ref: "workflow".to_string(),
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -34,7 +34,7 @@ use flotilla_resources::{
     apply_status_patch_checked as apply_resource_status_patch_checked, external_patches as convoy_external_patches, get_resource_kind,
     list_resource_kind, list_resource_kind_including_replicas, normalize_project_spec, repository_display_labels,
     resolve_project_issue_sources, terminal_session_attach_target, watch_resource_kind, watch_resource_kind_from,
-    watch_resource_kind_including_replicas, Checkout as ResourceCheckout, CheckoutIntegrationStatus,
+    watch_resource_kind_including_replicas, watch_resource_kind_replica_sources, Checkout as ResourceCheckout, CheckoutIntegrationStatus,
     CheckoutPhase as ResourceCheckoutPhase, CheckoutSpec as ResourceCheckoutSpec, CheckoutStatus as ResourceCheckoutStatus, ConditionValue,
     Convoy as ResourceConvoy, ConvoyIssue, ConvoyRepositorySpec, ConvoySpec, ConvoyStatusPatch, CrewSource,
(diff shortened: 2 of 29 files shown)


rjwittams:
Addressed in `922914a2`:

- `ProjectSpec::issue_source` now serializes `None` as an explicit JSON `null`, so clearing the field receives a causal dot and supersedes a replicated non-null value.
- An apply that changes one field no longer stamps unrelated conflicted fields. A pure no-op apply still intentionally resolves the submitted merged view, preserving the existing “choose the displayed sibling” flow.
- `replicaSources` now accepts camelCase, kebab-case, and snake_case variants, matching `includeReplicas`.
- Added a three-node `TerminalSession` relay test covering HomeBoundRuntime provenance preservation and anti-echo behavior.

Regression coverage runs against both in-memory and SQLite definition stores. The exact formatting, strict workspace clippy, and sandbox-safe full workspace test commands all pass locally.

rjwittams:
Addressed the follow-up finding in `ab99b0cf`.

Both replica backends now order every per-object event against the latest accepted sync timestamp. Deletes retain tombstones (durably in SQLite), so:

- a stale delete cannot remove a newer write;
- a stale write cannot resurrect a newer delete;
- a genuinely newer write can recreate the object;
- rejected stale events do not rewind the replica cursor.

Added shared in-memory/SQLite regressions for out-of-order write/delete delivery and a SQLite restart regression proving tombstone durability. I also corrected the existing replica contract's artificial timestamps so its delete is chronologically newer than its update.

Final-tree validation passes:

Issue #1102, opened 2026-07-26
Build a layered crew image and publish it to the lab registry


rjwittams:
## What to build

Fork and multi-repo work is blocked because contained vessels have no usable image: the only docker placement points at bare `ubuntu:24.04`, and the one built image (`flotilla-dev-env:latest`, 2026-03-30) is local-only and stale, which #1087 shows a placement cannot pull.

Build a **hand-curated, sensibly layered** crew image and publish it to the lab Forgejo OCI registry (`forgejo.lab.flotilla.work`, already live — `/v2/` answers 401). Per the #1088 partial ruling, placements then reference a known pullable tag; nothing builds at dispatch time.

Layer deliberately, cheapest-changing last, so rebuilds are not wholesale (Robert: "very easy to spend all your time doing image stuff"):
1. base OS + certs + git + curl;
2. language toolchains needed to *verify* crew work (Rust with the pinned nightly per CLAUDE.md, node);
3. forge CLI (`gh`) and utilities;
4. agent adapters (claude-code, codex) — the fastest-moving layer, so last.

`.flotilla/Dockerfile.dev-env` is prior art for content but was built as one 4.35GB blob; treat it as a source of package lists, not as the layering. A throwaway repo on Forgejo holding the Dockerfile is fine for now.

The image must satisfy the adapters a placement declares (`docker_per_vessel.agent_adapters`) — declaring `claude-code` and `codex` is a promise this image has to keep, and today nothing verifies it. Include a smoke check that both binaries run inside the built image.

Credentials are explicitly **out of scope** — `CreateOpts.tokens` is fed empty and belongs to #1050. The acceptance below is only that the image exists, pulls, and runs the adapters.

## Acceptance criteria

- [ ] Dockerfile with the layer order above, in a repository (Forgejo throwaway is acceptable), with a documented build command
- [ ] Image pushed to the lab Forgejo registry with an explicit version tag (not only `latest`)
- [ ] `docker run <tag> claude --version` and `codex --version` both succeed
- [ ] A PlacementPolicy referencing the pushed tag is applied and documented
- [ ] Rebuilding after an adapter-layer change does not rebuild the toolchain layers

## Not in scope

Lifecycle/staleness policy, whether image-ensure belongs in step plans, and the full layer-ownership contract — all still open on #1088.

## Blocked by

None - can start immediately

rjwittams:
[...]

Issue #1088, opened 2026-07-26
Image sourcing and lifecycle: where container images come from, and who builds them


rjwittams:
Part of #1046.

## Question

The first Plane-B contained dispatch (#1081 fork tracer) died at #1087: the placement tried to *pull* a locally-built image. That exposed an unruled seam — `PlacementPolicy.docker_per_vessel.image` is a literal tag, `.flotilla/environment.yaml` is a *build recipe*, and `ImageEnsured`/`EnvironmentCreated` exist as step results no verb drives.

Robert's starting position (2026-07-26): environment.yaml wants a rethink; the old behaviour kept building new images, which is wrong without careful layering ("very easy to spend all your time doing image stuff"); **better to start from a hand-curated image**; and building images inside step functions is strange — step plans are for short user-facing actions, not long-lived infrastructure.

Rule:
- **A)** Does a placement reference a hand-curated tag (built out-of-band, by a human or a script) or an image *source* flotilla ensures? Is the build recipe retired from the dispatch path entirely for v1?
- **B)** The layering contract — #1049 already ruled "layered image contract + one-run parity acceptance"; what are the layers concretely (base OS+toolchain / agent adapters / project deps), and who owns each?
- **C)** Lifecycle: when does an image get rebuilt, who notices it's stale, and how does a convoy know which image it ran on (settlement testimony)?
- **D)** Does image-ensure leave the step-plan machinery, and if so what drives it?

Adjacent and probably sequenced with #1050 (credentials): a contained vessel needs both a usable image and injected credentials before fork work can actually run — today `CreateOpts.tokens` is fed empty.

HITL grill with Robert.

rjwittams:
**The lab already has a container registry — no need to stand one up.** Forgejo ships an OCI registry and it is live on the lab instance: `https://forgejo.lab.flotilla.work/v2/` returns **401** (registry present, authenticate) rather than 404, on Forgejo 15.0.3. The packages API returns 403 for the planner token, so it's a token-scope question, not a capability question.

That gives option A of this grill a concrete shape: hand-curated images get built (by a human or an infra agent), pushed to `forgejo.lab.flotilla.work/<owner>/<image>`, and placements reference a pullable tag — which is the normal path, no special-casing of local images, and it makes #1087's "can't use a locally-built image" much less pressing (it stays a real gap for air-gapped/dev-loop cases, but stops being the blocker). Credential scoping for registry pull lands naturally in #1050.

[...]

PR #1112, opened 2026-07-26
feat: publish layered crew image

rjwittams:


rjwittams:
Reviewed both points; no change is needed in this image PR.

The root-ownership concern does not materialize for the checked-in policy: `DockerEnvironmentProvider::create` renders every `provisioned_mount` as `host_path:environment_path:ro`, so the container cannot create root-owned files in the host worktree. Choosing a writable-mount ownership model belongs to the placement/runtime contract rather than this image.

The tag/adapter synchronization point is part of the lifecycle and staleness policy explicitly deferred by #1102 to #1088. This PR keeps the deployed contract explicit and immutable (`2026-07-26.1`) and documents that adapter upgrades require a new release tag.

Issue #1111, opened 2026-07-26
Daemon becomes a zombie: process alive, socket accepts, nothing served, no panic


rjwittams:
## What happened

2026-07-26 ~22:26:30Z, kiwi's daemon stopped serving while remaining alive. Every CLI call fails:

```
error: daemon connection closed
error: failed to flush: Broken pipe (os error 32)
```

State observed:
- `pgrep -x flotillad` → one live process (started 22:12:58Z).
- `~/.config/flotilla/flotilla.sock` present; connections are accepted then immediately closed.
- **Daemon log: complete silence from 22:26:30Z onward** (verified 33 minutes later). Last entries were replicator backoff warnings and a routine repo delta.
- `daemon-panic.log`: absent/empty. No panic recorded.
- `sample` of the process: main thread parked in `Runtime::block_on`; **all 12 `tokio-rt-worker` threads idle with zero application frames** (all in the blocking-task/park path). No spinning, no deadlocked application code — the runtime is simply doing nothing.

Diagnosis: the socket accept loop / server task ended without bringing the process down, so the daemon is a zombie — indistinguishable from healthy by process check or socket presence, but serving nothing. This is worse than a crash: a crash gets respawned by `connect_or_spawn`, whereas this state persists indefinitely and looks fine to every liveness check we have.

## Suspected trigger

The last commands issued before the silence were, in order: `convoy delete crew-image-build-feta --force` (a **peer-homed** convoy — the record lives on feta, so this routes a remote command) and then a `convoy start`. The delete returned `daemon connection closed`, suggesting the daemon died mid-request. Not proven, but forced teardown of a peer-homed convoy is the strongest candidate; remote-command routing while a peer replicator is in backoff (it was — see #1104) is a plausible interaction.

## What to build

- The accept loop ending must be fatal: if the server task exits, the process exits, so respawn logic can do its job.
- A liveness signal that distinguishes serving from existing — the current socket-file-plus-pid check cannot. Consider a cheap health request the CLI can use before reporting a confusing pipe error.
- Investigate the peer-homed forced-teardown path for an unhandled error that can escape the request handler and take the listener with it.

## Acceptance criteria

- [ ] Killing/ending the server task terminates the process (no zombie state)
- [ ] A wedged or dead daemon produces an actionable CLI message, not `failed to flush: Broken pipe`
- [ ] Forced teardown of a peer-homed convoy is covered by a test, including while the peer's replicator is failing
[...]

PR #1116, opened 2026-07-26
feat: log daemon runtime liveness and supervisory-task abort

rjwittams:
Diagnostics for #1111 — a daemon that stopped serving while staying alive, leaving no trace whatsoever.

- **`DaemonRuntime::drop` logs before aborting.** Dropping that value silently aborts heartbeat, replica refresh, aggregator, and controller tasks, leaving a process that satisfies every liveness check we have (pid alive, socket accepting) while doing no background work. It now says so at ERROR with the task count.
- **A liveness watchdog** emits one INFO line per minute carrying uptime and supervisory-task count. The point is that a wedge becomes visible through the *absence of a known-cadence marker*, rather than through the absence of incidental debug chatter — which was the only evidence available in #1111, and only noticeable in hindsight.

Neither is the fix for #1111. That trigger was most likely a full disk, and the real defect is that the daemon never recovers once space returns (a fresh daemon on the same disk works immediately). These changes make the next occurrence diagnosable rather than archaeological.

Gates: `cargo +nightly-2026-03-12 fmt --check` clean; `cargo build --bin flotillad` clean.

https://claude.ai/code/session_019iE3vGfKm6bZh7B6i1CUuY

### DIFF
--- a/crates/flotilla-daemon/src/cli.rs
+++ b/crates/flotilla-daemon/src/cli.rs
@@ -31,7 +31,12 @@ pub async fn run(socket_path: &Path, config_dir: &Path, state_dir: &Path, timeou
     let repo_root_paths = repo_roots.into_iter().map(|p| p.into_path_buf()).collect();
     let server = DaemonServer::new(repo_root_paths, Arc::clone(&config), discovery, socket_path.to_path_buf(), timeout).await?;
     let daemon = server.daemon();
-    let _runtime = DaemonRuntime::start(daemon, Arc::clone(&config), Some(socket_path.to_path_buf())).await?;
+    let runtime = DaemonRuntime::start(daemon, Arc::clone(&config), Some(socket_path.to_path_buf())).await?;
 
-    server.run().await
+    let result = server.run().await;
+    // Every path out of `run` here is an intended stop (SIGTERM, SIGINT, idle
+    // timeout, explicit shutdown); say so, so the runtime's Drop ERROR keeps
+    // meaning "this went away when it should not have".
+    runtime.shutdown();
+    result
 }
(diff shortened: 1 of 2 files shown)


rjwittams:
Both fixed in 9a73289e — the first was a genuine defect in the change, not a nit.

**1. Drop ERROR on every normal shutdown — correct, and it undercut the whole point.** Verified the path: `cli.rs:34` holds `_runtime` across `server.run().await`, and `run` returns `Ok(())` on SIGTERM (server.rs:387), SIGINT (383), the idle-timeout shutdown signal (377) and explicit shutdown — so the ERROR would have fired on every routine exit, indistinguishable from the #1111 wedge it exists to reveal. Exactly the trade you named: no trace at all, swapped for constant false positives.

Implemented your suggested shape. `DaemonRuntime::shutdown` marks the stop as intended and logs at INFO; `Drop` shouts only when reached *without* that flag, and logs at DEBUG when it follows an expected shutdown. `cli.rs` now calls `runtime.shutdown()` after `run` returns, since every exit there is intended. Used a plain `bool` field rather than an `AtomicBool` — `shutdown` takes `self` and `Drop` has `&mut self`, so there is no shared access to synchronise.

**2. Watchdog/Drop count off-by-one — fixed.** The count is now computed as `tasks.len() + 1` before the push, with a comment saying why, so the periodic `daemon alive` line and the abort log agree. Worth fixing precisely because those two numbers exist to be cross-referenced while chasing a future wedge; disagreeing for no reason would send someone down a false trail.

Gates: `fmt --check` clean, `clippy -p flotilla-daemon --all-targets -D warnings` clean, 264 daemon lib tests pass.

PR #1121, opened 2026-07-27
fix: re-list replicas across peer generations

### LINKED ISSUE #1104: Peer restart strands the replica cursor: convoys on that peer vanish from the dispatcher's view
## What happened

feta's daemon was restarted (2026-07-26 21:36Z, routine bump). kiwi's peer link re-established immediately and `topology` shows feta connected. But **kiwi can no longer see any convoy homed on feta**: two convoys dispatched from kiwi with feta placement run happily (visible in `flotilla ls` *on feta*) and are entirely absent from `flotilla ls` on kiwi. The governor loses sight of its own fleet, silently.

kiwi's log shows why:

```
WARN replicator: resource replicator failed; restarting after backoff
  peer=9ad3ffb1931f9996979535186f45027f generation=1 kind="Host"
  error=WATCH resources: … /hosts?watch=true&resourceVersion=12919
… same for Convoy, Vessel, TerminalSession (resourceVersion=257, 7676, …)
```

Those resourceVersions belong to feta's **previous** store. The replicator is tailing `generation=1` and retrying a cursor that no longer exists, backing off and re-trying the same stale position rather than re-listing.

This is precisely the failure ADR 0004 warns about: ephemeral observed kinds are **generational**, `watch(FromVersion(v))` is valid *only within a generation*, and "a consumer … that sees the generation change must discard its view and re-list from the new generation — it must not expect version continuity across a restart." The replicator does not honour that contract on peer restart.

## Why it is severe

- **Silent.** The only signal is a WARN in the daemon log. `topology` says connected; `host list` says connected; the replica-status table simply *omits* feta rather than reporting an error (contrast the udder row, which does show its parse failure).
- **It breaks the thing multi-host exists for.** Remote-placement dispatch routes convoy creation to the peer (`ConvoyStartPrepared`), so the convoy is homed there and the dispatcher learns of it only by replication. With the cursor stranded, every convoy sent to that peer is invisible from the machine that sent it.
- It compounds #1092 (host heartbeat frozen after the *local* daemon restarts). Same family: **subscriptions that do not re-arm across a restart, in both directions.**

## What to build

- On peer reconnect, detect a generation change and **re-list from the new generation** rather than resuming a stale `resourceVersion` (ADR 0004's stated contract).
- Surface a stranded/failed replica in `replica status` — an omitted row must never mean "fine"; report the peer with its error, as the udder row does.
- Consider treating repeated WATCH-cursor failures as an automatic re-list trigger regardless of generation reporting.

## Acceptance criteria

- [ ] Restart a peer daemon; the local replicator re-lists and peer-homed resources reappear without a local restart
- [ ] A convoy dispatched to a peer is visible from the dispatching host across a peer restart
- [ ] Replica status shows an explicit row for a peer whose replication is failing
- [ ] Regression coverage for the generation-change path (InProcessDaemon level per the testing philosophy)

## Blocked by

None - can start immediately

rjwittams:
## Summary

- inspect the origin generation before resuming a stored replica cursor
- atomically replace replica rows and restart the watch when the generation changes or a resumed watch fails
- track per-peer, per-kind replication failures and surface them in `flotilla ls` replica status
- add regression coverage for proactive generation re-listing and explicit failed-peer status

## Root cause

The peer replicator resumed its stored `resourceVersion` before checking the restarted origin's generation. If the stale watch did not return one of the narrowly recognized cursor errors, the supervisor retried the same cursor indefinitely. Resource-replication health was also disconnected from fleet replica status, so the failed peer could be omitted from the status table.

## Validation

- targeted InProcessDaemon, overlay replication, generation-change, and fleet-status regressions

Closes #1104

### DIFF
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -657,6 +657,24 @@ fn replica_staleness(entry: &FleetReplicaCacheEntry, now: DateTime<Utc>) -> Flee
     }
 }
 
+fn format_resource_replication_failures(failures: &[ResourceReplicationFailure]) -> Option<String> {
+    if failures.is_empty() {
+        return None;
+    }
+    Some(format!(
+        "resource replication failed: {}",
+        failures.iter().map(|failure| format!("{}: {}", failure.kind, failure.message)).collect::<Vec<_>>().join("; ")
+    ))
+}
+
+fn join_replica_errors(first: Option<&str>, second: Option<&str>) -> Option<String> {
+    match (first, second) {
+        (Some(first), Some(second)) => Some(format!("{first}; {second}")),
+        (Some(message), None) | (None, Some(message)) => Some(message.to_string()),
+        (None, None) => None,
+    }
+}
+
 #[derive(Debug, Default)]
 struct ReplicaParseDiagnostics {
     skipped_records: usize,
@@ -1185,6 +1203,12 @@ struct FleetReplicaCacheEntry {
     last_error: Option<String>,
 }
 
+#[derive(Debug, Clone)]
+struct ResourceReplicationFailure {
+    kind: String,
+    message: String,
+}
+
 const SUPERSEDED_BY_ANNOTATION: &str = "flotilla.work/superseded-by";
 
 #[derive(bon::Builder)]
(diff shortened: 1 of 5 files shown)


rjwittams:
Fixed in 07da705f. A resumed watch that fails mid-stream now takes a fresh origin list before replacing the replica view and starting the replacement watch, so it cannot roll back to the pre-watch snapshot. Added `http_replicator_takes_a_fresh_list_after_a_resumed_watch_fails` to cover the LIST → failed WATCH → fresh LIST → replacement WATCH sequence.

rjwittams:
Acknowledged. The re-review confirms the blocking correctness issue is fixed and identifies no new issues. The routed-watch observation is test-support-only and does not affect the production HTTP path covered by this bugfix, so no further change is needed in this PR.

PR #1122, opened 2026-07-27
fix: deduplicate prepared convoy snapshots

### LINKED ISSUE #1108: Prepared workflow/placement snapshots leak into the UI and never get cleaned up
## What happened

Robert, reading the convoys table (2026-07-26): "we have 4 that are single-agent-contained, and two with workflows named for the convoy. Are they custom one offs?"

They are not. Remote-placement dispatch ships the *resolved* workflow spec to the target host, which materialises a snapshot named `<convoy>-remote-workflow-<content-hash>` (`ensure_prepared_workflow_snapshot`, crates/flotilla-core/src/in_process.rs:3190; `ensure_prepared_placement_snapshot` does the same for placement policies). The convoy's `workflow_ref` then points at the snapshot, so the table shows a synthesized name instead of the logical template.

The **intent is right** and matches ADR 0015 (intent completion at admission): pinning the exact spec stops a peer running a different version of `single-agent-contained` than the one admitted. Two things are wrong around it.

### 1. Snapshots accumulate forever

feta currently holds **49** WorkflowTemplates, of which ~42 are snapshots from long-dead convoys — `awareness-band-…`, `compact-rail-mock-…`, `attach-watch-fallback-…`, and so on. Convoy teardown does not remove them. Same presumably true for prepared placement policies.

### 2. The implementation detail leaks into the user-visible surface

A workflow column exists so a human can see *which workflow* a convoy runs. Showing `no-attach-regard-feta-remote-workflow-d517cb83f44e` defeats exactly that purpose, and makes locally-dispatched and remotely-dispatched convoys look categorically different when they are running the same workflow. The convoy should retain the **logical** template name for display (and for grouping/filtering), with the pinned snapshot as an internal reference.

Note the hash is a content hash, identical across convoys using the same template — so the name is doubly redundant: it repeats the convoy name that is already the row's identity.

## What to build

- Convoys carry both the logical workflow name (for display, grouping, and `--workflow` round-tripping) and the pinned snapshot reference (for execution).
- Prepared workflow and placement snapshots are garbage-collected — on convoy teardown, or by reference-counting content-hash-identical snapshots so one snapshot serves every convoy that pinned the same spec.
- Consider whether the content hash alone should name the snapshot (`workflow-snapshot-<hash>`), making dedup automatic and the name honest.

## Acceptance criteria

- [ ] The convoys table shows the logical workflow name regardless of local or remote dispatch
- [ ] Snapshots do not accumulate per convoy; a repeat dispatch of the same template creates no new snapshot
- [ ] Existing orphaned snapshots on feta are cleanable (a one-off sweep is acceptable)
- [ ] Placement snapshots covered by the same rule

rjwittams:
Closes #1108.

## What changed

- preserve each convoy's logical workflow and placement names while storing pinned execution snapshots in internal annotations
- name prepared workflow and placement snapshots solely by content hash, so identical specs deduplicate across convoys
- garbage-collect unreferenced prepared snapshots during convoy finalization
- sweep unreferenced legacy `<convoy>-remote-{workflow,placement}-<hash>` snapshots when repository resources initialize
- validate prepared snapshot names against their shipped contents

## Why

Remote admission previously replaced the logical references in `ConvoySpec` with convoy-qualified snapshot names. That leaked an execution detail into the convoys table and created one resource per convoy even when the admitted specs were identical. Teardown also never removed those resources.

The convoy now retains logical intent for display and CLI round-tripping, while annotations carry the pinned references used by reconcilers. A reference-aware collector preserves shared snapshots until the final convoy releases them.

## Validation

### DIFF
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -2330,16 +2330,13 @@ impl InProcessDaemon {
             return Err(format!("namespace `{requested_namespace}` is not served by this daemon (configured namespace: `{namespace}`)"));
         }
         let fallback_principal = PrincipalRef::implicit_for_namespace(&namespace);
-        let mut admission =
-            self.prepare_convoy_admission(&namespace, intent, dispatching_principal_ref.unwrap_or(&fallback_principal)).await?;
+        let admission = self.prepare_convoy_admission(&namespace, intent, dispatching_principal_ref.unwrap_or(&fallback_principal)).await?;
         let workflow_value = serde_json::to_value(&admission.workflow).map_err(|error| error.to_string())?;
-        let workflow_name = prepared_snapshot_name(&admission.name, "workflow", &workflow_value)?;
-        admission.spec.workflow_ref.clone_from(&workflow_name);
+        let workflow_name = prepared_snapshot_name("workflow", &workflow_value)?;
         let (placement_policy_name, placement_policy_spec) = match admission.placement_policy {
             Some(spec) => {
                 let value = serde_json::to_value(spec).map_err(|error| error.to_string())?;
-                let name = prepared_snapshot_name(&admission.name, "placement", &value)?;
-                admission.spec.placement_policy = Some(name.clone());
+                let name = prepared_snapshot_name("placement", &value)?;
                 (Some(name), Some(value))
             }
             None => (None, None),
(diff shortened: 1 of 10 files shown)


rjwittams:
Fixed in 00816b2d. Prepared admission now persists a convoy claim containing the pinned snapshot annotations and a `prepared-snapshot-pending` marker before materializing either snapshot. GC therefore protects first-time content throughout admission, while convoy reconciliation no-ops until snapshot creation succeeds and the marker is removed with optimistic retry. Admission failure deletes the incomplete claim and collects any partial snapshots. Added a regression test proving that a pending claim both protects its snapshot from GC and cannot bootstrap early.

rjwittams:
Addressed the remaining minor correctness points in f44d39cb. The new warning fields now use `%namespace`. Startup also recovers crash-interrupted claims before controllers start: claims with both pinned resources materialized have their pending marker cleared, while incomplete claims are discarded and partial snapshots swept. Regression coverage exercises both recovery outcomes. The name-pattern compatibility sweep and its current-scale list cost remain intentional, non-blocking tradeoffs for cleaning legacy resources.

rjwittams:
Thanks for the final pass. Confirmed: the remaining name-pattern compatibility sweep and current-scale list cost are intentional, non-blocking tradeoffs; there are no further code changes from this review.

PR #1126, opened 2026-07-27
fix: preserve Docker workspace mount mode

### LINKED ISSUE #1123: Docker workspace mount is read-only: reconciler asks for Rw, runtime erases the mode
## What happened

Found by the #1117 plumbing assessment (research doc in PR #1120): the `worktree_on_host_and_mount` bind mount for docker placements is **unconditionally read-only**. The reconciler requests `Rw`, the runtime erases the mode, and Docker receives `:ro`.

Consequence: a contained crew cannot write to `/workspace` at all. Every adapter writes the crew brief under the workspace (and Claude writes more besides), so **no contained vessel can function** — this is the first hard blocker in the contained path, ahead of everything else.

It also answers a question we had about Plane A: uid/gid mismatch was never the problem people expected, because nothing got far enough to hit it. The research verdict is "broken before uid/gid becomes relevant".

## What to build

Preserve the requested mount mode end to end so an `Rw` request produces a writable bind mount. Add coverage that asserts the generated Docker invocation carries the requested mode rather than a default — a mode silently downgraded to `:ro` is exactly the class of defect that unit tests over intermediate structs miss.

## Acceptance criteria

- [ ] An `Rw` mount request results in a writable `/workspace` in the container
- [ ] The generated Docker arguments are asserted against the requested mode
- [ ] A contained crew can write its brief into the workspace

## Blocked by

None - can start immediately. Note that a writable mount then exposes the uid/gid question (#1123's sibling: container user remapping is absent), which may surface next.

rjwittams:
## What changed

- carry the requested mount mode through provisioned mount metadata
- translate resource mount modes in the shared controller actuator path used by the daemon runtime
- generate Docker bind-mount arguments with the requested `:rw` or `:ro` mode
- assert the final Docker invocation includes a writable `/workspace` mount while retaining read-only reference mounts

## Validation

Closes #1123

### DIFF
--- a/crates/flotilla-controllers/src/actuators/mod.rs
+++ b/crates/flotilla-controllers/src/actuators/mod.rs
@@ -3,12 +3,14 @@ use std::{path::PathBuf, sync::Arc};
 use flotilla_core::{
     path_context::{DaemonHostPath, ExecutionEnvironmentPath},
     providers::{
-        environment::{CreateOpts, EnvironmentProvider, ProvisionedMount},
+        environment::{CreateOpts, EnvironmentProvider, ProvisionedMount, ProvisionedMountMode},
         terminal::TerminalPool,
         vcs::{CloneInspection, CloneProvisioner},
     },
 };
-use flotilla_resources::{DockerEnvironmentSpec, FreshCloneCheckoutSpec, TerminalSessionSource, TerminalSessionSpec};
+use flotilla_resources::{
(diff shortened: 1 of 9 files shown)


rjwittams:
Reviewed; no code changes needed. Existing container-label compatibility is intentionally not preserved during the repository's documented no-backwards-compatibility phase, and the noted trailing comma is the output of the pinned `cargo +nightly-2026-03-12 fmt`. Thanks for tracing the production path and final Docker argument coverage.

PR #1127, opened 2026-07-27
fix: fall back to watch for held controller seats

### LINKED ISSUE #1010: flotilla attach: fall back to watch with a banner when the controller seat is held; add --take
**Superseded by flotilla-org/cleat#177 — this is cleat behaviour, not flotilla behaviour.**

Original symptom (Robert, live 2026-07-25): convoy tabs try to attach and sit refused, with no explanation, because the session's controller seat is already held (cleat enforces one controller; watchers are read-only).

**Cause, corrected twice.** It is *not* the crew runner holding a seat — that runner launches detached, and the claim was withdrawn on #1011. The seat was held by **other zellij panes' `cleat attach` clients**: auto-materialised panes, and orphans left by a ghost zellij server after a restart.

**Layer, corrected (2026-07-27).** A flotilla-side implementation was attempted — detect the held seat via `cleat inspect`, substitute `cleat watch`, render our own banner — and rejected: "none of this should be flotilla behaviour". Every cleat consumer would reimplement the same logic, and seat policy would live outside the component that owns the seat. PR #1127 was closed unmerged.

The behaviour belongs in cleat and is tracked there:
- **flotilla-org/cleat#177** — force-take (demote controller to watcher), watcher banner as cleat chrome, attachment identity so banners can name the holder, and now also the plain automatic fallback: `cleat attach` on a held session should watch rather than refuse.

**Remaining flotilla-side work: none for the fallback itself.** Once cleat#177 lands, flotilla simply calls `cleat attach` and inherits the behaviour. Separate flotilla-side contributors to the original symptom, both tracked elsewhere: auto-materialised pane volume (reduced by #1099, which stopped `--no-attach` dispatches minting an implicit regard) and orphaned panes surviving a zellij restart (ghost-server cleanup).

Keeping this open only as the dogfood record; close it when cleat#177 lands and the symptom is verified gone.

rjwittams:
Closes #1010.

## What changed

- Inspect cleat controller-seat state while resolving terminal attaches.
- Automatically resolve held seats to `cleat watch` with a visible holder banner, including crew-runner inference until cleat exposes attachment identity.
- Retry a failed attach as watch to cover the inspect/attach race.
- Add `flotilla attach --take` and preserve takeover intent across remote attach hops.
- Detect whether the installed cleat supports `attach --take` and return a clear compatibility message linked to cleat#177 when it does not.

## Impact

Convoy tabs and direct `flotilla attach` calls no longer fail when the crew runner or another principal holds the single controller seat. Users land in a legible read-only watch instead and can request takeover explicitly when the cleat capability is available.

## Validation

### DIFF
--- a/crates/flotilla-core/src/executor.rs
+++ b/crates/flotilla-core/src/executor.rs
@@ -376,6 +376,8 @@ pub async fn build_plan(
         | CommandAction::QueryHostStatus { .. }
         | CommandAction::QueryHostProviders { .. }
         | CommandAction::Attach { .. }
+        | CommandAction::AttachTake { .. }
+        | CommandAction::AttachWatch { .. }
         | CommandAction::AttachTransient { .. }
         | CommandAction::QueryIssues { .. }
         | CommandAction::QueryIssueFetchByIds { .. }
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -6141,7 +6141,11 @@ impl InProcessDaemon {
                     let holder = if holder.is_empty() { fallback_holder.to_string() } else { holder };
                     (pool.watch_args(attach_target.session_id)?, Some(holder))
                 }
-                Ok(None) | Err(_) => (pool.attach_args(attach_target.session_id, attach_target.launch_command, &cwd, &Vec::new())?, None),
+                Ok(None) => (pool.attach_args(attach_target.session_id, attach_target.launch_command, &cwd, &Vec::new())?, None),
+                Err(error) => {
+                    warn!(session = attach_target.session_id, %error, "could not inspect terminal controller; attempting control attach");
+                    (pool.attach_args(attach_target.session_id, attach_target.launch_command, &cwd, &Vec::new())?, None)
+                }
             },
             AttachSeatRequest::WatchIfControlled => {
                 let holder = pool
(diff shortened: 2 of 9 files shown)


rjwittams:
Addressed in ee46e591: controller inspection failures now emit a warning with the session and error before deliberately failing open to the normal control attach. I also added a regression test covering that path.

The broader retry-classification concern needs a precise cleat failure contract; I filed #1128 to track that follow-up rather than weakening the acceptance-critical TOCTOU fallback in this PR. The `--take` capability probe intentionally tokenizes cleat's boolean flag from `attach --help` output.

rjwittams:
Addressed in ee46e591: controller inspection failures now emit a warning with the session and error before deliberately failing open to the normal control attach. I also added a regression test covering that path.

For the repeated retry-classification finding, I filed #1128: reliably narrowing the retry requires cleat to expose a precise controller-seat failure contract. The current fallback remains necessary for the inspect/attach race covered by this issue. The `--take` probe intentionally checks the tokenized boolean flag in cleat's help output.

rjwittams:
Closing unmerged — **wrong layer**, and the fault is in the brief rather than the work. The crew implemented exactly what #1010 asked for.

Robert's call (2026-07-27): seat management is not flotilla behaviour. cleat owns the session, the one-controller rule, and watch mode, so `cleat attach` should itself fall back to watching with a banner naming the holder. Doing it flotilla-side means every cleat consumer reimplements seat detection, fallback, and banner rendering, with the policy living outside the component that owns the seat.

Scope moved to **flotilla-org/cleat#177** (seat management: force-take, watcher banner, attachment identity), extended there to cover the plain automatic-fallback case as well as force-take. Once that lands, flotilla needs no seat logic at all — it just runs `cleat attach`.

PR #1130, opened 2026-07-27
fix: discover adapters inside contained environments

### LINKED ISSUE #1124: Contained crews cannot start: no interior adapter discovery, so no adapter is ever registered
## What happened

Found by the #1117 plumbing assessment (research doc in PR #1120). Two related absences:

- **Interior host discovery is absent.** Provisioning reads interior environment variables but never runs the existing host detectors against the interior `CommandRunner` — even though `DockerEnvironmentRunner` delegates through `docker exec` correctly, so the capability exists and is simply unused.
- **Consequently, interior adapter availability at session start is broken.** The runtime requires an *actually discovered* adapter, but the provisioned environment bag contains no binary assertions inside the container from which to register one.

So a contained vessel has no registered agent adapter and the crew cannot start, regardless of what the image contains.

Related and worth ruling together: **declared-adapter verification at admission is absent** — admission compares workflow requirements against `docker_per_vessel.agent_adapters`, which is an unverified promise the policy author makes about the image. Running the detectors inside the container would let that promise be checked rather than trusted, which is the same gap #1102's smoke check works around from the outside.

## What to build

Run the existing host detectors against the interior runner during provisioning so adapters present in the image are discovered and registered. Then consider whether admission should verify declared adapters against interior discovery instead of trusting the declaration.

## Acceptance criteria

- [ ] Interior discovery runs against the container and registers adapters found there
- [ ] A contained crew whose image provides claude-code/codex starts successfully
- [ ] A declared adapter absent from the image produces a clear failure naming the mismatch, not a silent non-start

## Blocked by

The read-only workspace mount blocks any end-to-end verification (sibling ticket).

rjwittams:
Closes #1124

## Summary

- run the existing host detectors against each provisioned environment’s interior command runner and environment variables
- register agent adapters discovered inside the container for contained terminal sessions
- carry placement-policy adapter declarations into the environment and reject images whose discovered adapters do not match
- destroy containers rejected during discovery, verification, or registration

## Root cause

Provisioning previously copied interior environment variables into an `EnvironmentBag` but never ran host detectors against the interior runner. The resulting bag had no binary assertions, so adapter discovery could never register Codex or Claude Code. The policy’s declared adapters were also discarded before provisioning and could not be verified.

## Validation

### DIFF
--- a/crates/flotilla-controllers/src/reconcilers/vessel.rs
+++ b/crates/flotilla-controllers/src/reconcilers/vessel.rs
@@ -165,6 +165,8 @@ impl Reconciler for VesselReconciler {
             Ok(strategy) => strategy,
             Err(message) => return Ok(VesselDeps::failed(message)),
         };
+        let declared_agent_adapters =
+            placement_policy.spec.docker_per_vessel.as_ref().map(|docker| docker.agent_adapters.clone()).unwrap_or_default();
 
         let (vessel_index, requirement) = match convoy
             .status
@@ -273,6 +275,7 @@ impl Reconciler for VesselReconciler {
                                 docker: Some(DockerEnvironmentSpec {
                                     host_ref: host_ref.clone(),
                                     image: image.clone(),
+                                    declared_agent_adapters: declared_agent_adapters.clone(),
                                     mounts: Vec::new(),
                                     env: env.clone(),
                                 }),
@@ -548,6 +551,7 @@ impl Reconciler for VesselReconciler {
                                 docker: Some(DockerEnvironmentSpec {
                                     host_ref: host_ref.clone(),
                                     image: image.clone(),
+                                    declared_agent_adapters: declared_agent_adapters.clone(),
                                     mounts: has_repositories
                                         .then(|| EnvironmentMount {
                                             source_path: workspace_root.clone(),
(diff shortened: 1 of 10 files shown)


rjwittams:
Fixed the live duplicate path by extracting `run_provisioned_host_detectors` and using it from both `EnvironmentManager` and the daemon controller runtime. The helper seeds the bag with every interior env var before extending it with detector assertions, so `FLOTILLA_ENVIRONMENT_ID` and custom image variables remain visible while Codex/Claude binaries are discovered. Added regression coverage for the core provisioning path, full env preservation, and container destruction when a declared adapter is absent.

rjwittams:
Fixed in `5c3f8417`. Raw interior env vars are now the sole source of `EnvVarSet` assertions; detector results are filtered to non-env assertions before extending the bag. The regression test counts `HOME` assertions and requires exactly one, while still verifying the Codex binary assertion is added.

rjwittams:
Agreed that detector ownership should be unified. The default production behavior is correct in this PR, while retaining an injected/custom detector set requires a broader `DiscoveryRuntime`/`EnvironmentManager` ownership change. Filed #1131 to make that dependency genuinely shared and preserve test/runtime isolation.

PR #1132, opened 2026-07-27
fix: normalize convoy project references

### LINKED ISSUE #1129: convoy start rejects the project name that project list displays
## What happened

`flotilla project list` displays projects in qualified `namespace/name` form (e.g. `flotilla/flotilla`, `flotilla/cleat`). Feeding that exact string back to `flotilla convoy start --project flotilla/flotilla` fails:

```
error: project flotilla/flotilla is not ready: resource not found: flotilla/flotilla
```

Only the bare resource name (`--project flotilla`) resolves. `snapshot_project_repositories` passes `project_ref` straight to `definitions::<Project>(namespace).get(project_ref)`, so the qualified form is looked up as a literal name within the namespace and misses.

The list surface teaches the user a string the start surface rejects. Also seen with the full address form `project/flotilla/flotilla`.

## What to build

Accept the qualified `namespace/name` form (and ideally the full `project/<ns>/<name>` address) in `--project`, splitting namespace from name before lookup — or normalise what `project list` prints to a string every subject-accepting command resolves. Ruling consistently for all subject-reference flags is preferable to special-casing this one.

## Acceptance criteria

- [ ] Every project identifier printed by `project list` is accepted by `convoy start --project`
- [ ] The error for a genuinely unknown project names the forms that were tried

rjwittams:
## Summary

- normalize convoy start project references before admission
- accept both `namespace/name` and `project/namespace/name` identifiers printed by `flotilla project list`
- preserve the canonical bare project resource name in persisted convoy specs and deduplication keys
- report the resolved qualified lookup target when a project is unknown

## Root cause

Convoy admission passed the raw `--project` value directly to a namespace-scoped Project resolver. As a result, `flotilla/flotilla` was treated as a literal resource name inside the `flotilla` namespace instead of being split into namespace and name.

## Validation

Closes #1129

### DIFF
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -3348,6 +3348,32 @@ fn required_admission_value<'a>(value: &'a str, field: &str) -> Result<&'a str, 
     }
 }
 
+fn resolve_project_ref(default_namespace: &str, value: &str) -> Result<(String, String), String> {
+    let value = required_admission_value(value, "project")?;
+    let address_value = value.strip_prefix(flotilla_protocol::view_address::SCHEME_PREFIX).unwrap_or(value);
+    if address_value.starts_with("project/") {
+        return match value.parse::<ViewAddress>() {
+            Ok(ViewAddress::Project { namespace, name }) => Ok((namespace, name)),
+            Ok(_) => unreachable!("project-prefixed addresses only parse as project views"),
+            Err(error) => Err(format!("invalid project reference {value}: {error}")),
+        };
+    }
+    match value.split('/').collect::<Vec<_>>().as_slice() {
+        [name] => Ok((default_namespace.to_string(), (*name).to_string())),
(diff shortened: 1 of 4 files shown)


rjwittams:
Fixed both findings in 9f3913b9.

Convoy-start normalization now lives in `normalize_convoy_start_intent` and is used by both the local command handler and `prepare_remote_convoy_start`. The existing remote host-direct dispatch test now supplies `flotilla/flotilla`, so it exercises the previously bypassed path end to end.

The parser now treats two-segment `project/widgets` as namespace/name while still accepting three-segment `project/project/widgets` as the full address form; a unit test covers both interpretations. Pinned formatting, warnings-as-errors clippy, and the full sandbox-safe workspace suite all pass.

rjwittams:
Acknowledged — no further code changes needed. The conflict-resolution merge retained the shared normalization and both regression tests, and the post-merge local format, clippy, and full workspace gates are green.

PR #1133, opened 2026-07-27
fix: prune peer replicator supervisors after terminal disconnect

### LINKED ISSUE #1014: Prune peer replicator supervisors after permanent disconnect
From PR #1008 review by @claude: https://github.com/flotilla-org/flotilla/pull/1008#issuecomment-5078981941

`PeerReplicatorSupervisors` retains each peer's active generation and cancellation token indefinitely. A newer generation cancels the previous tasks, but permanent peer retirement/disconnect has no corresponding supervisor cleanup hook. Replicators therefore remain retrying or parked on their socket-path source after the peer can no longer return.

PR #1008 makes the no-socket case a cheap, silent watch-channel wait, so this is not blocking that bugfix. It does make the pre-existing lifecycle gap more visible because inbound-only generations now have parked tasks rather than returning before spawn.

## Acceptance

- permanently retired/removed peers cancel their resource replicator tasks and are removed from the supervisor map
- transient transport reconnects, including same-generation reconnects, keep their supervisors alive so replication can heal
- tests distinguish permanent teardown from a reconnectable disconnect

Context: follow-up lifecycle cleanup identified during review of the socket path re-resolution fix.

rjwittams:
## Summary

- add explicit peer connection lifecycle events so the outbound runtime can distinguish terminal teardown from reconnectable transport drops
- cancel and remove resource replicator supervisors only when the terminal event matches the currently tracked peer generation
- preserve same-generation reconnect healing and protect newer generations from stale teardown events
- cover terminal cancellation/removal, stale-generation teardown, unknown peers, and same-generation source refresh

## Root cause

`PeerReplicatorSupervisors` only retired a generation when a newer generation arrived. When a connection-owning task ended permanently, its cancellation token and supervisor map entry remained alive, leaving resource replication tasks retrying or parked indefinitely.

Transient SSH drops and keepalive failures are intentionally excluded from teardown because their owning task continues its reconnect loop.

## Validation

Fixes #1014.

### DIFF
--- a/crates/flotilla-daemon/src/server.rs
+++ b/crates/flotilla-daemon/src/server.rs
@@ -55,6 +55,23 @@ pub(crate) struct PeerConnectedNotice {
     pub resource_socket_path: Option<PathBuf>,
 }
 
+/// Lifecycle event sent from connection sites to the outbound task.
+///
+/// `Connected` drives outbound state sync and starts resource replicators
+/// for the peer's generation. `Disconnected` is sent only from a
+/// connection-owning task that has no retry loop of its own (inbound
+/// `PeerConnection::run`, or the outbound per-target task's terminal
+/// shutdown branch) — never from a transient, still-retrying disconnect —
+/// so the outbound task can cancel and drop that peer's resource
+/// replicators. The generation is checked against the currently tracked
+/// generation before acting, so a stale/displaced connection's belated
+/// teardown can't cancel a newer, already-reconnected generation.
+#[cfg_attr(feature = "test-support", visibility::make(pub))]
+pub(crate) enum PeerConnectionEvent {
+    Connected(PeerConnectedNotice),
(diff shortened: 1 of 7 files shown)


rjwittams:
Confirmed the `ForwardResult::Shutdown` scope: it is returned only when forwarding into the runtime-wide peer-data channel fails because its receiver has closed. That means the peer runtime is already terminating, so retaining the pre-existing early return before `disconnect_peer_and_rebuild` is intentional here; the supervisor event prevents the independently spawned replication tasks from outliving that terminal connection owner.

I’m keeping the integration-level cancellation test as a non-blocking follow-up possibility. The event branch is a direct call into the unit-tested generation guard, while an end-to-end assertion would require exposing supervisor internals or building a resource-watch cancellation probe solely for the test. The current unit tests plus the existing connected-event wiring coverage exercise the behavior without adding a test-only seam.

PR #1141, opened 2026-07-27
fix: enforce convoy free-space admission floor

### LINKED ISSUE #1135: Admission refuses placement on a host below a free-space floor
## Parent

Split from #1113 during its grill (2026-07-27): the free-space floor is admission policy, not lifecycle. It belongs to the deterministic admission floor named by #1115 (Quartermaster v1) — a refusal that needs no judgment, evaluated at dispatch.

## What happened

Six concurrent convoys on one host filled its boot volume (2026-07-26): every process failed with ENOSPC, the daemon wedged permanently (#1111), and unrelated builds broke. Nothing checked free space at dispatch, so the sixth convoy was admitted onto a nearly-full disk exactly as readily as the first.

The lifecycle re-cut (#1113 / ADR 0021) removes the steady-state accumulation; this floor is the backstop against the pathological wave.

## What to build

Placement admission checks the candidate host's free space against a configured floor and refuses with an actionable message (host, free, floor, and what to do: reap, prune, or pick another host). The floor is per-host configuration with a sensible default; the check runs wherever placement is decided, so remote dispatch is covered too.

## Acceptance criteria

- [ ] Dispatch to a host below the floor is refused, naming host, free space, floor, and remediation
- [ ] The floor is configurable per host with a default that would have stopped the 2026-07-26 wave
- [ ] Remote placement (multi-host) enforces the same check as local
- [ ] A refused dispatch leaves no partial convoy state behind

## Blocked by

None — can start immediately. The judgment layer above the floor (wait vs wake a box vs cloud sandbox) is #1115's scope, not this ticket's.

rjwittams:
## Summary

- enforce a 20 GiB default free-space floor on the candidate host before convoy persistence
- configure the floor per host through `[admission].free_space_floor_gib` in `daemon.toml`
- enforce the check on the remote execution daemon before prepared workflow or placement snapshots are written
- return an actionable refusal naming the host, measured free space, configured floor, and remediation options

## Testing

Closes #1135

### DIFF
--- a/crates/flotilla-core/src/admission.rs
+++ b/crates/flotilla-core/src/admission.rs
@@ -1,0 +1,71 @@
+use std::path::Path;
+
+use sysinfo::Disks;
+
+const BYTES_PER_GIB: u64 = 1024 * 1024 * 1024;
+
+pub(crate) fn check_free_space_floor(host: &str, path: &Path, floor_gib: u64) -> Result<(), String> {
+    if floor_gib == 0 {
+        return Ok(());
+    }
+
+    let floor_bytes =
+        floor_gib.checked_mul(BYTES_PER_GIB).ok_or_else(|| format!("free-space floor for host `{host}` is too large: {floor_gib} GiB"))?;
+    let free_bytes = available_space(path)
+        .ok_or_else(|| format!("placement refused on host `{host}`: free space could not be measured for {}", path.display()))?;
+    check_measured_free_space(host, free_bytes, floor_bytes)
+}
+
+fn available_space(path: &Path) -> Option<u64> {
+    let disks = Disks::new_with_refreshed_list();
+    disks
+        .list()
+        .iter()
+        .filter(|disk| path.starts_with(disk.mount_point()))
+        .max_by_key(|disk| disk.mount_point().components().count())
+        .map(|disk| disk.available_space())
+}
+
+fn check_measured_free_space(host: &str, free_bytes: u64, floor_bytes: u64) -> Result<(), String> {
+    if free_bytes >= floor_bytes {
+        return Ok(());
+    }
+
+    Err(format!(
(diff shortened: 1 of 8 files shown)


rjwittams:
Addressed all six findings in `483c58e9`:

1. Fixed: `CommandAction::ConvoyCreate` now runs the same local free-space admission check before workflow resolution or adoption-side effects. Added a regression test proving refusal creates no Convoy. The three convoy-placing entry points (`ConvoyStart`, `ConvoyStartPrepared`, and `ConvoyCreate`) are now all gated.
2. Verified as a real scope limitation of the current policy, not something that can be safely generalized in this PR: checkout destinations are independently configurable per repository and may span volumes, while the host-level floor deliberately probes the stable state-directory volume. The operator docs now explicitly identify that volume as a proxy and require an equivalent external guard when checkout volumes are separate.
3. Fixed: daemon-config loading, disk enumeration, and available-space probing now run in `tokio::task::spawn_blocking`, so slow filesystem/mount inspection does not block a Tokio worker.
4. Confirmed intentional fail-closed behavior: an unavailable capacity measurement must not silently disable a deterministic admission floor. The docs now state that measurement failure refuses placement and explain how an external equivalent guard permits setting the floor to zero.
5. Fixed: displayed free space is truncated (never rounded up), so a below-floor value cannot render as equal to an integer-GiB floor. Added regression coverage for one byte below 20 GiB.
6. Fixed: `ConvoyStart` checks local free space before `prepare_convoy_admission`, avoiding name-generation and other admission work on a host that will refuse placement.

Validation passed locally with the exact repository gates:

rjwittams:
Acknowledged. The follow-up review confirms all six findings are addressed and identifies no remaining blockers; the two observations are intentional, documented tradeoffs and require no further code changes in this PR.

PR #1156, opened 2026-07-27
feat: add credential grants and consumer adapters

### LINKED ISSUE #1139: Enact the credential model: CredentialSpec, grants, consumer adapters, preflight
## Parent

Map #1046 · designed in #1050, recorded as ADR 0022 (`docs/adr/0022-credential-declarations-grants-and-local-material.md`). The ADR is the contract; this ticket is the enactment.

## What to build

1. **`CredentialSpec` declarations + grants** as definitions-class config: name, consumer adapter, source (`file | env | issue-command`), lifecycle (`static | refreshable | issued`), placement requirements; grants bind names to selectors, stance first, refined by project/repo. Default deny.
2. **Host-local credential store**: material resolved from host files/env/commands at provisioning; never enters resources, replication, or snapshots.
3. **Consumer adapters** for the initial set: `gh` (env token), `forgejo` (token file), `claude` (env key), `codex` (login transformation via `codex login --with-api-key`), `docker-registry` (docker login/config). Each adapter: delivery form, transformation, writable-cache handling, and **mandatory preflight** that fails provisioning with a bounded error before the crew reports started.
4. **Admission integration**: held-credential set is an admission fact; a workflow requiring credentials a host lacks is refused early with a naming message.
5. **Contained crews default-deny from day one**; uncontained crews unchanged this ticket (allowlist is follow-on).

## Acceptance criteria

- [ ] A contained crew receives exactly its granted credentials via adapters; nothing ambient
- [ ] Missing/unusable credential → bounded preflight failure naming credential + adapter (never a silently retrying crew)
- [ ] Declarations + grants replicate; no secret material appears in any resource, snapshot, or log
- [ ] Admission refuses a workflow whose grants the host cannot satisfy
- [ ] Codex path proven end to end (login transformation, not env passthrough)

## Blocked by

Operator setup ticket (machine account + tokens on hosts) for the forge identity parts; model-API adapters can proceed immediately.

rjwittams:
## Summary

- add replicated `CredentialSpec` and `CredentialGrant` definitions with stance-first, project/repository-refined selectors
- publish host-local held-credential admission facts and reject placements that cannot satisfy contained workflow grants
- resolve credential material only on the provisioning host and deliver it through mandatory `gh`, Forgejo, Claude, Codex, and Docker Registry adapter preflights
- propagate granted credential references through vessel environments and terminal sessions without placing secret material in resources, replicas, or snapshots
- keep uncontained crew behavior unchanged while making contained credential grants default-deny

## Impact

Contained crews receive only credentials selected by matching grants. Missing declarations, unavailable host material, failed login transformations, and failed authentication probes stop provisioning with bounded errors naming the credential and adapter. Codex credentials are transformed through `codex login --with-api-key` into a writable per-environment `CODEX_HOME` rather than exposed as an API-key environment variable.

## Validation

Closes #1139


rjwittams:
Fixed all three findings in `b38263c7`.

Local `ConvoyStart` and standalone `ConvoyCreate` now pin a content-addressed workflow snapshot after grant resolution, so the reconciler consumes the admitted `credential_refs`. The new `local_convoy_admission_pins_the_grant_resolved_workflow` regression test exercises the ordinary local path end to end.

Docker environment teardown now evicts both cached material and prepared/preflight state for that environment, with a focused eviction test. Docker Registry preparation also preserves the login/pull failure as the primary error when scratch-cache cleanup fails, appending the cleanup failure as secondary context.

rjwittams:
Fixed in `3f93b4e5`.

The Docker Registry adapter now retains its credential-scoped Docker config for the provisioned environment's lifetime and passes that exact path into the provider's actual container creation command. `DockerEnvironmentProvider::create` emits `docker --config <credential-cache> run --pull <policy> ...`, so `Always` and the other pull policies use the authenticated config rather than ambient host auth. Failed provisioning and normal teardown both remove the retained config.

Coverage now proves the config survives login/pull preflight until teardown, proves `docker run --pull always` receives the same explicit config, and exercises grant-resolved snapshot pinning through standalone `ConvoyCreate` as well as `ConvoyStart`.

`Issued` is intentionally minted once per environment and evicted with that environment; an inline comment now distinguishes that lifecycle from per-preparation `Refreshable` resolution and static caching.

rjwittams:
Fixed both findings in `29e49e52`.

The registry-config map entry is now extracted into a local before either filesystem await, so the mutex only guards the map mutation and unrelated environment setup/teardown is not serialized on directory removal.

The retained Docker credential directory is explicitly chmod-equivalent to mode `0700` immediately after creation, before `docker login` can write `config.json`. The lifetime test now asserts the exact permission bits as well as cleanup on teardown.

rjwittams:
Acknowledged. Final re-review confirms the previously reported issues are resolved at `29e49e52` and found no new blocking or follow-up work.

rjwittams:
It feels like some places we are letting very docker specific things into generic areas - eventually we will support other vessel provisioning so we should think about this (probably a followup)

rjwittams:
Agreed; `CreateOpts::docker_config_dir` exposes a Docker-specific mechanism through the generic environment-provider seam. I filed #1160 to deepen that interface for Docker and future vessel providers without broadening this already-reviewed credential feature during its rebase.

rjwittams:
Verified. Normal teardown is covered, but restart loses the in-memory provisioned-environment handle and can orphan the retained mode-0700 registry auth directory. I filed #1161 for restart-safe recovery plus end-to-end landing/reclaim coverage. This remains a bounded follow-up because the underlying handle-rehydration gap predates this PR and the retained directories are access-restricted.

PR #1157, opened 2026-07-27
Attach through transport hops with send keys

### LINKED ISSUE #1125: Prefer send-keys hops over command wrapping for attach chains
## What to build

1. **Purpose-keyed strategy**: attach chains use send-keys (promote `AlwaysSendKeys` from test-only to the attach default); non-interactive command execution keeps wrapping. `CombineStrategy` selection moves from a global default to purpose-derived; the per-hop override seam remains but nothing uses it.
2. **Prove send-keys end to end** through a real docker hop: `flotilla attach` to a contained vessel enters the host shell, types the inner transient attach, lands in the container's crew session. Add the readiness mechanism: where sessions are cleat-managed use cleat `wait`/`expect`; no sleeps, no prompt regexes.
3. **Quoting as a tested concern**: send-keys payload construction covered for names with spaces, quotes, and non-ASCII; tty echo accounted for. Each hop quotes exactly once.
4. **Failure modes**: a hop shell dying mid-chain produces a diagnosable error naming the hop, not a hang.
5. **Docs**: `hop_chain` module docs record the purpose split; correct the #1117 assessment doc's "docker hop wraps" line (it described the selected default, not the architecture). Leave an obvious resolver seam (comment-level) for future depth/length limits — do not implement them.

## Acceptance criteria

- [ ] Attach through ssh→docker chain works via send-keys with no multiply-quoted composite command
- [ ] Non-interactive execution through the same boundaries still wraps
- [ ] Quoting property/unit tests for hostile names pass
- [ ] Mid-chain shell death yields a hop-named error
- [ ] Module docs + assessment correction landed

## Blocked by

Benefits from #1124 (interior discovery, in flight) for a fully-working contained target to attach to; can start with resolution-level + ssh-hop coverage before that lands.

rjwittams:
## What changed

- resolve interactive attach hop chains by entering each SSH or Docker boundary and sending the next transient attach command one hop at a time
- keep non-interactive execution paths fully wrapped through the same purpose-keyed resolver seam
- carry structured attach plans over the protocol and execute multi-action plans through a recorded cleat bridge
- wait for shell screen stability, surface hop-specific failures, and clean up the bridge after attach
- test hostile names with spaces, quotes, and Unicode, recursive `AttachTransient`, readiness/echo behavior, shell death, and the real Docker path (ignored unless its external prerequisites are present)

## Validation

Closes #1125

### DIFF
--- a/crates/flotilla-core/src/executor.rs
+++ b/crates/flotilla-core/src/executor.rs
@@ -16,7 +16,8 @@ use std::{
 use flotilla_protocol::{
     arg::Arg,
     qualified_path::{PathQualifier, QualifiedPath},
-    CheckoutTarget, Command, CommandAction, CommandValue, HostName, NodeId, PreparedWorkspace, ResolvedPaneCommand,
+    CheckoutTarget, Command, CommandAction, CommandValue, HostName, NodeId, PreparedWorkspace, ResolvedAttachAction, ResolvedAttachPlan,
+    ResolvedPaneCommand,
 };
 use tracing::{debug, error, info};
 
@@ -759,7 +760,8 @@ impl StepResolver for ExecutorStepResolver {
                 let cmd = resolve_attach_command(&session_id, self.registry.as_ref(), self.providers_data.as_ref()).await?;
                 // Step-planned attach paths resolve provider sessions directly
                 // and have no resource-level binding to stamp.
-                Ok(StepOutcome::Produced(CommandValue::AttachCommandResolved { command: cmd, binding: None }))
+                let plan = ResolvedAttachPlan(vec![ResolvedAttachAction::Command(vec![Arg::Literal(cmd)])]);
+                Ok(StepOutcome::Produced(CommandValue::AttachCommandResolved { plan, binding: None }))
             }
             StepAction::EnsureCheckoutForTeleport { branch, checkout_key, initial_path } => {
                 if let Some(path) = initial_path {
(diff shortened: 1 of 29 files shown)


rjwittams:
Fixed the orphaned bridge concern in 8211ceef. Active `flotilla-attach-*` bridges are now registered before launch, removed during every normal/error exit, and drained by the panic hook plus SIGINT/SIGTERM handling before the process exits. SIGINT preserves the conventional exit status 130.

The screen-stability behavior is intentional for this ticket: #1125 explicitly chooses cleat readiness primitives over sleeps or prompt regexes. Host-key/password prompts remain transport configuration failures rather than shell prompts we can portably recognize here.

rjwittams:
Fixed in 0931901a. `run_attach_plan` now installs the SIGINT/SIGTERM cleanup handler itself, covering standalone `flotilla attach`, convoy auto-attach, and TUI callers. Installation is guarded by `Once`, so the TUI's earlier startup registration remains useful without creating duplicate signal tasks.

rjwittams:
Fixed in a49b2bbe. Panic-hook installation is now idempotent and intrinsic to `run_attach_plan`, so CLI, convoy auto-attach, and TUI paths share the same panic cleanup. The bridge registry also recovers a poisoned mutex before draining, and both hook doc comments now reflect pre-terminal/idempotent use.

rjwittams:
Verified. No further code change needed: the remaining helper-extraction note is explicitly non-blocking and the two cleanup paths intentionally differ in ownership (`remove` for one bridge versus `drain` for shutdown). Thanks for the fresh full pass.

PR #1162, opened 2026-07-27
feat: split convoy repository source and target refs

rjwittams:
## Summary

- replace each convoy repository's overloaded `base_ref` with distinct `source_ref` and `target_ref` declarations
- use `source_ref` only for checkout provisioning and render `target_ref` in crew work context
- capture the merged change-request base and record advisory per-repository target mismatch facts without blocking the `Landing` → `Landed` transition
- update the Convoy CRD and admission, provisioning, brief, projection, and lifecycle coverage

## Validation

- `cargo +nightly-2026-03-12 fmt --check`
- `cargo clippy --workspace --all-targets --locked -- -D warnings`
- `cargo test --workspace --locked`

Closes #1155

Issue #1155, opened 2026-07-27
split convoy repository source_ref from declared target_ref landing output


rjwittams:
## What to build

\`ConvoyRepositorySpec.base_ref\` currently does silent double duty: provisioning input (where the checkout forks from) and undeclared output (where the change is supposed to land). Split it into two per-repo fields of different kinds:

- **\`source_ref\`** — provisioning input consumed at admission: the ref the checkout/worktree forks from. Today's \`base_ref\` role, renamed.
- **\`target_ref\`** — the convoy's declared landing output: the ref this repo's change lands onto. Bare ref name for now.

Both fill from the project's \`repositories[].default_branch\` (falling back to the repository's observed default branch), and are equal for ordinary implement PRs. No new dispatch flags. No backward compatibility (Plane-B, wire formats free to change).

\`target_ref\` drives three consumers:

1. **Crew brief**: the work-context repository lines render the target (e.g. \`- \`repo\` — url (target \`kitty-image-plumbing\`)\`) so the crew opens the PR with the right base. Fix the fork-stance brief layer wording "the stack branch named in the dispatch inputs" to point at the work context instead (inputs are never rendered into the brief).
2. **Settlement comparison — advisory only**: the Landing→Landed condition stays "no change request outstanding" (ADR 0021, unchanged). At settlement, compare where the change request actually landed (its base at merge) against \`target_ref\`: match is silent; mismatch is recorded as an observed fact on the convoy and surfaced (event emission when the events substrate exists; a recorded status fact now). Never a wedge — a human retargeting and merging a PR elsewhere is landing curation and wins (ADR 0023: observed history is evidence; the declaration is the convoy's claim, not a gate).
3. **Multi-repo**: both fields are per-repository entries, as \`base_ref\` already is.

Explicitly out of scope: DAG/stack declarations, rebase-obligation computation, generation-style rebase convoys (\`<dag-descr>-N/<branch-descr>\`) — those belong to #1150. This issue only establishes the declared per-repo landing target.

## Acceptance criteria

- [ ] \`ConvoyRepositorySpec\` carries \`source_ref\` and \`target_ref\`; all admission paths (project-resolved and direct-repository) populate both; \`base_ref\` is gone
- [ ] Checkout provisioning consumes \`source_ref\`; nothing consumes \`target_ref\` as a provisioning input
- [ ] Crew brief work context renders the per-repo target; fork-stance layer wording updated; brief snapshot/assertion tests updated with justification
[...]

PR #1164, opened 2026-07-27
fix: hold convoy landing on fresh change-request evidence

### LINKED ISSUE #1163: Landing gate trusts stale/absent CR evidence; convoy reaped with open PR
## What happened

Convoy `convoy-repository-ref-split` (#1155): crew pushed its branch, opened PR #1162, and ran `flotilla crew complete` → phase Landing (correct, ADR 0021 claim). The reconciler then condition-wrote **Landed and tore the vessel down while PR #1162 was open** — destroying the warm crew mid-shepherd. Branch survived via teardown's `CommitsPastBase` preservation; the PR survived on GitHub.

## Defect chain

1. `checkout_integration.rs` `inspect_landed`: `gh pr list --head <branch>` returning empty → `landed = True` ("no change request exists for branch"). Conflates *nothing to land* (untouched checkout) with *not yet visible* (PR created moments ago, stale observation). ADR 0023: absence of evidence is not evidence of absence.
2. `DaemonConvoyTeardownRuntime` does not override `ConvoyTeardownRuntime::no_change_request_outstanding`, so the Landing→Landed decision consumes the **cached** checkout `integration.landed` condition via the trait default — which the trait doc explicitly describes as the in-memory-reconciler behavior, with real runtimes expected to re-probe.
3. Result: a race between the periodic integration observation and the Landing evaluation. Any convoy whose crew completes before the post-PR re-observation gets spuriously Landed + torn down.

## Fix

- Daemon override: `no_change_request_outstanding` re-probes each checkout's integration freshly (via `inspect_checkout_integration`) at the decision edge; `Unknown` conditions hold Landing (never release on ignorance).
- `inspect_landed` no-PR case: `True` only when the branch has no commits past its bootstrap ref (`refs/flotilla/bootstrap/<branch>`); if the branch has commits beyond bootstrap and no PR is found → `False` ("branch has commits but no change request found"). Bootstrap-ref-less (adopted) branches keep current behavior.
- Tests for: stale-condition race (fresh probe wins), commits-but-no-PR holds Landing, untouched checkout still lands.

## Evidence

Feta store 2026-07-27 ~20:4xZ: `convoy-repository-ref-split | phase: Landed | work: None`; PR #1162 OPEN; worktree and cleat session gone; branch preserved in clone. Nine sibling convoys in flight exposed to the same race.

rjwittams:
Fixes #1163 — a Landing-phase convoy could be condition-written to Landed and torn down **while its PR was open**, because the decision consumed a cached checkout `integration.landed` observed before the PR existed, and the no-PR case mapped to landed=True unconditionally.

Observed live: `convoy-repository-ref-split` (#1155) — crew pushed, opened PR #1162, completed; convoy landed + vessel reaped ~immediately; branch survived only via teardown's `CommitsPastBase` preservation.

- `DaemonConvoyTeardownRuntime` now overrides `no_change_request_outstanding` to freshly re-probe each checkout's integration at the decision edge (persisting the observation); Unknown/unprobeable holds Landing — absence of evidence never releases the gate.
- `inspect_landed`'s no-PR case returns True only when the branch has no commits past its bootstrap ref; commits-without-visible-PR is now False ("the observation may predate the change request"). Branches without a bootstrap ref (adopted/external) keep the historical answer.
- Unit tests cover: commits-past-bootstrap holds, untouched branch lands, missing bootstrap ref lands, open PR holds, merged PR lands.

Note: `adopted_checkout_with_explicit_transport_applies_fork_stance_config` fails on my machine on clean main too (pre-existing, environment-dependent — green in CI on #1156); not touched here.

### DIFF
--- a/crates/flotilla-core/src/checkout_integration.rs
+++ b/crates/flotilla-core/src/checkout_integration.rs
@@ -234,6 +234,21 @@ async fn inspect_landed(
     base_ref: Option<&str>,
     observed_at: &str,
 ) -> (IntegrationCondition, Option<LandedEvidence>) {
+    // Git evidence first: a branch with no commits beyond its base has
+    // nothing to land, so it counts as landed without consulting the forge at
+    // all. This keeps untouched checkouts landable in environments where `gh`
+    // is absent or unauthenticated.
+    let comparison = compare_branch_to_base(runner, checkout_path, base_ref).await;
+    if let BaseComparison::Counted { base_ref, count: 0 } = &comparison {
+        return (
+            IntegrationCondition::builder()
+                .value(ConditionValue::True)
+                .details(vec![format!("branch has no commits beyond {base_ref}")])
+                .observed_at(observed_at.to_string())
+                .build(),
+            None,
+        );
+    }
     match runner
         .run_output(
             "gh",
@@ -278,7 +293,7 @@ async fn inspect_landed(
                         )
                     }
                 }
-                None => (inspect_no_change_request(runner, checkout_path, base_ref, observed_at).await, None),
+                None => (landed_without_change_request(&comparison, observed_at), None),
             },
             Ok(_) | Err(_) => (
                 IntegrationCondition::builder()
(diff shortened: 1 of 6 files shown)


rjwittams:
Third commit (`ce2a923e`) responds to both open findings — with one deliberate divergence from the suggested fix.

**Finding #2 (latch-before-decision): the suggestion pointed at the real bug, but mirroring the latch would have reintroduced #1163.** Tracing the inconsistency you flagged led to `UpdateIntegration::apply`'s unconditional latch: any persisted `landed=True` — including the *vacuous* True recorded while the branch was untouched — was permanent. That defeated this whole PR: the decision edge would persist a fresh `False` against an open change request, `apply()` would silently upgrade it back to `True`, and the next pass's recent-evidence fast path would land the convoy anyway. Mirroring the latch into the decision would have made that failure immediate instead of one-pass-delayed.

The fix goes the other way: **the latch (store apply and both in-function call sites) now requires `landed_evidence`**. A merged change request doesn't unmerge — evidence-backed True still latches, and the one-pass self-heal you traced still works for it. A vacuous "nothing to land yet" is not a landing fact and now yields to fresh evidence everywhere. The decision deliberately reads the raw fresh observation.

**Finding #3 (wiring coverage): added.** `landing_holds_when_fresh_probe_contradicts_stale_vacuous_landed` is an `InProcessDaemon`-level replay of #1163 through `convoy_change_requests_settled` with an injected runner: stale vacuous True + fresh open-CR probe holds Landing on the probing pass, persists `False` un-latched, and still holds on the following in-TTL cached pass. Two store-level tests pin both latch directions (vacuous yields, evidence latches).

rjwittams:
**v4: rebuilt on main** (6d8c94ab) rather than rebased — #1162 and #1168 merged into the same integration code, and main's `base_ref`-aware inspection (`baseRefName` evidence, rev-list counting) superseded the probe shape from v1–v3. The branch now carries only what main still lacks:

1. **Evidence-only latch** (`checkout.rs`): `landed=True` latches in the store only when `landed_evidence` is present. This was the root defect — a vacuous True ("nothing to land yet") latched permanently and defeated every probe-side fix.
2. **Git-first untouched check** (`checkout_integration.rs`): the base comparison runs before any forge call; zero commits beyond base answers True git-only (no gh dependency), and a missing change request with real commits beyond base is False. Absorbed main's `inspect_no_change_request` into a single `BaseComparison`-driven flow.
3. **Decision-edge freshness** (`in_process.rs` + `runtime.rs`): Landing→Landed re-probes integration state older than a 30s `LANDING_EVIDENCE_TTL` via `convoy_change_requests_settled`, instead of consuming conditions cached up to 6h. Adopted checkouts stay exempt, mirroring the teardown exemption.

Two of main's #1168 teardown tests were updated for the new shape: the gh lookup now requests `baseRefName`, and the untouched side of the multi-repo test lands git-only without reaching gh (its mock now reflects a squash-merged side A with commits beyond base).

Full gates on the rebuild: fmt clean, clippy clean, workspace tests green except the three pre-existing env-dependent failures tracked in #1172.

rjwittams:
On the non-blocking ordering observation ([review](https://github.com/flotilla-org/flotilla/pull/1164#issuecomment-5096882350)): agreed it's a real cost asymmetry, keeping the git-first order deliberately. The gh-first variant makes the untouched-branch answer depend on the forge again, and that's precisely the path that has to stay forge-independent — the #1163 hot-loop variant was an adopted checkout with no forge access wedging on `Landed=Unknown`. An open-PR steady state costs one or two extra local `git rev-list`/`rev-parse` spawns per probe, and probe cadence is already bounded (30s `LANDING_EVIDENCE_TTL` at the decision edge, the checkout reconciler's refresh TTL elsewhere). If multi-host refresh-tick cost ever shows up in practice, reordering per-transport (git-first locally, gh-first over peer transports) would be the shape to reach for rather than giving up the git-only property globally.

PR #1168, opened 2026-07-27
fix: allow untouched checkouts through teardown gate

### LINKED ISSUE #1154: teardown Landed gate false-positives on untouched multi-repo checkouts
## Observed

Multi-repo convoy sidebar-region-stack (flotilla + andamento checkouts) completed with all real work on the andamento side (andamento#44, merged). \`flotilla convoy delete\` refused: \`checkout-sidebar-region-stack-flotilla [...]: Landed=False (no PR found for branch)\` — but that checkout had **zero commits over origin/main and a clean tree**. The branch existing is not evidence of unlanded work. Required \`--force\` after manual verification.

## Fix

The teardown integration gate should treat a checkout whose branch has no commits beyond its base ref and a clean worktree as trivially landable (nothing to land), instead of demanding a PR for every branch. Keeps the gate meaningful for checkouts that actually diverged.

## Acceptance criteria

- [ ] Untouched checkout (branch == base, clean) passes the Landed gate without \`--force\`
- [ ] Dirty or diverged checkout without a PR still refuses
- [ ] Test covering the multi-repo one-sided-work teardown case

rjwittams:
Closes #1154.

## What changed

- Treat a checkout with no change request as landed only when `git rev-list <base>..HEAD` reports zero commits.
- Use the checkout's declared base ref for managed worktrees and fresh clones, with `origin/HEAD` fallback for observed non-main checkouts.
- Keep divergent no-PR branches blocked, reporting their commit count in the `Landed` condition.
- Cover a two-repository convoy where one checkout has a merged PR and the other is untouched, plus a divergent no-PR refusal.

## Why

A branch name alone is not evidence of unlanded work. Multi-repository convoys can legitimately do work in only one checkout; the untouched checkout should not force deletion to use `--force`. At the same time, a fully pushed divergent branch without a PR must remain protected by the teardown gate.

## Validation

### DIFF
--- a/crates/flotilla-core/src/checkout_integration.rs
+++ b/crates/flotilla-core/src/checkout_integration.rs
@@ -30,11 +30,20 @@ impl fmt::Display for EmbeddedRepository {
     }
 }
 
-pub fn checkout_branch_from_spec(spec: &CheckoutSpec) -> &str {
+fn checkout_branch_from_spec(spec: &CheckoutSpec) -> &str {
     match spec {
         CheckoutSpec::Worktree(spec) => &spec.r#ref,
         CheckoutSpec::FreshClone(spec) => &spec.r#ref,
         CheckoutSpec::Observed(spec) => &spec.r#ref,
+    }
+}
+
+fn checkout_base_ref_from_spec(spec: &CheckoutSpec) -> Option<&str> {
+    match spec {
+        CheckoutSpec::Worktree(spec) => spec.base_ref.as_deref(),
+        CheckoutSpec::FreshClone(spec) => spec.base_ref.as_deref(),
+        CheckoutSpec::Observed(spec) if spec.is_main => Some(&spec.r#ref),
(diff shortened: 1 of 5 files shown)


rjwittams:
Addressed the ADR mismatch in 694163b7: the no-change-request case now records that `Landed` requires zero commits beyond the checkout's base ref, while divergent no-CR branches remain unsettled.

I left the origin-HEAD fallback local for now. It has only two call sites with different error context, so extracting a shared helper would not yet add meaningful leverage.

rjwittams:
Thanks — no code change for the non-blocking subprocess suggestion. `<ref>..HEAD` is not a ref compared with itself: the left operand is the named base ref while the right operand is the checkout's current commit. They normally coincide for an observed main checkout, but retaining the probe detects stale or detached divergence instead of assuming the observed spec and working tree still agree at teardown time.

PR #1169, opened 2026-07-27
fix: expose follower provider suppression

### LINKED ISSUE #903: fix: GitHub providers silently fail to activate on feta despite satisfied gates
## What to build

Root cause is established (see comments): `follower = true` in `daemon.toml` selects `FactoryRegistry::for_follower()`, which intentionally registers no change_request, issue_tracker, cloud_agent, or ai_utility factories. The remaining work is **visibility — no silent third state**:

1. **Provider listings** (`flotilla host <name> providers`, `flotilla repo <name> providers`): categories suppressed by follower mode appear explicitly as `disabled: follower mode` rows (or an equivalent labelled section), not omitted.
2. **Admission errors**: when convoy admission fails to find an issue provider (or CR tracker) on a follower daemon, the error names the cause and the remedy: this host runs in follower mode; dispatch from a leader host. Keep the existing error for genuinely-missing providers on non-follower daemons.
3. Every factory outcome remains one of: provider active, named unmet requirement, or labelled suppression. Nothing silently absent.

Do not change what follower mode suppresses — retirement criteria are a separate thread on map #1046.

## Acceptance criteria

- [ ] `flotilla host <follower> providers` shows the four suppressed categories labelled `disabled: follower mode`
- [ ] Issue-provider admission failure on a follower daemon says follower mode is the cause and points at leader dispatch
- [ ] Non-follower behavior unchanged (test both modes)
- [ ] CI gates pass (pinned fmt, clippy -D warnings, workspace tests)

## Blocked by

None — can start immediately.

rjwittams:
## Summary

- represent follower-mode factory suppression explicitly for change requests, issue trackers, cloud agents, and AI utilities
- show suppressed categories as `disabled: follower mode` in host and repository provider listings
- explain follower mode and leader dispatch when convoy issue admission cannot resolve an issue provider
- preserve the existing missing-provider error and provider listings on leader daemons

## Root cause

Follower daemons intentionally use `FactoryRegistry::for_follower()`, which omits external-provider factories. Provider queries previously represented only active factories and unmet requirements, leaving intentional suppression as a silent third state.

## Validation

Closes #903

### DIFF
--- a/crates/flotilla-core/src/convert.rs
+++ b/crates/flotilla-core/src/convert.rs
@@ -265,6 +265,7 @@ pub fn provider_health_to_host_statuses(health: &HashMap<(&'static str, String),
             // keys, but will break if a display name diverges from its key.
             implementation: name.to_lowercase(),
             healthy: *healthy,
+            disabled_reason: None,
         })
         .collect();
     statuses.sort_by(|a, b| a.category.cmp(&b.category).then_with(|| a.name.cmp(&b.name)));
@@ -392,13 +393,15 @@ mod tests {
             category: "vcs".into(),
             name: "Git".into(),
             implementation: "git".into(),
-            healthy: true
+            healthy: true,
+            disabled_reason: None,
         }));
         assert!(statuses.contains(&HostProviderStatus {
             category: "cloud_agent".into(),
             name: "Claude".into(),
             implementation: "claude".into(),
-            healthy: false
+            healthy: false,
+            disabled_reason: None,
         }));
     }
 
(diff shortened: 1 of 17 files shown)


rjwittams:
Fixed both findings in 7c35bfc7. The TUI Hosts panel now renders suppressed providers as `(disabled: follower mode)` instead of a healthy checkmark, with a focused regression test. Follower detection now compares suppression membership independently of vector order, also covered by a regression test. Pinned format, clippy with warnings denied, and the full workspace test suite all pass after the changes.

rjwittams:
Thanks for the re-review. The remaining helper-extraction note is acknowledged as non-blocking; the two call sites are small and operate on different protocol row types, so I am leaving them local in this focused fix.

PR #1171, opened 2026-07-27
Structured daemon logging and mesh-aware logs command

### LINKED ISSUE #1145: Structured daemon logging: jsonl files, rotation, and a mesh-aware logs verb
## Parent

Enactment of #1051 Q1 (see its resolution comment).

## What to build

1. **JSON-lines file logging** on flotillad via the existing `tracing` machinery: `~/.local/state/flotilla/log/flotillad.jsonl`, size-based rotation keeping N generations, host-local, never replicated. Structured fields per the existing convention; per-subsystem module-path targets; level directives configurable per target (`RUST_LOG`-style in config). Runtime level adjustment deferred.
2. **`flotilla logs` verb**: `--host <peer> --since <dur> --level <lvl> --target <mod>` streams a peer's jsonl through the existing mesh connection. Tail/filter passthrough only — jq is the analysis tool once lines arrive.
3. Stderr behaviour unchanged for interactive runs; file layer is additive.

## Acceptance criteria

- [ ] Daemon writes rotating jsonl with structured fields; rotation bounded in size and generations
- [ ] `flotilla logs --host feta --since 2h --level warn` returns filtered lines from the remote daemon without ssh
- [ ] Per-target level directives honored from config
- [ ] Nothing replicates log content; retrieval is on-demand streaming

## Out of scope (recorded in #1051)

Metrics/dashboards/OTLP; centralisation and collection of other log families (crew sessions, cleat recordings, transcripts) — important, deferred; object-store archival (#748 / ADR 0021 extract stage).

rjwittams:
## What changed

- write daemon tracing events as structured JSON-lines at `~/.local/state/flotilla/log/flotillad.jsonl`
- rotate log files by configurable size and retained generation count
- support `RUST_LOG`-style daemon filter directives in `daemon.toml`, including per-target levels
- add `flotilla logs --host <peer> --since <duration> --level <level> --target <module>`
- route log reads on demand over the existing peer command path while keeping log content out of replicated stores
- preserve interactive stderr logging as an additive tracing layer

## Validation

- real-process smoke test verifying JSONL output and filtering through `flotilla logs | jq`

Closes #1145


rjwittams:
Fixed both substantive findings in c38726cc. The `--since` cutoff now uses `checked_sub_signed` and returns a normal error for out-of-range wire values, with a regression test covering the reported 9,000,000,000,000-second case. The interactive formatter now explicitly writes to stderr.

I also changed the command summary from “Stream” to “Read” so it does not imply follow mode, and documented that logging configuration is applied at daemon startup, including rotation bounds.

rjwittams:
Addressed the blocking-I/O observation in e3d9baa5: the retained-file scan now runs in `tokio::task::spawn_blocking`, so an on-demand local or peer log query cannot occupy a Tokio worker while reading the bounded generations. I deliberately left the selected result uncapped so the command does not silently omit matching JSONL; it remains a one-shot request/response transfer rather than follow mode or replicated log state.

rjwittams:
Addressed in f931a7a1: the human-readable tracing layer is now installed only when `stderr.is_terminal()`. Direct interactive `flotillad` runs keep their stderr output, while detached daemons no longer mirror the structured stream into the unrotated `daemon-panic.log`; Rust panics and pre-init errors still use that redirected stderr file. Verified with the exact format check, workspace Clippy (`-D warnings`), and the full `flotilla-daemon` test suite.

rjwittams:
CI follow-up for the `main` refresh: merging #1168 exposed stale exact-command fixtures in its two new teardown tests. The shared integration query had gained `baseRefName` in #1162, but those mocks still registered the older `gh pr list --json number,state,mergedAt` shape, producing `Landed=Unknown`. Commit 7919179a updates only the fixtures (including the merged PR target branch); the focused convoy-delete loop and full workspace suite are green. No production teardown behavior was changed.

PR #1178, opened 2026-07-27
fix: isolate admission tests from host environment

### LINKED ISSUE #1172: three daemon/core tests are host-environment-dependent and fail on macOS desks
## What happened

Running `cargo test --workspace --locked` on a macOS desk (kiwi) fails three tests that are green in CI:

- `in_process::tests::adopted_checkout_with_explicit_transport_applies_fork_stance_config` (flotilla-core) — `assertion failed: inspection.spec.is_fork()`
- `runtime::tests::provisioned_environment_discovers_and_registers_interior_agent_adapters` (flotilla-daemon)
- `runtime::tests::provisioned_environment_is_destroyed_when_declared_adapter_is_missing` (flotilla-daemon)

The latter two panic at daemon init: `failed to resolve local node id: "Cannot determine machine identity. Set machine_id in daemon.toml."` — machine identity resolution reaches the real host (works on CI Linux via /etc/machine-id, fails on macOS where the probe path differs / the test's mock runner has no response). All three reproduce on clean main (verified 2026-07-27 at d3e32e0f).

## Why it matters

Testing philosophy says behavior tests run through injected collaborators, not real host state. Env-dependent failures train people to ignore local red, which nearly masked a real CI failure during the #1164 emergency fix tonight (fail-fast also stopped the workspace run at the first failing crate, hiding the daemon suite entirely — worth remembering `--no-fail-fast` when triaging).

## What to build

- Inject machine-identity resolution in the two provisioned-environment tests (config `machine_id` or a mock response), so they pass anywhere.
- Diagnose the fork-stance inspection test's host dependence and inject it away.
- All three green on a mac desk with no daemon.toml machine_id and no network.

### LINKED ISSUE #1175: Admission disk-floor probe makes ~19 workspace tests fail on low-disk machines
## What happened

`cargo test --workspace` on kiwi (13-15 GiB free at the time) failed 19 tests across `flotilla-core` (8) and `flotilla-daemon` (11) with:

```
placement refused on host `kiwi`: 15.6 GiB free is below the 20.0 GiB floor; reap settled convoys, run scripts/prune-target.sh, or pick another host
```

Every convoy-admission-path test (`convoy_admission_snapshots_every_project_repository`, `local_convoy_admission_pins_the_grant_resolved_workflow`, `convoy_create_*`, `crew_provisioning_*`, `sqlite_adopted_checkout_flow_*`, ...) reaches `check_free_space_floor` → `measure_available_space` (`crates/flotilla-core/src/admission.rs`), which reads the **real** disk via `sysinfo::Disks`. On machines above the floor (CI runners, feta) everything passes; below it, a fifth of the admission suite fails for reasons unrelated to the code under test. This cost a real debugging cycle during a PR rebase — the failures looked like a bad merge until the error text was read closely.

Introduced by #1141 (free-space admission floor). The floor itself is working as designed at runtime — this is purely about tests.

## What should change

Route the disk measurement through an injected collaborator (same pattern as `CommandRunner`/`EnvVars` per the discovery guidance in CLAUDE.md), so `InProcessDaemon`-level tests inject a fixed free-space value and only `admission.rs`'s own unit tests (plus perhaps one real-backed contract test) touch `sysinfo`. Acceptance:

- [ ] `cargo test --workspace` passes on a machine with < 20 GiB free
- [ ] Admission-refusal behavior (below floor, at floor, unmeasurable) covered via the injected value
- [ ] Runtime heartbeat's `disk_free_bytes` still reports the real measurement

Related: #1172 tracks other env-dependent test failures (machine identity, fork stance); this is the same disease, different organ.

rjwittams:
## Summary

- route convoy admission disk-floor measurement through an injected collaborator
- give test discovery runtimes a deterministic 100 GiB capacity while keeping process discovery real-backed
- cover below-floor, exactly-at-floor, and unmeasurable admission outcomes with injected values
- canonicalize the adopted-checkout fixture path so macOS `/var` aliases cannot hide fork configuration
- provide explicit machine identities to provisioned-environment tests

## Root cause

Admission used `sysinfo::Disks` directly, so otherwise deterministic daemon tests failed whenever the host had less than the configured 20 GiB floor. Two provisioned-environment tests also relied on ambient machine identity, and the adopted-checkout test wrote configuration under a path that canonicalizes differently on macOS.

The runtime heartbeat still calls the real `measure_available_space` function and continues reporting actual `disk_free_bytes`.

## Validation

Closes #1175
Closes #1172

### DIFF
--- a/crates/flotilla-core/src/admission.rs
+++ b/crates/flotilla-core/src/admission.rs
@@ -1,29 +1,46 @@
-use std::path::Path;
+use std::{path::Path, sync::Arc};
 
 use sysinfo::Disks;
 
 const BYTES_PER_GIB: u64 = 1024 * 1024 * 1024;
 
-pub(crate) fn check_free_space_floor(host: &str, path: &Path, floor_gib: u64) -> Result<(), String> {
+pub(crate) trait AvailableSpaceProbe: Send + Sync {
+    fn measure(&self, path: &Path) -> Option<u64>;
+}
+
+struct SystemAvailableSpaceProbe;
+
+impl AvailableSpaceProbe for SystemAvailableSpaceProbe {
+    fn measure(&self, path: &Path) -> Option<u64> {
+        let disks = Disks::new_with_refreshed_list();
+        disks
+            .list()
+            .iter()
+            .filter(|disk| path.starts_with(disk.mount_point()))
+            .max_by_key(|disk| disk.mount_point().components().count())
+            .map(|disk| disk.available_space())
+    }
+}
+
+pub(crate) fn system_available_space_probe() -> Arc<dyn AvailableSpaceProbe> {
+    Arc::new(SystemAvailableSpaceProbe)
+}
+
+pub(crate) fn check_free_space_floor(probe: &dyn AvailableSpaceProbe, host: &str, path: &Path, floor_gib: u64) -> Result<(), String> {
     if floor_gib == 0 {
         return Ok(());
     }
 
     let floor_bytes =
         floor_gib.checked_mul(BYTES_PER_GIB).ok_or_else(|| format!("free-space floor for host `{host}` is too large: {floor_gib} GiB"))?;
-    let free_bytes = measure_available_space(path)
+    let free_bytes = probe
+        .measure(path)
         .ok_or_else(|| format!("placement refused on host `{host}`: free space could not be measured for {}", path.display()))?;
     check_measured_free_space(host, free_bytes, floor_bytes)
 }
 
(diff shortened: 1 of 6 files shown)


rjwittams:
Review-cycle summary: verified the automated review against the current diff. No code changes are needed; both observations are explicitly non-blocking, and the duplicated setup remains local to two tests while the stateless system probe construction is intentional.

rjwittams:
Evaluated against the current diff. No change is needed: both observations are explicitly non-blocking, the setup duplication is limited to two focused tests, and constructing the stateless system probe directly keeps the public heartbeat measurement real-backed.

PR #1179, opened 2026-07-27
fix: preserve remote host self reports in fleet health

### LINKED ISSUE #1177: Fleet view shows '-' for remote hosts' self-reported health in a real mesh
## What happened

First live run of `flotilla fleet` (#1173, deployed `fa0bfbf8` on kiwi + feta) on the 3-host mesh:

- kiwi (local row): full data — version, generation, uptime, disk, staleness `current`, diagnosis `agree`.
- feta: heartbeat and replica sync fresh (seconds old), Replica Gen correct — but Version/Daemon Gen/Uptime/Disk all `-` and diagnosis `unknown`.
- udder: same shape (expected — it still runs a pre-#1173 daemon that publishes no health fields).

feta is *not* expected: its daemon self-reports everything locally. `flotilla resource list hosts` **on feta** shows its own Host status with `daemon_version: 0.1.0`, `daemon_generation: 06933366…`, `disk_free_bytes: 459371896832` and a fresh heartbeat. The same Host resource read **from kiwi** has the fresh `heartbeat_at` but none of the health fields.

So the peer's self-reported status fields don't reach the aggregating host — the local copy of a remote Host resource gets its heartbeat maintained (peer-ping path) while version/generation/disk exist only in the origin store, and the fleet aggregation reads the local copy. The PR's review flagged exactly this risk: the aggregation was tested against `InMemoryBackend` only, never end-to-end across real sqlite-backed daemons.

Symmetric confirmation: feta's own list shows the remote host row with `ver: None` and a fresh heartbeat too.

## Expected

The fleet view's whole premise (#1146) is surfacing independent observations without collapsing them — the remote host's *self-report* (via replica) and the local *observation* (link state, heartbeat receipt) should both be present, letting the Diagnosis column actually compare them (`agree`/`disagree`) instead of degenerating to `unknown` for every remote host.

## Notes

- Reproduced immediately after both daemons were bounced onto `fa0bfbf8` with a fresh feta store, so this is not stale-format residue.
- udder needs a separate bump before it can participate at all; its `-` row is correct behavior for a pre-#1173 daemon.

rjwittams:
## Summary

- identify each replicated Host self-report by the origin node's canonical host environment
- prevent a fresher peer-observation Host from replacing the origin daemon's version, generation, uptime, heartbeat, and disk status
- add an end-to-end regression using two SQLite-backed daemons and the real resource replication topology

## Root cause

Each daemon replicates every Host resource it observes. The fleet-health aggregator grouped all replicated Host resources by replica origin, so a remote daemon's fresher local observation of another host could replace that daemon's own Host status. In a two-host mesh, feta's observation of kiwi therefore won over feta's own self-report and left feta's health columns empty on kiwi.

The aggregator now accepts a replicated Host status only when the resource name matches the origin node's canonical host environment ID. Link state remains the local observation, while health fields come from the remote daemon's self-report.

## Validation

- `cargo test -p flotilla-core --locked fleet_health_keeps_link_heartbeat_and_generation_disagreements_visible -- --nocapture`
- `cargo test -p flotilla-daemon --locked --features test-support --test overlay_replication sqlite_daemons_expose_remote_host_self_report_in_fleet_health -- --nocapture`
- `cargo test --workspace --locked`: all tests reached in the run passed except `in_process::tests::adopted_checkout_with_explicit_transport_applies_fork_stance_config`; the identical failure reproduces on untouched base commit `fa0bfbf8`

Closes #1177

### DIFF
--- a/crates/flotilla-core/src/host_registry.rs
+++ b/crates/flotilla-core/src/host_registry.rs
@@ -125,7 +125,6 @@ impl HostRegistry {
         HostListResponse { hosts: host_entries }
     }
 
-    #[cfg(test)]
     pub(crate) async fn environment_id_for_node(&self, node_id: &NodeId) -> Option<EnvironmentId> {
         self.node_environments.read().await.get(node_id).cloned()
     }
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -5208,7 +5208,19 @@ impl InProcessDaemon {
                 }
                 ResourceProvenance::Local => None,
                 ResourceProvenance::Replica { origin_root, .. } => {
-                    self.host_registry.host_name_for_node(origin_root).await.or_else(|| configured_by_node.get(origin_root).cloned())
+                    // An origin replicates every Host it observes, including this daemon's Host.
+                    // Only the Host matching the origin's canonical environment is its self-report.
+                    let is_self_report = self
+                        .host_registry
+                        .environment_id_for_node(origin_root)
+                        .await
+                        .and_then(|environment_id| environment_id.host_id().map(ToString::to_string))
+                        .is_some_and(|host_id| host_id == resource_host.object.metadata.name);
+                    if !is_self_report {
+                        None
+                    } else {
+                        self.host_registry.host_name_for_node(origin_root).await.or_else(|| configured_by_node.get(origin_root).cloned())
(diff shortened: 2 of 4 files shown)


rjwittams:
Thanks for verifying the origin/self-report identity invariant and the SQLite-backed regression. No code changes are needed from this review.

PR #1180, opened 2026-07-27
fix: recover dropped reconcile actuations

### LINKED ISSUE #1165: refused convoy reclaim hot-loops with no backoff, starving the daemon
## What happened

After `convoy-repository-ref-split` was spuriously Landed (#1163), automatic reclaim entered a hot loop: `verify_convoy_teardown_gate` probed the already-removed worktree path, got `Clean=Unknown (git status could not run: No such file or directory)`, refused — and retried every ~4ms, ~250/sec, for over an hour. Effects: 10MB log rotation every ~90s (all 4 retained rotations were spin, destroying the evidence of an unrelated concurrent failure), sustained CPU burn, and the reconcile loop starved badly enough that 9 freshly dispatched convoys' provisioning actuations were lost (#1166).

## What to build

1. **Backoff**: a refused reclaim retries on a backoff schedule (or only on resync/change events), never in a tight loop. Repeated identical refusals log once with a counter, not per attempt.
2. **Idempotent gate for missing paths**: a checkout whose worktree path no longer exists is *already removed* from the integration-safety perspective — the gate should distinguish "cannot probe (hold)" from "nothing left to probe (proceed with reclaim of the remaining resources)". Teardown half-done must be completable, not permanently refused.
3. Force-deleting the convoy also left a ghost `checkout-*` resource behind (Ready, orphaned) — reclaim should remove child resources even when the gate is skipped.

## Evidence

Feta daemon.log rotations 20:21–20:27 local 2026-07-27; loop visible as consecutive identical WARNs 4-5ms apart; stopped immediately by `convoy delete --force`.

### LINKED ISSUE #1166: vessel provisioning actuations lost under load are never re-driven
## What happened

Nine convoys dispatched 2026-07-27 ~19:14Z (while the daemon was in the #1165 reclaim hot-loop) all wedged identically: convoy Active, work Ready, crew Pending, vessel **Provisioning** with no message, checkout resources created but **never reconciled** (no status at all), zero terminal sessions — for over an hour, across many 60s resync cycles, with the liveness watchdog reporting all 14 supervisory tasks healthy. Breaking the hot loop did not recover them. **Restarting the daemon recovered all nine within seconds** — provisioning re-drove from spec and every crew launched.

## Problem

Provisioning actuation is effectively edge-triggered: if the initial actuation is lost (here: starved/dropped during the reconciler spin), resync passes observe "Provisioning, waiting on checkout" and re-issue nothing. A reconciler must converge from any observed state: a vessel in Provisioning whose expected child actuations are absent should re-issue them (or mark failed), not wait forever.

## What to build

- Vessel/checkout reconcile passes detect missing-but-expected actuations (checkout with no status after a deadline, session refs with no session) and re-drive them.
- A stuck-Provisioning deadline surfaces in vessel status (`message`) rather than silence.
- InProcessDaemon-level test: drop the first actuation, assert resync recovers without restart.

Related: #1153 (restart-related liveness), #1165 (the starvation source in this incident).

rjwittams:
## Summary

- honor reconciler `requeue_after` outcomes so statusless checkout provisioning is retried after a dropped completion
- keep provisioning vessels convergent by retrying missing child resources and surfacing a stalled-status message after the deadline
- make reclaim verification idempotent for missing checkout paths and unchanged integration observations, while suppressing repeated identical refusal warnings

## Root cause

The controller loop ignored `ReconcileOutcome::requeue_after`, making provisioning effectively edge-triggered when an initial completion was lost. Reclaim verification also rewrote timestamp-only status on every refusal, feeding its own watch loop and repeatedly logging the same error.

## Validation

- focused controller, resource, core, and daemon regression tests
- `TMPDIR="$PWD/.codex-tmp" cargo test --workspace --locked` (all changed-area tests pass; two pre-existing macOS-only daemon fixtures fail because their fake runner cannot resolve `IOPlatformUUID`, while Linux CI reads `/etc/machine-id`)

Closes #1165
Closes #1166

### DIFF
--- a/crates/flotilla-controllers/src/reconcilers/checkout.rs
+++ b/crates/flotilla-controllers/src/reconcilers/checkout.rs
@@ -10,6 +10,7 @@ use flotilla_resources::{
 use tracing::warn;
 
 const CHECKOUT_INTEGRATION_REFRESH_AFTER: Duration = Duration::from_secs(6 * 60 * 60);
+const CHECKOUT_PROVISIONING_REQUEUE_AFTER: Duration = Duration::from_secs(1);
 
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum CheckoutRemoval {
@@ -187,7 +188,13 @@ where
             None
         };
 
-        ReconcileOutcome::new(patch)
+        let mut outcome = ReconcileOutcome::new(patch);
+        if obj.status.as_ref().map(|status| status.phase).unwrap_or(CheckoutPhase::Pending) == CheckoutPhase::Pending
+            && !matches!(deps, CheckoutDeps::Failed(_))
+        {
+            outcome.requeue_after = Some(CHECKOUT_PROVISIONING_REQUEUE_AFTER);
+        }
+        outcome
     }
 
     async fn run_finalizer(&self, obj: &ResourceObject<Self::Resource>) -> Result<(), ResourceError> {
--- a/crates/flotilla-controllers/src/reconcilers/vessel.rs
+++ b/crates/flotilla-controllers/src/reconcilers/vessel.rs
@@ -131,12 +131,11 @@ impl VesselDeps {
     }
 
     fn provisioning(
-        obj: &ResourceObject<Vessel>,
         placement_policy: &ResourceObject<PlacementPolicy>,
         waiting_for: impl Into<String>,
         actuations: Vec<Actuation>,
     ) -> Self {
-        Self { patch: provisioning_patch(obj, placement_policy, waiting_for.into()), actuations }
+        Self { patch: provisioning_patch(placement_policy, waiting_for.into()), actuations }
     }
 
     fn ready(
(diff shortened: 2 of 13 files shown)


rjwittams:
Addressed all three polish points in `374714fa`: removed the obsolete vessel parameter, pluralized multi-checkout waiting messages, and sorted reclaim refusals before formatting so repeated multi-checkout failures deduplicate deterministically. The backend's semantic no-op comparison remains the write-suppression boundary because the stuck message is time-dependent and must become visible once its deadline passes. Focused regressions and the exact workspace clippy command pass locally.

rjwittams:
Fixed in `81013f25`. The real checkout runtime now treats redrive as an ensure operation: an existing worktree is reused only when it belongs to the expected clone and ref, and an existing fresh clone is reused only when its origin and branch match. Both paths reconstruct the prepared commit/provenance instead of rerunning `git worktree add` or `git clone`. Added real-git point regressions for lost completion after each operation's on-disk side effects; the checkout-runtime group, live controller dropped-completion test, formatting, and exact workspace clippy all pass locally. I left the extra post-success requeue intact because it is what permits recovery when the completion write itself is the dropped edge; the subsequent Ready pass is harmless and terminates the retry.

rjwittams:
Verified the final pass and no further code changes are needed. The optional direct tracing-log assertion is intentionally left out: the point regression already proves identical reclaim observations do not rewrite status/retrigger the watch loop, while building additional log-capture machinery would expand beyond the incident behavior and the assignment's point-test scope.

PR #1182, opened 2026-07-28
Recover crew sessions after daemon restart

### LINKED ISSUE #1153: daemon restart leaves crew work Running with no live session
## Observed

Feta daemon restarted at 2026-07-27T14:04Z. Three active convoys (attach-send-keys-strategy, convoy-lifecycle-landing-landed, credential-spec-grants-adapters) had their work restarted by the daemon (work.ready_at/started_at stamped 14:04:36-37), but their crew terminal sessions never came back: \`terminal_session_refs\` pointed at cleat sessions that did not exist (\`cleat list\` showed none of them). Work state remained \`Running\` for ~4 hours with no live process behind it; \`flotilla ls\` showed crew \`stopped\` while the convoy said \`Active\`/\`Running\`.

All three worktrees held substantial uncommitted mid-implementation changes. Recovery was manual per the #1151 tier-2 ladder: relaunch \`codex --yolo\` in each vessel via cleat under the original terminal session name with a resume brief.

## Problems

1. **False liveness**: work \`Running\` must be contradicted by the observed fact that the session backing it is gone. A liveness condition (session-exists / process-alive) should flip the work to an interrupted/needs-recovery state visible in \`flotilla ls\` and the convoy phase surface.
2. **Restart half-recovers**: the daemon restart re-provisioned/restarted work state but did not relaunch the agent sessions (or the relaunch failed silently). Either relaunch with the #1151 recovery ladder (native resume via hook-recorded session ids where possible, fresh start with crew launch flags otherwise) or don't claim Running.

## Acceptance criteria

- [ ] Work backed by a missing/dead terminal session is not reported \`Running\`; a distinct recoverable state is surfaced
- [ ] Daemon restart either restores crew sessions or surfaces the interruption for recovery
- [ ] InProcessDaemon-level test covering restart-with-lost-sessions

rjwittams:
## Summary

- add explicit interrupted phases for convoys, work, crew work, and vessels so lost backing sessions are never presented as running
- propagate interrupted state through protocol, awareness, manifest, and TUI projections
- use the level-triggered reconciliation path to relaunch an interrupted agent session from its durable brief, then restore the work and convoy to active
- add an `InProcessDaemon` restart test that removes a crew session, observes the durable interruption, verifies relaunch with a new crew identity, and continues the handoff workflow

## Root cause

After a daemon restart, persisted terminal resources could outlive the backing terminal-pool sessions. Reconciliation marked the terminal stopped, but active agent work remained `Running` and no recovery actuation was issued.

## Impact

Active crew work now surfaces a recoverable interruption while its missing session is relaunched. Tool sessions and genuinely failed sessions retain their existing failure behavior.

## Validation

Closes #1153

### DIFF
--- a/crates/flotilla-controllers/src/reconcilers/vessel.rs
+++ b/crates/flotilla-controllers/src/reconcilers/vessel.rs
@@ -1,4 +1,9 @@
-use std::{collections::BTreeMap, marker::PhantomData, path::PathBuf, time::Duration};
+use std::{
+    collections::{BTreeMap, BTreeSet},
+    marker::PhantomData,
+    path::PathBuf,
+    time::Duration,
+};
 
 use chrono::{DateTime, Utc};
 use flotilla_core::agent_adapter::{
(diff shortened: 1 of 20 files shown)


rjwittams:
Fixed the stale work-level interruption message in b031692e. `WorkRunning` now clears `WorkState.message`, and the recovery test applies the patch and verifies both the work and crew messages are cleared.

The timeout and multi-role notes remain observations rather than changes for this PR: unrecoverable terminal launches still escalate through `TerminalSessionPhase::Failed`, while stopped roles are intentionally processed one at a time in the reconciler's existing level-triggered blocking pattern. Each successful relaunch exposes the next stopped role on the following pass, so recovery converges without treating the role set as an exhaustive snapshot.

Issue #1192, opened 2026-07-28
Projects must be durable fleet-wide definitions, not demand-materialized side effects


rjwittams:
## Ruled direction (Robert, 2026-07-28 — detail in comments)

Follow the k8s manifest / GitOps pattern. The resource layer has not drifted from it (full-envelope documents, namespaced stores, idempotent `resource apply` — verified live) and no drift is wanted. What's missing is the loop and the conventions, not machinery.

## Why (the durability split)

"Durable" splits into durable-across-restarts (store persistence — already true) and durable-across-resets (what matters while no-compat store wipes are the sanctioned migration path). Effective durability = "what re-derives this after a wipe?", which sorts store contents into:

1. **Re-derivable from code/disk** — builtins, whole-repo projects of tracked repos, observed checkouts. Resets safe modulo coverage bugs (#1191).
2. **Genuine runtime state** — convoy lifecycle, work status. Resets legitimately cost this until federation/log replication bounds it.
3. **User declarations whose only home is the store** — applied projects, credential grants, user workflow templates, placement policies. Zero effective durability today; they die silently with every reset (feta 2026-07-27 lost the zellij + presentation project definitions this way).

This issue eliminates category 3. Also observed and to be removed: the TUI's view replay issues `project_add` at startup, resurrecting some whole-repo projects by side effect — recovery by coincidence that hides the category (opening a view must never create durable resources).

## Shape

- **Manifest home**: `~/dev/project-map` (lab Forgejo planning repo), git-versioned. Distribution = checkout per host, the same way the fleet shares code; leader-applied/federated distribution deferred (#1188).
- **Mechanism**: a continuous, level-triggered reconciler whose desired state is the checkout — apply on change and on divergence. **Not a boot hook**: daemon restart must never be a semantic event; restart resumes the loop, and restart with no manifest change is a no-op by construction.
- **Idempotent re-apply** is table stakes.
- **Drift: visible, not corrected** (the Argo lesson; same philosophy as the fleet view — surface independent observations, don't collapse them). Never-applied manifests apply. A managed live object diverging from git gets a drift condition + surfaced row, with explicit resolution: `sync` (git wins) or adopt (live wins, written back to the repo), plus a suspend/unmanaged annotation for deliberate hotfix windows.
- **Ownership**: provenance/ownership labels scope drift detection and any future prune to manifest-managed objects; hand-applied resources stay untouched until adopted.
[...]

PR #1193, opened 2026-07-28
fix: reconcile builtin workflow templates

### LINKED ISSUE #1187: Stale stored builtin workflow templates silently downgrade containment
## What happened

First attempted contained convoy (`grouping-rule-capture`, `--workflow single-agent-contained`) placed **host-direct on kiwi** instead of docker-on-feta. Root cause: kiwi's store holds a `single-agent-contained` WorkflowTemplate created 2026-07-21 — before containment landed — with `stance: "trusted"` and the old pre-`CrewSource` crew shape. Builtin registration is create-if-missing, so the stale stored copy shadows the code's `Stance::Contained` builtin forever on any store that predates the change. Admission fetched the stale template, `contained` evaluated false, and the placement sort's local+host-direct preference did the rest.

## Problems

1. **Builtins drift from code with no reconciliation.** In the no-compat phase the code's builtin IS the definition; stores that predate a change keep serving the old one. Builtins should be code-owned: overwritten (or simply not persisted) at daemon boot. A user-modified builtin, if that's meant to be supported, needs an explicit marker rather than silent create-if-missing.
2. **Containment downgrades are silent.** A workflow whose name promises containment ran trusted with zero signal. Whatever the template story, admission stamping a non-docker placement for a workflow the *code* considers contained deserves a loud refusal or at minimum a warning surface on the convoy.

## Acceptance criteria

- [ ] Daemon boot reconciles builtin workflow templates to the code's current spec (no-compat: overwrite)
- [ ] A regression test: store containing a stale trusted-stance copy of a builtin → contained dispatch still gets containment (or refuses loudly)
- [ ] Observed live 2026-07-28; evidence in this issue

rjwittams:
## Summary

- reconcile every code-owned builtin workflow template spec during daemon startup
- overwrite stale stored builtin definitions while preserving their metadata
- cover stale trusted `single-agent-contained` startup state through host-direct dispatch refusal

## Testing

Closes #1187

### DIFF
--- a/crates/flotilla-daemon/src/runtime.rs
+++ b/crates/flotilla-daemon/src/runtime.rs
@@ -451,12 +451,12 @@ async fn register_startup_resources(
     ensure_host_direct_environment_exists(&backend, namespace, profile).await?;
     discover_local_clones(daemon, &backend, namespace, profile).await?;
     ensure_default_policies(&backend, namespace, profile).await?;
-    ensure_default_workflow_templates(&backend, namespace).await?;
+    reconcile_builtin_workflow_templates(&backend, namespace).await?;
     daemon.materialize_tracked_repo_projects().await?;
     Ok(())
 }
 
-fn default_workflow_templates() -> Vec<(&'static str, WorkflowTemplateSpec)> {
+fn builtin_workflow_templates() -> Vec<(&'static str, WorkflowTemplateSpec)> {
     vec![
         (
             "scratch",
@@ -480,15 +480,21 @@ fn default_workflow_templates() -> Vec<(&'static str, WorkflowTemplateSpec)> {
     ]
 }
 
-async fn ensure_default_workflow_templates(backend: &ResourceBackend, namespace: &str) -> Result<(), String> {
+async fn reconcile_builtin_workflow_templates(backend: &ResourceBackend, namespace: &str) -> Result<(), String> {
     let templates = backend.clone().using::<WorkflowTemplate>(namespace);
-    for (name, spec) in default_workflow_templates() {
+    for (name, spec) in builtin_workflow_templates() {
         match templates.get(name).await {
-            Ok(_) => continue,
-            Err(ResourceError::NotFound { .. }) => {}
+            Ok(existing) => {
+                templates
+                    .update(&InputMeta::from(&existing.metadata), &existing.metadata.resource_version, &spec)
+                    .await
+                    .map_err(|err| format!("reconcile builtin workflow template {name}: {err}"))?;
+            }
+            Err(ResourceError::NotFound { .. }) => {
+                templates.create(&empty_meta(name), &spec).await.map_err(|err| format!("seed builtin workflow template {name}: {err}"))?;
+            }
             Err(err) => return Err(format!("check workflow template {name}: {err}")),
         }
-        templates.create(&empty_meta(name), &spec).await.map(|_| ()).map_err(|err| format!("seed workflow template {name}: {err}"))?;
     }
     Ok(())
 }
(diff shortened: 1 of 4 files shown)


rjwittams:
Assessed. Reserving these builtin names and reconciling their specs without a compatibility/customization marker is the explicit behavior required by #1187 for the current no-compat phase. The suggested startup breadcrumb is non-blocking future observability work, so this PR remains scoped to reconciliation and the containment regression.

rjwittams:
Governor review (direction context: #1192, which ruled the manifest-reconciler pattern this startup pass is a special case of):

The core is right — builtins are code-owned and a stale stored copy must lose (#1187's ruling, and your daemon-level dispatch-refusal test covers the incident exactly). Three change requests to align it with the ruled reconciler semantics:

1. **No-op when unchanged**: `reconcile_builtin_workflow_templates` currently calls `update` on every boot even when `existing.spec` already equals the builtin — resource-version churn and watch events for nothing. Guard with `if existing.spec == spec { continue; }`. Restart with no change must be a no-op by construction; restart must never be a semantic event.
2. **Loud on actual change**: when the overwrite really replaces a divergent spec, emit a WARN naming the template ("stored spec diverged from code builtin; overwriting") — the deleted preserve-test shows this PR converts silent-preserve into silent-stomp, and silent is the part that hurt us. Overwrite stays; it just can't be invisible.
3. **Ownership label**: stamp builtin-managed templates with a provenance/ownership label (e.g. `flotilla.work/managed-by: builtin`) on create and update, so this reconciler is visibly the builtin special case of the #1192 manifest loop and future drift/override machinery has the marker it needs. A short doc-comment pointing at #1192 would help the next reader.

None of these grow the diff much; all three keep this mechanism compatible with where the manifest loop is headed (including eventual deliberate builtin overrides from a manifest repo, which need exactly that ownership marker to be distinguishable from drift).

rjwittams:
Fixed all three requests in 727c8f4f:

- Equal builtin specs now return without any update call, including an equal but previously unlabelled object, so its resourceVersion stays stable and no watch event is emitted.
- A successful overwrite of a divergent spec emits WARN with the template name.
- Actual create and divergent-update paths stamp `flotilla.work/managed-by: builtin`, while preserving unrelated labels. The reconciler doc-comment links #1192 as the ruled-manifest-loop context.

Regression coverage asserts the equal-spec no-op, warning text/template name, ownership label, unrelated-label preservation, and stable resourceVersion after reconciliation. The exact format, clippy, and locked workspace test gates pass locally.

rjwittams:
Verified all three requests in `727c8f4f` — structure is right. One residual worth a final tweak, plus a nit to fold in:

The equal-spec early-continue also skips **label convergence**: a stock builtin with a matching spec but no `managed-by` label stays unlabelled forever (until its spec happens to diverge). The ownership label is the very thing future drift/override machinery will scope by, so an unlabelled-but-managed builtin masquerades as hand-applied — undermining the label's purpose on exactly the objects most likely to exist in real stores. Treat the label as part of desired state: if spec matches but the label is missing, do one metadata-converging update (a one-time write per template per store, then no-op forever — still level-triggered, still restart-quiet in steady state). While there: the inner `if spec_diverged` guard after the early continue is always true — drop it.

rjwittams:
Fixed in 1b3b7720.

The managed-by label is now part of desired state: a matching but unlabelled (or differently labelled) builtin receives one metadata-converging update, and the next reconciliation skips the update with a stable resourceVersion. The divergent-spec path remains separate and now emits its WARN directly after a successful update, with no redundant `if spec_diverged` guard.

Focused convergence/reconciliation tests pass, as do the exact nightly format check, workspace clippy with warnings denied, and locked workspace test suite.

rjwittams:
Assessed against the subsequent governor follow-up. Commit 1b3b7720 intentionally supersedes the matching-unlabelled no-op: ownership is part of desired state, so that case now converges once and is write-free thereafter. It also removes the redundant `if spec_diverged` guard. The separate metadata-helper reuse suggestion is cosmetic and non-blocking; `mark_builtin_managed` keeps the builtin ownership operation explicit and scoped to this reconciler.

rjwittams:
Assessed. The duplicated update call is intentional here: it keeps label-only convergence separate from the divergent-spec path, where WARN is emitted only after the update succeeds, without reintroducing a conditional warning guard. This is cosmetic and non-blocking; keeping the current structure.

rjwittams:
Verified `1b3b7720`: semantics are exactly as requested — skip only when spec matches *and* the label is present; one-time metadata-converging update when the label is missing; overwrite + WARN on divergence; dead guard gone. Review cycle complete from my side — ready to merge.

PR #1194, opened 2026-07-28
fix: collect Docker version probe output

### LINKED ISSUE #1189: Interior discovery in docker environments never collects --version probe output
## What happened

Contained smoke on feta (`docker-crew-smoke`, image `flotilla-crew:2026-07-26.1`) failed provisioning in ~1.4s with:

```
image `…flotilla-crew:2026-07-26.1` declares agent adapters `claude-code`, `codex`, but interior discovery did not find them
```

The image is fine — `docker run --rm <image> claude --version` → `2.1.220 (Claude Code)`, `codex --version` → `codex-cli 0.145.0`, both at `/usr/local/bin`.

## Evidence (docker events, exec_create paired with exec_die by execID)

The container was created, the worktree mounted rw at `/workspace`, and discovery ran. Two distinct exec behaviors:

| probe | outcome |
|---|---|
| `sh -lc env` | exit 0, 28ms |
| `which test` | exit 0, 22ms |
| `which cmux` / cmux.app path | exit 1 (correctly absent) |
| `git --git-dir /ref/repo rev-parse --git-dir` | exit 128 |
| `git --version` | **exec created, never died** |
| `gh --version` | **never died** |
| `claude --version`, `/root/.claude/local/claude --version` | **never died** |
| `codex --version`, `agent --version`, `zellij --version`, `cleat --version`, `shpool version`, `gemini --version`, `docker --version` | **never died** |

`git --version` completes in milliseconds anywhere — a created-but-never-reaped exec means the container runner starts output-capturing execs and never collects their results (the `which`/`env` probes evidently go through a different, working path). Discovery therefore sees an empty binary bag, `verify_declared_agent_adapters` fails, `discard_failed_environment` tears the container down (~1s after start), and the pending execs die with it.

## Where to look

The docker `CommandRunner`'s exec implementation in the daemon provisioning path (`crates/flotilla-daemon/src/runtime.rs`, `probe_provisioned_environment` and the container runner it uses) — specifically the attach/wait/output-collection half of exec, versus whatever the `which`-style probes use.

## Context

Blocker #3 for the first contained crew, after #1187 (stale builtin template downgraded stance) and #1188 (cross-store host-direct environment lookup). The rest of the chain demonstrably works: admission requires docker for contained, placement resolves, worktree cut + mounted, container starts, discovery execs — it fails only at reading probe output.

rjwittams:
## Summary

- omit `docker exec -w` when a command runner receives a relative working directory
- preserve explicit Docker working directories for absolute interior paths
- cover the never-reaped version-probe behavior with a timeout-backed regression test

## Root cause

Interior host detectors run version commands with `.` as their working directory. `DockerEnvironmentRunner` forwarded that as `docker exec -w .`, even though Docker exec working directories are container paths and the default container working directory already provides the intended relative-cwd behavior. The daemon created those execs, but their attached output path did not complete, so discovery never observed the installed agent adapters.

Relative runner paths now leave `-w` unset, allowing Docker to use the container's configured working directory. Absolute paths such as `/workspace` continue to be forwarded explicitly.

## Impact

Contained environment provisioning can collect `git`, `claude`, `codex`, and other `--version` probe output and verify declared agent adapters instead of tearing down a valid image.

## Validation

- focused red/green regression: `cargo test -p flotilla-core --locked relative_cwd_version_probe_does_not_hang -- --nocapture`

Closes #1189

### DIFF
--- a/crates/flotilla-core/src/providers/environment/runner.rs
+++ b/crates/flotilla-core/src/providers/environment/runner.rs
@@ -9,8 +9,8 @@ use crate::providers::{
 };
 
 /// A `CommandRunner` decorator that executes all commands inside a Docker container
-/// via `docker exec`. The caller's working directory (a path inside the container)
-/// is forwarded as a `-w` flag; the host-side cwd is always `/` (irrelevant).
+/// via `docker exec`. Absolute working directories are forwarded as `-w`; relative
+/// paths use the container's configured working directory. The host-side cwd is `/`.
 pub struct DockerEnvironmentRunner {
     container_name: String,
     inner: Arc<dyn CommandRunner>,
@@ -126,6 +126,15 @@ mod tests {
                 .expect("version probe should succeed");
 
         assert_eq!(output, "git version 2.51.0\n");
+    }
+
+    #[test]
+    fn interactive_exec_forwards_absolute_workdir() {
+        let runner = DockerEnvironmentRunner::new("my-container".into(), Arc::new(MockRunner::new(Vec::new())));
+
+        let args = runner.docker_exec_args("sh", &["-lc", "pwd"], Path::new("/workspace"), true);
+
+        assert_eq!(args, ["exec", "-i", "-w", "/workspace", "my-container", "sh", "-lc", "pwd"]);
     }
 
     #[tokio::test]
--- a/crates/flotilla-core/src/providers/environment/runner.rs
+++ b/crates/flotilla-core/src/providers/environment/runner.rs
@@ -21,6 +21,19 @@ impl DockerEnvironmentRunner {
         Self { container_name, inner }
     }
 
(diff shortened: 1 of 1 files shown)


rjwittams:
Review follow-up pushed in e18dc70f:

- Updated the `DockerEnvironmentRunner` documentation to distinguish relative paths (container default workdir) from absolute paths (`-w`).
- Added direct coverage for the interactive absolute-workdir argument branch (`exec -i -w /workspace ...`).
- Left the three short `Vec<String>` → `Vec<&str>` conversions at their trait-call sites; extracting that boundary-only conversion would not improve the command-building semantics or reduce meaningful complexity.

The exact local CI gates all pass again: pinned rustfmt, workspace/all-target Clippy with warnings denied, and the full locked workspace test suite.

PR #1204, opened 2026-07-28
fix: publish project checkouts independently

### LINKED ISSUE #1191: Observed-checkout publication still coupled to Plane-A tracking; view-materialized projects render empty
## What happened (kiwi, 2026-07-28, after the #1190 store reset)

The daily driver shows only the flotilla and zellij projects. The other eight projects (cleat, porthole, andamento, presentation, …) exist as Project resources — materialized from `open-views.toml` repo views — but have **zero observed checkouts**, so andamento renders them empty/hidden.

Root cause: observed-checkout publication (`ensure_project_checkout`) is driven by **Plane-A repo inspection**, and kiwi's Plane-A config (`~/.config/flotilla/repos/`) tracks exactly two repos: flotilla and zellij. The other repos were deliberately retired from Plane-A tracking during the transition (`repos.plane-a-retired`), and their local working copies (`~/dev/cleat`, `~/dev/andamento`, …) are now invisible to observation. The old store *looked* correct only because it carried observed checkouts created by since-retired code paths — the #1190 reset deleted that fossil record and exposed the gap.

## What should change

Project/view materialization must own the observation of its repos' local checkouts: when a Project's repo has a local working copy on this host, publish its observed checkout (and worktrees) without requiring Plane-A tracking. Plane-A inspection can keep doing it for tracked repos during the straddle, but the view/project path must be self-sufficient — it is the surviving one.

## Acceptance criteria

- [ ] A repo referenced only by a view/Project (untracked in Plane-A) with a local working copy gets observed checkout resources on the daemon's host
- [ ] Cold-start equivalence test: a daemon bootstrapped from config onto an **empty store** reaches the same projected checkout set as one with history — this is the test shape that catches "state accumulated by retired code" generally, and it would have caught both this and #1190's class
- [ ] The daily driver shows all configured projects with their local checkouts after a store reset

## Context

Straddle-seam bug: projects moved to Plane B (#1044 view-driven materialization), checkout observation stayed Plane-A-coupled. Surfaced immediately after the #1190 recovery reset. See also the reshape umbrella (#604) — this is an argument for the observed-resource substrate taking over local checkout observation sooner rather than later.

rjwittams:
## What changed

- make project reconciliation publish observed checkouts directly instead of relying on Plane-A repository refresh
- extend repository inspection with local checkout enumeration so Git-backed projects publish every worktree, not only the main checkout
- add coverage proving an untracked, view-only project reconstructs the same checkout set on an empty store as it does with historical Repository and Project resources

## Why

Observed checkout publication was still coupled to Plane-A tracking. Projects materialized from local paths could therefore exist without any checkout observations after the observed store was rebuilt, leaving project views empty even though the working copies remained on disk.

Project materialization now owns checkout observation. Plane-A refresh can continue publishing the same facts during the transition, but it is no longer required for the project path to work.

## Impact

Repositories referenced only through Project/view materialization are visible with their main checkout and worktrees after a fresh store boot. Replaying materialization is idempotent whether or not durable Repository and Project history already exists.

## Validation

Closes #1191

### DIFF
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -4475,7 +4475,7 @@ impl InProcessDaemon {
 
         ensure_repository_and_default_project_workflow(&self.resource_backend, &namespace, &key, &repository_spec).await?;
         if let Some(checkout) = checkout {
-            self.ensure_project_checkout(&namespace, &key, &repository_spec, checkout).await?;
+            self.reconcile_project_checkouts(&namespace, &key, &repository_spec, checkout).await?;
         }
 
         let default_name = normalize_project_name(&repository_spec.leaf_slug())?;
@@ -4512,6 +4512,7 @@ impl InProcessDaemon {
         let repository_spec = &inspection.spec;
         let repository_key = repository_spec.key();
         ensure_repository_and_default_project_workflow(&self.resource_backend, &namespace, &repository_key, repository_spec).await?;
+        self.reconcile_project_checkouts(&namespace, &repository_key, repository_spec, inspection.checkout.clone()).await?;
         let repositories = self.resource_backend.clone().using::<Repository>(&namespace);
         let stored = repositories.get(&repository_key.to_string()).await.map_err(|error| error.to_string())?;
         if stored.spec != *repository_spec {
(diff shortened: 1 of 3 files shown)


rjwittams:
Fixed in 8852a263. Project checkout publication now reuses `observed_resources::reconcile_checkouts`, so a branch change updates the observation in place and removed worktrees are retracted instead of wedging subsequent refreshes. I added a regression that exercises both drift cases, plus a mock-runner unit test covering the real `GitRepositoryInspector` worktree enumeration and mapping. The constructor now also documents why its creation-path template is irrelevant during enumeration.

Revalidated with formatting, clippy, the focused project/Git-inspector tests, and the sandbox-safe full workspace suite.

rjwittams:
Thanks for the re-review and confirmation. I filed the two non-blocking efficiency observations together as #1213: https://github.com/flotilla-org/flotilla/issues/1213. They share the same lightweight-observation seam and can be addressed without expanding this correctness fix.

PR #1205, opened 2026-07-28
feat: add manifest apply reconciler

### LINKED ISSUE #1196: Manifest reconciler, first slice: additive apply loop with ownership labels
## Parent

#1192 (ruled design — k8s manifest pattern; this is the first slice).

## What to build

A continuous, level-triggered reconciler in the daemon that applies a directory of resource manifests as desired state. **Not a boot hook**: it runs as a supervised loop (poll mtime/content-hash at a modest interval; fs-watch optional); daemon restart merely resumes it, and a pass with no changes performs zero writes.

- Config: `[manifests] dir = "<path>"` in `daemon.toml` (absent ⇒ loop disabled). The fleet convention will point this at a `~/dev/project-map` subdirectory checkout; the code treats it as an opaque directory.
- Documents: full-envelope resource files (`apiVersion`/`kind`/`metadata`/`spec`, JSON and/or YAML) applied via the existing apply path.
- **Ownership**: every manifest-applied object gets `flotilla.work/managed-by: manifest` plus a source annotation (relative file path) and a last-applied content-hash annotation (reuse the `prepared_snapshot_name` digest machinery).
- **Apply rules (drift-visible, not corrected — per the #1192 ruling):**
  1. Object absent → create with labels/annotations.
  2. Object present, manifest-managed, stored spec == last-applied hash, manifest changed → fast-forward update (git moved, live didn't).
  3. Object present, manifest-managed, stored spec != last-applied (live drift) → do **not** overwrite; WARN with object identity and both digests (the drift-condition *surface* is a later slice; the log line is the slice-1 visibility).
  4. Object present but unmanaged (no ownership label) → never touched; WARN once that a manifest names an unmanaged object (adoption is a later, explicit act).
- **No prune** in this slice (additive only), explicitly documented as such.

## Acceptance criteria

- [ ] Fresh store + populated manifest dir ⇒ all documents applied with ownership labels (cold-start equivalence for declarations)
- [ ] Steady state performs zero writes (assert via resource-version stability across passes)
- [ ] Fast-forward, live-drift-skip, and unmanaged-skip paths each covered by tests using the in-memory backend
- [ ] Malformed document ⇒ named error (file path + reason), loop continues with remaining documents
- [ ] Daemon restart with unchanged dir ⇒ no writes, no events

## Blocked by

None — can start immediately.

rjwittams:
Closes #1196.

## What changed

- add optional `[manifests] dir = "..."` daemon configuration
- run a supervised, level-triggered manifest reconciliation loop every five seconds
- load recursive JSON and YAML full-envelope resource documents through the existing dynamic apply path
- stamp manifest ownership, relative source path, and the shared prepared-snapshot content digest
- create absent objects, fast-forward unchanged live state, and refuse live drift or unmanaged collisions
- isolate malformed files so later documents still apply
- keep this first slice explicitly additive with no pruning

## Validation

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1131,6 +1131,7 @@ dependencies = [
  "rusqlite",
  "serde",
  "serde_json",
+ "serde_yml",
  "tempfile",
  "tokio",
  "tokio-util",
--- a/crates/flotilla-core/src/config.rs
+++ b/crates/flotilla-core/src/config.rs
@@ -366,6 +366,15 @@ pub struct DaemonConfig {
     pub logging: DaemonLoggingConfig,
     #[serde(default)]
     pub environments: BTreeMap<String, StaticEnvironmentConfig>,
+    #[serde(default)]
+    pub manifests: Option<ManifestsConfig>,
+}
+
+/// Host-local directory whose resource documents are continuously applied as
+/// additive desired state.
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
+pub struct ManifestsConfig {
+    pub dir: PathBuf,
 }
 
 /// Host-local structured daemon logging settings.
--- a/crates/flotilla-core/src/config.rs
+++ b/crates/flotilla-core/src/config.rs
@@ -367,13 +367,13 @@ pub struct DaemonConfig {
     #[serde(default)]
     pub environments: BTreeMap<String, StaticEnvironmentConfig>,
     #[serde(default)]
-    pub manifests: Option<ManifestsConfig>,
+    pub manifests: Option<ResourceManifestsConfig>,
 }
 
 /// Host-local directory whose resource documents are continuously applied as
 /// additive desired state.
 #[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
-pub struct ManifestsConfig {
+pub struct ResourceManifestsConfig {
     pub dir: PathBuf,
 }
 
--- a/crates/flotilla-core/src/config/tests.rs
+++ b/crates/flotilla-core/src/config/tests.rs
@@ -626,6 +626,7 @@ fn load_daemon_config_missing_file_returns_default() {
     let config = store.load_daemon_config().unwrap();
     assert!(!config.follower);
     assert_eq!(config.host_name, None);
+    assert_eq!(config.manifests, None);
 }
 
 #[test]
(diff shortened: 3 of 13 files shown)


rjwittams:
Fixed the digest bug in c5034396. Manifest and live specs now both pass through the registered typed `T::Spec` representation before hashing, so omitted Serde defaults remain stable. The new `omitted_serde_defaults_remain_steady_and_can_fast_forward` regression covers a `WorkflowTemplate` with omitted defaults across steady-state and a later edit.

I also preserved unrelated existing labels and annotations during safe fast-forwards, with coverage proving another controller's label survives. For the recursion note, `DirEntry::file_type()` does not follow symlinks, so symlinked directories take neither the directory nor regular-file branch and cannot create a traversal cycle.

rjwittams:
Fixed in the latest commit. Root-directory read failures now stay inside the manifest loop with a deduplicated warning, retry at the configured poll interval, and log recovery once the directory becomes available; they no longer consume the bounded supervisor restart budget. `loop_recovers_when_manifest_directory_appears_later` covers the delayed-directory case.

I also extended the fresh-directory test to load a recursive subdirectory containing a multi-document YAML file, covering both robustness notes.

rjwittams:
Addressed both points in the latest commit. Recursive scanning and all manifest file reads now execute together inside `tokio::task::spawn_blocking`, leaving the async worker available for daemon I/O and controllers. I also renamed the daemon module/type/config type to `resource_manifest`, `ResourceManifestReconciler`, and `ResourceManifestsConfig` to distinguish this resource-envelope apply loop from the existing `flotilla-manifest` presentation metadata crate; the issue-contract `[manifests]` key remains unchanged.

The exact formatting, Clippy, and full workspace test gates pass on this revision.

rjwittams:
Addressed in caaf826e. The reconciler now clears an object's unmanaged-warning dedup entry whenever it observes valid manifest ownership, so losing that ownership again emits a fresh warning instead of remaining silently deduplicated. `unmanaged_warning_is_rearmed_after_ownership_is_restored` covers the unmanaged → managed → unmanaged transition.

Nightly formatting, workspace Clippy with `-D warnings`, and the focused daemon regression pass.

rjwittams:
The unmanaged-warning item in this review was a head timing race: caaf826e landed immediately before the review posted. The current head clears the dedup entry on restored ownership and covers the unmanaged → managed → unmanaged transition. The later review confirms the final head contains the fix. No further action is needed from this intermediate-head review.

rjwittams:
Thanks for confirming the final head independently. Agreed that the remaining kind-casing observation is cosmetic and does not affect reconciliation correctness; no additional change is warranted in this slice. All checks are green on caaf826e.

PR #1206, opened 2026-07-28
fix: harden aggregator projection liveness

### LINKED ISSUE #1190: Aggregator serves empty local projections: subscribers see only peer replicas
## What happened (live, kiwi, 2026-07-28)

After a daemon respawn (06:54Z, `2029c82c`), the daily-driver TUI restart showed a near-empty sidebar and `project/flotilla/flotilla` with **every section empty**. Captured a fresh `flotilla watch --json` subscription:

- `result_set convoys (scope: null)`: **0 rows** — the durable store holds 2 Active convoys at the same moment (`resource list convoys`)
- `result_set independents`: 0 rows
- `result_set checkouts`: exactly 2 rows, both `host: udder` (`/home/robert/dev/flotilla`, `/home/robert/dev/small-repo-b` → BurntSushi/ripgrep) — i.e. **only peer-replicated observations**; kiwi's own observed checkouts were never published (none exist for its tracked repos)
- Plane-A is healthy underneath: `flotilla status` shows flotilla 47 / zellij 520 work items, providers all ok; `repo_snapshot` events flow

So the local half of the aggregator — durable-store convoys projection and local observed-checkout publication — is dead on this daemon instance, while peer replicas merge in fine. The fleet-merged projections render peers-only with no complaint, which is what made it look like "missing data" rather than "dead pump."

Boot-window logs show nothing at WARN+ except transient replicator reconnects (a peer was being bounced at 06:57) and two `zellij action timed out (list-tabs)` at TUI-restart time. No bootstrap/awareness-published markers.

## Problems

1. Whatever pumps local store state into the aggregator projections did not run (or died silently) for this daemon lifetime. A task death or a skipped bootstrap must be loud.
2. **No invariant ties projections to the store**: "local store has N convoys ⇒ a subscriber's convoys result set has ≥ N local rows" would have turned this into an alarm instead of a mystery. A periodic projection-parity self-check (store count vs projection count, divergence ⇒ degraded condition in host status / fleet view) is cheap.
3. **No way to inspect the observed plane or aggregator inputs** (`resource list` reads the durable store only; there is no dump of what the aggregator holds per query, per contributing host). Diagnosis required stitching `watch` JSON by hand. See #1183 — an `--observed`/aggregator-dump verb is the missing tool.

## Why tests are blind to this whole class

Unit and `InProcessDaemon` tests hand each stage its inputs; the daemon *runtime's* task wiring (spawned pumps, bootstrap ordering) has no boot-to-subscriber test asserting store/projection parity. Empty projections are structurally valid output. Belongs to the #1176 liveness-harness family: this is the projection-plane sibling of the reconciler findings.

## Recovery attempted

Rebuild at current main + daemon bounce (also deploys #1182's session recovery for the two crews active on this host); result to be noted in comments.

rjwittams:
## Summary

- publish failed controller supervision and projection parity as degraded Host conditions
- surface degraded conditions in `fleet health` and the TUI fleet view
- quarantine undecodable persisted replica partitions, discard their cursors, and force a full origin relist
- name cached-object decode failures with kind, namespace, name, and origin

## Root cause

The aggregator's initial replica-inclusive list decoded an obsolete cached Convoy and returned an error for the whole source. Its bounded supervisor eventually exhausted the restart budget, but that terminal state existed only in logs. The healthy peer half of the projection remained visible, masking the dead local pump.

Replica data is rebuildable cache state. This change validates it before a replicator trusts its stored cursor; an undecodable partition is atomically removed with its cursor and tombstones so the origin is fully resynced.

## Validation

Closes #1190

### DIFF
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -5285,6 +5285,12 @@ impl InProcessDaemon {
                 is_local,
             );
             let (crew_count, convoys) = counts.remove(&host).unwrap_or_default();
+            let degraded_conditions = status
+                .into_iter()
+                .flat_map(|status| status.conditions.iter())
+                .filter(|condition| condition.value == ConditionValue::False)
+                .map(|condition| format!("{}: {}", condition.condition_type, condition.message))
+                .collect();
 
             rows.push(
                 FleetHostRow::builder()
@@ -5305,6 +5311,7 @@ impl InProcessDaemon {
                     .maybe_disk_free_bytes(status.and_then(|status| status.disk_free_bytes))
                     .staleness(staleness)
                     .observation_agreement(observation_agreement)
+                    .degraded_conditions(degraded_conditions)
                     .build(),
             );
         }
--- a/crates/flotilla-core/src/in_process/tests.rs
+++ b/crates/flotilla-core/src/in_process/tests.rs
@@ -20,18 +20,18 @@ use flotilla_protocol::{
 };
 use flotilla_resources::{
     Checkout as ResourceCheckout, CheckoutPhase as ResourceCheckoutPhase, CheckoutSpec as ResourceCheckoutSpec,
-    CheckoutStatus as ResourceCheckoutStatus, Convoy, ConvoyPhase, ConvoyRepositorySpec, ConvoySpec, ConvoyStatus, CredentialConsumer,
-    CredentialGrant, CredentialGrantSelector, CredentialGrantSpec, CredentialLifecycle, CredentialPlacementRequirements, CredentialSource,
-    CredentialSpec, CredentialSpecSpec, CrewSource, CrewSpec, CrewWorkPhase, CrewWorkState, Environment as ResourceEnvironment,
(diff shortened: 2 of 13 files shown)


rjwittams:
Added the suggested core wiring coverage. The fleet-health test now publishes a degraded `HostStatus` through a replicated Host resource and asserts that `fleet_health_internal` emits the named condition in `FleetHostRow.degraded_conditions`.

The broader readiness behavior is intentional: restart-budget exhaustion means the daemon has permanently stopped a supervised subsystem for this lifetime. Conservatively withdrawing the host from new placement prevents further work from being routed onto a partially dead control plane; the condition still names the exact failed subsystem. The origin-wide cache scan is also intentionally bounded to reconnect/bootstrap recovery for now, with an origin-scoped validation API left as an optimization if fleet size makes it material.

rjwittams:
Verified the follow-up review: it confirms the added end-to-end fleet-health coverage and reports no new blocking findings. No further code change is needed. The single provisioning namespace remains the parity check's deliberate scope.

PR #1207, opened 2026-07-28
fix: reconcile whole-repository projects

### LINKED ISSUE #1199: Whole-repo project specs are never reconciled to the generator (default_workflow_ref stuck on old value)
After #1198 flipped the whole-repo default to `single-agent-trusted`, stored whole-repo Projects still carried `single-agent-contained` — materialization is create-if-missing, so generator changes never reach existing stores (admission reads the stored spec, so the flip was inert for existing projects). Same disease as #1187, project edition; #1193 built the cure pattern (reconcile code-owned objects at startup: no-op on equal, loud on divergence, ownership label — with the user-customization question handled via the label). Whole-repo projects need the same treatment, respecting user-customized fields (issue_source, display_name overrides) — probably: reconcile only generator-owned fields on objects matching the whole-repo shape. Found live on kiwi 2026-07-28 (hand-patched flotilla + zellij as interim). Cold-start equivalence (#1191 AC) would also have caught this.

rjwittams:
## Summary

- reconcile generator-owned whole-repository Project fields during materialization and repeated project adds
- preserve user-owned display names, issue sources, and unrelated metadata while stamping explicit generator ownership
- make metadata-only definition applies converge once, then remain no-ops

## Testing

Closes #1199

### DIFF
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -45,8 +45,8 @@ use flotilla_resources::{
     ResourceBackend, ResourceError, ResourceObject, ResourceProvenance, TerminalBrief, TerminalCrewContext, TerminalCrewMessage,
     TerminalSession as ResourceTerminalSession, TerminalSessionIdentity, TerminalSessionPhase as ResourceTerminalSessionPhase,
     TerminalSessionSource, TerminalSessionStatusPatch, Vessel, WatchEvent, WatchStart, WorkCompletionAuthority,
-    WorkPhase as ResourceWorkPhase, WorkflowTemplate, WorkflowTemplateSpec, CONVOY_LABEL, HEARTBEAT_READY_TTL_SECS, ROLE_LABEL,
-    VESSEL_LABEL, VESSEL_REF_LABEL,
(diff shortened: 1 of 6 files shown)


rjwittams:
Fixed in `262d1763`. `ProjectApply` now loads an existing Project and seeds the apply metadata from `InputMeta::from(&existing.metadata)`; only a missing Project uses fresh metadata. The new `project_apply_preserves_existing_metadata` integration test changes the spec through the command path and verifies that the generator ownership label, an unrelated label, and an annotation all survive.

I also reran the pinned format check, workspace Clippy with warnings denied, and the full locked workspace suite. The first full-suite attempt hit an unrelated transient `ETXTBSY` in the fake-zellij manifest test; that test passed independently and the complete workspace suite then passed on rerun.

rjwittams:
Verified. The re-review confirms the metadata-preservation fix and regression coverage; no further change is required. The optional future multi-host metadata concern does not apply to this reconciler because every host derives the existing metadata and converges the same ownership label.

rjwittams:
Acknowledged. The fresh review confirms the rebased diff has no blocking issues and that the prior metadata-preservation fix remains correct. No code change is required.

PR #1208, opened 2026-07-28
fix: clean up zombie crew sessions

### LINKED ISSUE #1202: Work/session records survive convoy deletion; recovery resurrects zombie crews
Two observations from feta, 2026-07-28: (1) deleted convoy `docker-crew-smoke`'s work row still renders `running` in `flotilla ls` hours after teardown — work-state records are not cleaned up with the convoy; (2) the `admission-free-space-floor` crew session (convoy long gone, killed with `cleat kill --purge` twice) **keeps coming back**: #1182's session recovery faithfully resurrects it from the stale work record at each daemon start. Teardown must delete work/session records with the convoy, and recovery must not restore sessions whose convoy no longer exists (a liveness condition against the owning resource, not just the session).

rjwittams:
## Summary

- delete convoy-owned terminal-session records directly during convoy finalization, including records orphaned from their vessel
- reject and garbage-collect terminal-session recovery when the owning convoy is absent or deleting
- recover disappeared eager crew sessions while work is either Launching or Running

## Root cause

The resurrector is the durable control-plane state in `resources.sqlite`, not cleat: a surviving `TerminalSession` is reconciled directly, while a surviving `Convoy`/`Vessel` work graph recreates a missing terminal record. Convoy teardown previously relied only on the vessel cascade, which let partial/orphaned session records survive. Recovery also treated Running work as live but excluded eager work still Launching, and terminal reconciliation did not validate the owning convoy before ensuring a session.

## Validation

Closes #1202

### DIFF
--- a/crates/flotilla-controllers/src/reconcilers/terminal_session.rs
+++ b/crates/flotilla-controllers/src/reconcilers/terminal_session.rs
@@ -3,10 +3,10 @@ use std::sync::Arc;
 use async_trait::async_trait;
 use chrono::{DateTime, Utc};
 use flotilla_resources::{
-    controller::{ReconcileOutcome, Reconciler},
-    Environment, EnvironmentPhase, ResourceBackend, ResourceError, ResourceObject, TerminalAttention, TerminalAttentionSource,
-    TerminalAttentionState, TerminalSession, TerminalSessionPhase, TerminalSessionStatusPatch, TerminalSessionTag, TypedResolver,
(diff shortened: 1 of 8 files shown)


rjwittams:
Addressed the teardown-order observation in bf452c59. Convoy finalization now marks/deletes vessels before running the direct terminal-session sweep, so a vessel cannot observe a missing session while it is still eligible to reconcile and recreate it. The terminal sweep remains after the vessel cascade as the orphan/partial-teardown backstop.

rjwittams:
Re-check acknowledged. No further changes are needed; the teardown-order concern is resolved in bf452c59 and the remaining findings confirm the intended behavior.

PR #1210, opened 2026-07-28
fix: isolate root-owned checkout debris

### LINKED ISSUE #1201: Container-written root-owned files in mounted worktrees permanently kill the checkout controller
The abandoned contained smoke left root-owned files (container wrote `.flotilla/briefs/coder.md` as root) in the host-mounted worktree. Host-side teardown `rm` as the daemon user fails forever; the checkout controller crash-loops on it and **exhausts its restart budget — silently disabling all checkout provisioning on the host**. Four freshly-dispatched convoys sat Active/vesselless for over an hour on feta (2026-07-28 ~13:00Z) until the debris was removed via a scratch container and the daemon bounced. Wants: (1) run crew containers as the host uid/gid (the research doc notes flotilla has no --user plumbing at all — docs/superpowers/research/2026-07-27-contained-vessel-plumbing.md) or chown on teardown via a root helper container; (2) checkout-teardown failure on one checkout must not take the whole controller down (isolate per-checkout); (3) budget exhaustion as a surfaced degraded condition is already an AC on #1190 — this incident is its second live validation.

rjwittams:
## Summary

- run Docker-backed crew environments as the daemon host's numeric uid/gid
- isolate named-object reconciliation and finalizer failures so one checkout cannot terminate a controller loop
- mark the affected checkout failed with the reconciliation error while allowing other checkouts to continue

## Root cause

Docker environments inherited the image's configured user, commonly root, while writing through a read-write host worktree mount. Root-owned debris then made host-side checkout teardown fail. The generic controller loop propagated that object-specific finalizer error out of `run()`, causing supervision to replay the same failing checkout until the shared restart budget was exhausted.

## Impact

Container writes now retain host ownership on Unix. A permanently failing checkout teardown is logged and reflected on that checkout, while the controller continues provisioning and reconciling unrelated checkouts. Watch and resync failures remain supervisory.

## Validation

Closes #1201

### DIFF
--- a/crates/flotilla-controllers/src/reconcilers/checkout.rs
+++ b/crates/flotilla-controllers/src/reconcilers/checkout.rs
@@ -216,4 +216,12 @@ where
     fn finalizer_name(&self) -> Option<&'static str> {
         Some("flotilla.work/checkout-cleanup")
     }
+
+    fn error_patch(&self, obj: &ResourceObject<Self::Resource>, error: &ResourceError) -> Option<CheckoutStatusPatch> {
+        let message = format!("checkout reconciliation failed: {error}");
+        if obj.status.as_ref().is_some_and(|status| status.phase == CheckoutPhase::Failed && status.message.as_deref() == Some(&message)) {
+            return None;
+        }
+        Some(CheckoutStatusPatch::MarkFailed { message })
+    }
 }
--- a/crates/flotilla-controllers/src/reconcilers/checkout.rs
+++ b/crates/flotilla-controllers/src/reconcilers/checkout.rs
@@ -217,8 +217,8 @@ where
         Some("flotilla.work/checkout-cleanup")
     }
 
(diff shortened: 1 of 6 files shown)


rjwittams:
Addressed the two substantive points in `9ed7d9db`.

1. The status hook is now `finalizer_error_patch` and is invoked only when `run_finalizer` fails. Metadata conflicts, dependency failures, and actuation failures remain isolated to the named object but do not permanently move a checkout to `Failed`; they retry on later events/resyncs.
2. Added an end-to-end `ControllerLoop<CheckoutReconciler>` regression: one checkout teardown returns the incident-shaped permission error and is observed as `Failed` with its finalizer retained, while a second deleting checkout completes and disappears.

The unconditional numeric `--user uid:gid` behavior is intentional for this issue's contract. There is no current environment user override, and preventing host-mounted worktree writes from being owned by the image user is the required invariant. The full pinned format, clippy, and locked workspace test gates pass after the revision.

rjwittams:
Thanks for the follow-up verification. Confirmed the Test check has now completed successfully as well; all four required checks are green. No further code changes are needed from this review cycle.

Issue #1140, opened 2026-07-27
Operator setup: crew machine account, host tokens, Forgejo crew user


rjwittams:
## Parent

Map #1046 · #1050/ADR 0022. HITL task — human-performed setup that #1139's forge-identity parts block on.

## Checklist (Robert)

- [ ] Create a GitHub machine account for crew work (suggested: `flotilla-crew`), invite it to `flotilla-org` with write on the working repos
- [ ] Generate its PAT (classic or fine-grained with repo scope) and place it host-locally on kiwi and feta (suggested: `~/.config/flotilla/credentials/github-crew-token`, 0600)
- [ ] Create a Forgejo crew-class agent user on the lab hub with tracker scopes (`write:issue,read:repository`), token placed per host alongside the existing per-agent tokens
- [ ] (Later, when App minting lands) register a GitHub App for per-crew installation tokens — not needed for first enactment

## Notes

Material is host-local by ADR 0022 — do not put these tokens in any repo, resource, or replicated config. Names/paths above are suggestions; whatever is chosen gets recorded in the host credential store convention in #1139.

rjwittams:
Addition for the fork tracer (#1081): the machine account also needs **write access to rjwittams/zellij** (collaborator invite, or whatever grant shape you prefer) so a contained crew can push its branch and open the fork PR there. Recommend the PAT be fine-grained and scoped to exactly the crew-writable repos (rjwittams/zellij now; extend per fork project). Without this the tracer's crew fails at push — the credential path (ADR 0022 grants + gh adapter preflight from #1139) is part of what the tracer proves.

rjwittams:
Checklist revised (ruled 2026-07-27): **GitHub App instead of machine-account PAT.**

Why: fine-grained PATs cannot be created via API (web-UI only, long-lived, plus a second account to manage). App installation tokens are mintable programmatically (1-hour expiry) and each mint can be narrowed to specific repositories and permissions — per-repo scoping happens at mint time, so it's **one app, not one per repo**. The app is its own identity (`flotilla-crew[bot]`), which satisfies ADR 0022's "crews never carry human forge identity" without a pseudo-user. The mint script becomes an ADR 0022 `issue-command` credential source, plugging directly into the #1139 adapter/preflight machinery.

## Operator steps (manual, ~10 min)

[...]

PR #1211, opened 2026-07-28
docs: amend ADR 0022 — subscriptions primary, lease-located material, forge/model split

rjwittams:
Carries the 2026-07-28 credential rulings from #1140 into the ADR: forge-vs-model identity split (subscriptions deliberately primary for the target segment), material as lease-located rather than host-pinned (codex conch pool; setup-token multiplicity), HITL up-front minting as first-class flows with flotilla as login inventory, codex ambient-key correction, zero-prompts acceptance test, proxy-compatible delivery interface.

### DIFF
--- a/docs/adr/0022-credential-declarations-grants-and-local-material.md
+++ b/docs/adr/0022-credential-declarations-grants-and-local-material.md
@@ -1,6 +1,6 @@
 # Credentials: replicated declarations, host-local material, stance-first grants
 
-**Status:** Accepted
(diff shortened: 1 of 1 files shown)


rjwittams:
Correct catch — the bullet flattened the research doc's own walk-back, and the nuance is operationally load-bearing (crews run the interactive TUI; the containment smoke's login screen was this exact case). Fixed in the follow-up commit with your suggested split preserved, plus the corollary the doc records: two interactive crews on two different accounts is the one narrow case that revives per-crew `CODEX_HOME`s.

PR #1212, opened 2026-07-28
fix: detect and surface sleep inhibitor failures

### LINKED ISSUE #1181: Sleep inhibitor silently fails (systemd-inhibit exits 1) — host slept under an active crew
## What happened

feta (Linux) suspended at ~2026-07-28T00:0xZ while hosting an active convoy (`work-session-liveness`, crew provisioning) and a running daemon. The daemon believed it held a sleep inhibitor, but its log shows the acquire/die loop repeating every 30s for hours:

```
INFO flotilla_daemon::sleep_inhibitor: acquired system sleep inhibitor program="systemd-inhibit" reason="Flotilla vessel crew is active"
WARN flotilla_daemon::sleep_inhibitor: system sleep inhibitor process exited; reacquiring if still required status=exit status: 1
```

`systemd-inhibit` exited with status 1 **immediately after every acquisition**, so no inhibition was ever actually held. The reacquire loop masks the failure: each iteration logs a reassuring "acquired" line, the WARN reads as transient, and nothing escalates when the child exits instantly N times in a row. The host slept; the crew's terminal-session launch actuation was lost mid-provisioning (vessel Ready, convoy Active, no cleat session — the #1153 false-liveness shape); recovery required wake-on-lan and a daemon bounce.

## Problems

1. **Exit status of the inhibitor child is treated as noise.** An immediate exit 1 means the inhibition call itself is broken (permissions, missing polkit authorization, wrong invocation for this distro — cause unknown, worth diagnosing on feta). That is a persistent failure, not a transient one.
2. **No escalation or surfacing.** After k consecutive immediate failures the daemon should mark the capability degraded — surface it in host status (a `sleep_inhibition: failed` condition the fleet view could show) rather than looping forever at WARN.
3. **The failure defeated its reason string**: "vessel crew is active" was precisely the situation the inhibitor exists for.

## Acceptance criteria

- [ ] Diagnose and fix why `systemd-inhibit` exits 1 on feta (likely needs a session/polkit-compatible invocation)
- [ ] Immediate-exit loops are detected and surfaced as a degraded host condition instead of infinite WARN+reacquire
- [ ] Host status / fleet view expose sleep-inhibition health

rjwittams:
## What changed

- confirm inhibitor helper processes survive startup before reporting acquisition
- on Linux, try the broad logind sleep inhibitor first, then use a verified PowerDevil `kde-inhibit --power` fallback, with logind idle inhibition as the non-desktop fallback
- escalate three consecutive acquisition failures into a durable `sleep_inhibition: failed` Host status condition
- project sleep-inhibition health into the fleet-health protocol, CLI table, and TUI fleet panel
- preserve the condition across periodic Host heartbeat updates

## Root cause

Reproducing on feta showed `systemd-inhibit --what=sleep --mode=block` exits immediately with status 1 because flotillad belongs to a remote logind session. Feta's polkit policy challenges remote-session callers for block-sleep, while `systemd-inhibit` does not enable interactive authentication. PowerDevil runs in the active local Wayland session and owns feta's five-hour auto-suspend policy.

An idle-only logind lock was tested and rejected as insufficient: PowerDevil maps it to screen-management inhibition, while its auto-suspend path requires an interrupt-session inhibition. The verified KDE fallback becomes an active PowerDevil sleep inhibition after its enforcement delay and releases with the daemon-bound payload.

## Impact

Active crews on feta now obtain a session-compatible sleep inhibitor instead of silently running without protection. If every strategy fails, the daemon stops emitting misleading acquisition messages, records the failure after three consecutive attempts, and exposes it in fleet health.

## Validation

- live reproduction on feta of the original exit-1/polkit error
- live verification that the generated KDE fallback appears as an active PowerDevil sleep inhibition and releases cleanly

Closes #1181

### DIFF
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -5291,6 +5291,7 @@ impl InProcessDaemon {
                     .crew_count(crew_count)
                     .convoy_count(convoys.len())
                     .maybe_disk_free_bytes(status.and_then(|status| status.disk_free_bytes))
+                    .sleep_inhibition(status.map(|status| status.sleep_inhibition.clone()).unwrap_or_default())
                     .staleness(staleness)
                     .observation_agreement(observation_agreement)
                     .degraded_conditions(degraded_conditions)
--- a/crates/flotilla-daemon/src/runtime.rs
+++ b/crates/flotilla-daemon/src/runtime.rs
@@ -291,6 +291,7 @@ impl DaemonRuntime {
             spawn_sleep_inhibitor_task(
                 daemon.resource_backend(),
                 options.namespace.clone(),
+                profile.host_id.clone(),
                 options.controller_supervision.clone(),
                 runtime_health.clone(),
             ),
@@ -749,13 +750,16 @@ where
 fn spawn_sleep_inhibitor_task(
     backend: ResourceBackend,
     namespace: String,
+    host_id: String,
     supervision: ControllerSupervision,
     runtime_health: RuntimeHealth,
 ) -> JoinHandle<()> {
     tokio::spawn(async move {
         supervise_controller("sleep_inhibitor", supervision, runtime_health, move || {
             let convoys = backend.clone().using::<Convoy>(&namespace);
-            async move { sleep_inhibitor::run(convoys).await }
+            let hosts = backend.clone().using::<Host>(&namespace);
+            let host_id = host_id.clone();
+            async move { sleep_inhibitor::run(convoys, hosts, host_id).await }
         })
         .await;
     })
@@ -975,6 +979,7 @@ async fn apply_host_heartbeat_with_credentials(
         daemon_started_at: Some(health.started_at),
         disk_free_bytes,
         conditions,
+        sleep_inhibition: host.status.as_ref().map(|status| status.sleep_inhibition.clone()).unwrap_or_default(),
     };
     hosts.update_status(&profile.host_id, &host.metadata.resource_version, &status).await.map_err(|err| err.to_string())?;
     daemon.refresh_connected_peer_host_heartbeats().await;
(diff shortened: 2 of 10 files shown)


rjwittams:
Addressed both findings in 8bde46f4:

- KDE D-Bus verification now has a 3-second timeout and kills the child on timeout/drop, with a regression test covering a hung verifier.
- The controller now uses health-sensitive retry intervals: 5 seconds while acquiring, 30 seconds while healthy, and 5 minutes after escalation to Failed. A changed failure message still emits an error-level degraded event.

The exact format, workspace clippy, and full workspace test gates pass locally.

rjwittams:
Can we make sure this isn't *requiring* KDE

rjwittams:
KDE is not required. On Linux the strategies are tried in order:

1. `systemd-inhibit --what=sleep`
2. verified `kde-inhibit --power`
3. `systemd-inhibit --what=idle`

If `kde-inhibit` is absent, spawning it fails and acquisition immediately continues to the final systemd strategy. If `qdbus6` is absent or cannot verify KDE's inhibitor, only the KDE strategy is rejected and the final systemd strategy is still attempted. On macOS the implementation uses `caffeinate` and never touches KDE tooling.

rjwittams:
The post-fix review confirmed the blocking findings are addressed. I filed the valid non-blocking heartbeat status-writer race as follow-up #1219: https://github.com/flotilla-org/flotilla/issues/1219

That refactor can move heartbeat-owned fields onto the conflict-retrying `HostStatusPatch::Heartbeat` path without expanding this sleep-inhibitor fix.

rjwittams:
Addressed both current-diff review observations in 022956b7:

- A successfully acquired long-lived inhibitor now has its piped stderr drained asynchronously, preventing future helper output from filling the pipe and blocking the process.
- CLI fleet-health Diagnosis now includes `⚠ SLEEP INHIBITION FAILED`, and composes it with degraded-condition and observation-disagreement diagnoses. The human-output test covers the combined result.

The exact format, workspace Clippy, and full workspace test gates pass locally after these changes.

rjwittams:
Removed the diagnosis formatter's `unreachable!` in e84f9594. The empty-diagnosis fallback is now total, while an actual disagreement still enters the composed diagnosis before that fallback.

The exact format, workspace Clippy, and full workspace test gates pass locally.

PR #1214, opened 2026-07-28
test: add reconciler liveness contract harness

### LINKED ISSUE #1176: Reconciler liveness harness — slice 1: injectable clock, contract battery, two enrollees
## Ruled contract (grilled 2026-07-28; original motivation and evidence in the history below)

Slice one of the liveness-testing direction, batched as one convoy: **injectable clock + reusable contract battery + two enrolled reconcilers.** Supervisor/controller-loop code is explicitly OUT of scope (live crews in #1190/#1201 are in it); event/level equivalence and adversarial world generation are named later slices; deterministic simulation stays deferred until Plane-A retirement.

### 1. Injectable clock

A `Clock` trait (`now() -> DateTime<Utc>`), injected through construction like `CommandRunner`/`EnvVars` — never a global or thread-local. `SystemClock` default; `VirtualClock` (manual advance) in test-support. Thread it to the TTL/staleness consumers the battery needs: the checkout refresh/cleanup paths and the convoy landing decision edge (`LANDING_EVIDENCE_TTL`, `CHECKOUT_INTEGRATION_REFRESH_AFTER` sites). Inline `Utc::now()` may remain elsewhere for now; enrolled paths must use the injected clock.

### 2. Contract battery (test-support module in `flotilla-resources`)

Enrollment mirrors the storage contract tests: a per-reconciler fixture supplying (world builder, reconciler step, fixpoint predicate); properties generic over that triple. The battery drives the **reconcile step directly** (no tokio task machinery) against an in-memory backend wrapped in a **write-counting decorator** (a first-class harness piece — also the tool for "restart is a no-op" assertions elsewhere). Properties:

- **Bounded convergence**: fixpoint within a constant pass bound (default 10).
- **Quiescence at fixpoint**: further passes with unchanged inputs ⇒ zero writes and zero actuations.
- **Staleness edges**: advance `VirtualClock` across each enrolled TTL; assert the decision path re-probes rather than consuming stale state (the #1163 class).
- **Actuation-drop recovery**: where actuations are values, discard one pass's actuations; assert a later pass re-emits them (the #1166 class).
- **Degradation, not wedging**: contradictory worlds (resource Ready but path missing — the #1165/#1201 trigger shape) ⇒ held condition with bounded retries, never unbounded loops or silent success.

### 3. First enrollees

1. **Checkout reconciler** (`flotilla-controllers`): full battery including actuation-drop (its actuations are values).
2. **Convoy landing/teardown path** (`convoy/reconcile.rs` + the `in_process` decision edge): convergence, quiescence, staleness edges; actuation-drop only where effects are value-shaped (state the limitation honestly in code comments rather than forcing it).

### Acceptance criteria

- [ ] `Clock` trait injected through construction; enrolled paths take it; `VirtualClock` advances deterministically in tests
- [ ] Battery module in `flotilla-resources` test-support with the five properties, generic over enrollment fixtures
- [ ] Write-counting backend decorator usable standalone
- [ ] Both enrollees pass the battery; at least one property per enrollee demonstrably fails against a deliberately re-broken variant (prove the battery bites — e.g. revert the #1163 latch guard in a test-local copy)
- [ ] No changes to supervisor/controller-loop code
- [ ] CI runs the battery as ordinary `cargo test`

## Original Brainstorm (2026-07-27, retained for context)

[The class-fix motivation: #1163 lived in time, #1165 in repetition, #1166 in loss; safety is tested, liveness is not. Deferred by ruling: DST until the system has a more complete feel and Plane-A is retired; adversarial world generation longer-term and off the PR gate. See comment history.]

rjwittams:
## Summary

- introduce a shared injectable clock with deterministic virtual time and thread it through checkout freshness and convoy landing decisions
- add a reusable direct-step liveness battery plus a standalone write-counting backend decorator
- enroll checkout and convoy landing with convergence, quiescence, staleness, degradation, and supported actuation-drop contracts
- prove both staleness contracts bite test-local re-broken variants and cover the production in-process landing edge

## Validation

Closes #1176

### DIFF
--- a/crates/flotilla-controllers/Cargo.toml
+++ b/crates/flotilla-controllers/Cargo.toml
@@ -19,6 +19,7 @@ sha2 = "0.10"
 serde_json = { workspace = true }
 
 [dev-dependencies]
+flotilla-resources = { path = "../flotilla-resources", features = ["test-support"] }
 tokio = { workspace = true, features = ["macros", "rt-multi-thread"] }
 rstest = { workspace = true }
 tempfile = "3"
--- a/crates/flotilla-controllers/src/reconcilers/checkout.rs
+++ b/crates/flotilla-controllers/src/reconcilers/checkout.rs
@@ -4,8 +4,8 @@ use async_trait::async_trait;
 use chrono::{DateTime, Utc};
 use flotilla_resources::{
     controller::{ReconcileOutcome, Reconciler},
-    Checkout, CheckoutBranchProvenance, CheckoutIntegrationStatus, CheckoutPhase, CheckoutSpec, CheckoutStatus, CheckoutStatusPatch, Clone,
-    ClonePhase, IntegrationCondition, ResourceBackend, ResourceError, ResourceObject, TypedResolver,
+    Checkout, CheckoutBranchProvenance, CheckoutIntegrationStatus, CheckoutPhase, CheckoutSpec, CheckoutStatus, CheckoutStatusPatch, Clock,
+    Clone, ClonePhase, IntegrationCondition, ResourceBackend, ResourceError, ResourceObject, SystemClock, TypedResolver,
 };
 use tracing::warn;
 
@@ -60,11 +60,16 @@ pub trait CheckoutRuntime: Send + Sync {
 pub struct CheckoutReconciler<R> {
     runtime: Arc<R>,
     clones: TypedResolver<Clone>,
+    clock: Arc<dyn Clock>,
 }
 
 impl<R> CheckoutReconciler<R> {
     pub fn new(runtime: Arc<R>, backend: ResourceBackend, namespace: &str) -> Self {
-        Self { runtime, clones: backend.using::<Clone>(namespace) }
+        Self::with_clock(runtime, backend, namespace, Arc::new(SystemClock))
+    }
+
+    pub fn with_clock(runtime: Arc<R>, backend: ResourceBackend, namespace: &str, clock: Arc<dyn Clock>) -> Self {
+        Self { runtime, clones: backend.using::<Clone>(namespace), clock }
     }
 }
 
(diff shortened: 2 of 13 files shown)


rjwittams:
Addressed both observations in 1ab7db2c. The resources harness and `VirtualClock` are now behind a propagated `test-support` feature, with the controllers test dependency opting in, and a default-feature production check verifies the harness is excluded. The convoy fixpoint now requires at least one completed direct reconcile pass, so bounded convergence can no longer pass vacuously at pass zero. The exact formatting, clippy, and full workspace test gates all pass after these changes.

rjwittams:
Addressed the remaining cleanup in 9dd4bc98 by removing the unused RegardLifecycle::with_system_clock constructor. The shared Clock and SystemClock re-exports remain intact. The exact formatting, clippy, and full workspace test gates pass after the change.

rjwittams:
Addressed the test-precision observation in 432f4a09. ConvoyFixpoint::at_fixpoint now derives only from completed passes and actual convoy phase, so a contradictory scenario that wrongly reaches Landing/Landed triggers the battery’s specific successful-fixpoint assertion; held() still expresses the expected degraded outcome. Exact formatting, clippy, and full workspace tests pass.

rjwittams:
Thanks for the careful re-trace. On WriteCountingResolver, counting before await is intentional: the decorator measures mutation attempts, not only successful persistence. For quiescence, a controller attempting create/update/delete at a fixpoint is the behavior under test even if the underlying backend rejects that attempt. Failure/retry scenarios can separately inspect backend results alongside this attempt count, so I’m leaving the semantics unchanged.

rjwittams:
Acknowledged; thank you for the independent re-trace and confirmation that no issues remain.

rjwittams:
The unused RegardLifecycle::with_system_clock noted here was removed in 9dd4bc98; the later review confirmed the cleanup and the required Clock/SystemClock re-export remains.

Issue #1188, opened 2026-07-28
Contained placement on a remote host resolves its host-direct environment in the wrong store


rjwittams:
## Ruled contract (Robert, 2026-07-28 — grilled; evidence and history in comments)

**Direction: resource-store federation slice** (ownership routing was considered and retracted — it bakes in single-placement-host convoys, which multi-repo/fork-DAG futures falsify).

### The slice

1. Convoy + pinned snapshots live on the **admitting leader's store** (leader-only content — issue snapshots, workflow snapshot, grant resolution — is already pinned at admission; followers never need providers).
2. The convoy replicates like everything else; **placement-host reconcilers watch replicated-in vessels/checkouts** with a placed-on-me filter and actuate against their local environments — dissolving this bug's cross-store env lookup rather than patching it.
3. **Single-writer per store**: actuators never write into the owner's objects; each host appends its facts about the convoy to its own store/log, and views merge owner-spec with actuator-facts (the #1179 fleet-health pattern).

### Field-ownership split (ruled)

- **Owner-authored** (admitting store only): spec, metadata incl. pin annotations, deletion intent, and the **phase roll-up** — lifecycle transitions are decisions derived from merged facts, with exactly one decider: the owner's reconciler.
- **Actuator-observed** (each actuating host, own store): vessel/checkout/work status, session liveness, integration conditions + landed evidence, crew completion claims — per-host fact streams, merged aggregator-side.

### Offline semantics (ruled)

**Act on last-known state, destructive carve-out**: actuators keep driving provisioning/crew work from the last replica while the owner is unreachable (idempotent actuation + level-triggering make reconnect catch-up safe); teardown/deletion freeze — "owner unreachable" is the strongest form of evidence-not-fresh. This falls out of the field split for free: facts flow and work progresses owner-down, but phase (and thus owner-decided teardown) freezes. Cost accepted: phase display lags while the owner is offline — the fleet view's staleness column presents that honestly.

Future thread, out of scope: explicit ownership *takeover* (laptop owners handing off convoys).

## Original evidence (2026-07-28)

Contained dispatch from kiwi with a feta-targeted docker policy failed: `host-direct environment host-direct-d783f8a7… not found` — the `worktree_on_host_and_mount` checkout strategy resolved the target host's environment in the dispatching store. Workaround until this lands: dispatch on the placement host (follower hosts need issueless/instruction convoys), or host-direct placements, which admission handles cross-store today.

## Relation

[...]

PR #1217, opened 2026-07-28
Make placement decisions legible and add raw resource deletion

### LINKED ISSUE #1209: Location legibility slice 1: host names, placement decision records, resource delete verb
## Parent

#1183 (umbrella). This is the model-agnostic slice, ruled dispatchable ahead of the #1188 federation work; verb *routing* stays parked with #1188.

## What to build

Three pieces, grilled 2026-07-28:

### 1. Host names on operator surfaces

Every surface that shows a host shows its resolved display name (kiwi/feta/udder), not the machine-id UUID: `flotilla ls`, `resource list/get` human output, convoy/vessel status messages, admission refusal strings. The UUID remains in JSON output; humans get names.

### 2. Placement decision records — one shape, two levels

A shared `PlacementDecision` status type: chosen policy name, target host (ref + display name), and the considered-but-refused candidates with the refusal reasons admission/provisioning already computes. Written **once at decision time**:
- **Convoy status**: admission's selection.
- **Vessel status**: that vessel's own placement resolution (grows more important as placements diversify beyond host-direct-where-the-convoy-lives — docker, per-vessel policies, federation).
Rendered in `ls`/`get`/TUI detail. Replicates with its resource.

### 3. `flotilla resource delete <kind> <name> [--namespace]`

The operator escape hatch (the 2026-07-28 ghost-record outage required raw sqlite surgery for want of it — #1202). Semantics: **raw in authorization, not in propagation**:
- Bypasses teardown gates and lifecycle authority; deletes exactly one object; exact kind+name, no fuzzy matching.
- Goes through the store's normal deletion path: watch event emitted, tombstone written, replicas converge, projections drop the row. (Counter-example to prevent: the incident's sqlite DELETE wrote no tombstone, so peer replicas may still carry the ghost.)
- Loud: prints the deleted object's full identity plus a note that controllers may recreate code-owned objects (correct level-triggered behavior).
- No `--force` tiers — the safe paths are the orchestrated verbs; this stays a sharp, documented tool.

## Acceptance criteria

- [ ] No UUID-as-host on any human-facing surface listed above
- [ ] `PlacementDecision` on both convoy and vessel status, written at decision time, visible in ls/get; refusal reasons preserved verbatim
- [ ] `resource delete` emits the same watch/tombstone/replica propagation as any store delete; a peer's replica provably drops the object (test via in-memory backends per testing philosophy)
- [ ] Deleting a code-owned builtin is observed to be reconciled back (documented, tested)

rjwittams:
## Summary

- preserve stable host IDs in machine-readable JSON while rendering human-facing host names in resource, fleet, and TUI views
- record a shared placement decision on convoys and vessels, including the chosen policy and host plus exact refused-candidate reasons
- add `flotilla resource delete <kind> <name> [--namespace]` through the normal resource deletion, tombstone, watch, and replication path, with a warning that controllers can recreate code-owned resources
- cover placement propagation, legible presentation, replica deletion, and builtin policy recreation

## Impact

Operators can now understand where and why work was placed without decoding host UUIDs, inspect rejected placement candidates, and use an explicit sharp recovery command to remove an exact resource object.

## Validation

Closes #1209

### DIFF
--- a/crates/flotilla-controllers/src/reconcilers/vessel.rs
+++ b/crates/flotilla-controllers/src/reconcilers/vessel.rs
@@ -9,6 +9,7 @@ use chrono::{DateTime, Utc};
 use flotilla_core::agent_adapter::{
     append_convoy_work_context, build_crew_brief_with_options, CrewAssignment, CrewBriefMember, CrewBriefTemplateResolver,
 };
+use flotilla_protocol::PlacementDecision;
 use flotilla_resources::{
     canonicalize_repo_url, clone_key,
     controller::{
@@ -104,9 +105,11 @@ enum PlannedPatch {
     Provisioning {
         observed_policy_ref: String,
         observed_policy_version: String,
+        placement_decision: Option<PlacementDecision>,
         waiting_for: String,
     },
     Ready {
+        placement_decision: Option<PlacementDecision>,
         environment_ref: String,
         image: Option<ImageStamp>,
         checkout_refs: BTreeMap<RepositoryKey, String>,
@@ -142,25 +145,15 @@ impl VesselDeps {
 
     fn provisioning(
         placement_policy: &ResourceObject<PlacementPolicy>,
+        placement_decision: Option<PlacementDecision>,
         waiting_for: impl Into<String>,
         actuations: Vec<Actuation>,
     ) -> Self {
-        Self { patch: provisioning_patch(placement_policy, waiting_for.into()), actuations }
-    }
-
-    fn ready(
-        environment_ref: String,
-        image: Option<ImageStamp>,
-        checkout_refs: BTreeMap<RepositoryKey, String>,
-        terminal_session_refs: Vec<String>,
-        requested_stance: Stance,
-        effective_stance: Stance,
-        actuations: Vec<Actuation>,
-    ) -> Self {
-        Self {
-            patch: PlannedPatch::Ready { environment_ref, image, checkout_refs, terminal_session_refs, requested_stance, effective_stance },
-            actuations,
+        let mut waiting_for = waiting_for.into();
+        if let Some(decision) = &placement_decision {
+            waiting_for = waiting_for.replace(&decision.target_host.reference, &decision.target_host.display_name);
         }
+        Self { patch: provisioning_patch(placement_policy, placement_decision, waiting_for), actuations }
     }
 
     fn failed(message: impl Into<String>) -> Self {
@@ -202,6 +195,7 @@ impl Reconciler for VesselReconciler {
             }
             Err(err) => return Err(err),
         };
+        let placement_decision = convoy.status.as_ref().and_then(|status| status.placement_decision.clone());
         let strategy = match placement_strategy(&placement_policy.spec) {
             Ok(strategy) => strategy,
             Err(message) => return Ok(VesselDeps::failed(message)),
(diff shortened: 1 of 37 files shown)


rjwittams:
Addressed all four points in e888c04b:

1. Malformed policies now retain their exact refusal reason and use an explicit `no target host` display value instead of being silently dropped; a regression test covers this.
2. The charter now states that raw delete targets the connected daemon's local store. This matches #1209's scope, which explicitly parks peer verb routing with #1188.
3. Waiting-message legibility now replaces the structured `host-direct-<host-ref>` environment name, not arbitrary host-ref substrings. A unit test verifies both the intended rewrite and that unrelated names remain untouched.
4. Added coverage proving default placement falls through a credential-refused candidate to a viable candidate while preserving the exact credential refusal reason.

The exact local fmt, clippy, and full workspace test commands all pass after these changes.

rjwittams:
Verified both observations; neither needs another code change in this slice:

1. #1209's acceptance criterion explicitly scopes resource human rendering to `resource list/get`; `resource watch` was not included. Extending watch requires changing its streaming formatter and host-name refresh semantics, so it stays outside this issue rather than being folded into the completed list/get work.
2. The exact-match traversal is intentionally limited to human rendering of otherwise generic dynamic resource JSON. The canonical JSON path is returned before any replacement, and embedded strings/resource names are not changed, as covered by `human_resource_rendering_replaces_exact_host_ids_but_not_embedded_object_names`. A fully typed renderer for every registered dynamic resource would be a broader presentation-layer redesign; the current exact-match mapping satisfies this slice without weakening machine identity.

Thanks for the second pass and confirmation that the earlier findings are resolved.

rjwittams:
Fixed in de708d0b. Crew-session placement lookup is now keyed by `(namespace, convoy name)` and uses the terminal session's namespace at lookup time, preventing decisions from identically named convoys in another namespace from leaking into fleet/TUI rows.

Added `fleet_list_scopes_crew_session_placement_decisions_by_namespace_and_convoy`, which publishes conflicting decisions for `flotilla/shared` and `zzz/shared` and verifies the local session receives only the `flotilla` decision. The exact local fmt, clippy, and full workspace test commands pass.

rjwittams:
Fixed in c20464ed. Convoy admission now records `PlacementDecision` through a write-once `ConvoyStatusPatch::SetPlacementDecision` applied by the retrying `apply_status_patch` path, so a concurrent bootstrap status update no longer causes admission to report a stale-resourceVersion failure or lose the decision.

Added a patch contract test proving the first decision is sticky and concurrent status fields are preserved. The previously intermittent `sqlite_backed_runtime_reconciles_convoy_create_into_namespace_view` test passes with this path, as do the exact local fmt, clippy, and full workspace test commands.

rjwittams:
Verified the two non-blocking observations. Neither changes this slice's behavior or acceptance criteria: the legacy placement `fields["host"]` fallback remains compatible for existing data/tests, and raw resource deletion is intentionally unconditional while returning the identity fetched immediately before deletion. No further code change is warranted here.

Thanks for re-verifying the retry path and confirming there are no blocking findings.

rjwittams:
Confirmed the latest main integration was mechanical and preserves all reviewed fixes. Since that review, `origin/main` advanced once more with only the Claude workflow trigger fix (#1221); it merged cleanly without touching the implementation. The exact local formatting, clippy, and full workspace test gates pass on the new merge head. No further code change is required for this review.

rjwittams:
Fixed in the new head. Human `resource list/get` formatting now treats the host-list query as optional enrichment: if that secondary query fails or returns an unexpected value, it renders the already-fetched resource as canonical raw JSON instead of failing the command.

Added `human_resource_rendering_falls_back_to_raw_json_when_host_lookup_fails` to cover the transient-error path. The exact local formatting, clippy, and full workspace test gates pass.

rjwittams:
Confirmed. The graceful fallback is now covered and all earlier review fixes remain intact. The exact local gates and all four GitHub checks pass on `1aa6c727`; no further changes are needed.

PR #1220, opened 2026-07-28
fix: make fresh clone retries idempotent

### LINKED ISSUE #1218: Fresh-clone provisioning is not retry-idempotent: collides with its own partial directory
rail-render-tests failed three times with `fatal: destination path '~/dev/flotilla-repos/andamento' already exists and is not an empty directory`. The first was stale debris, but attempts two and three ran against a **pre-cleaned** directory — the fresh-clone actuation creates the dir, something fails or re-drives, and the retry collides with its own partial clone. Under level-triggered re-drive (#1180) every actuation must be retry-idempotent; clone isn't. Fix shape: clone into a temp sibling and atomically rename into place on success, removing partials on failure — or detect-and-resume/clean a partial dir (no .git or incomplete = remove and re-clone; valid clone at right remote = adopt). Test: fail a clone mid-flight, assert the retry converges. Evidence convoy: rail-render-tests (andamento#50), feta, 2026-07-28.

rjwittams:
## Summary

- clone fresh checkouts into a deterministic sibling staging directory
- remove partial staging directories when preparation fails
- atomically publish a fully prepared clone to its final checkout path
- cover interrupted clone actuation followed by a successful level-triggered retry

## Root cause

Fresh-clone actuation passed the final checkout path directly to `git clone`. If Git created that directory and then failed, the next reconciliation treated the partial directory as an existing checkout and could not reuse or replace it. In an embedded vessel checkout, `git -C` could also walk up and inspect the enclosing repository, producing a misleading origin mismatch.

## Validation

Closes #1218

### DIFF
--- a/crates/flotilla-daemon/src/runtime.rs
+++ b/crates/flotilla-daemon/src/runtime.rs
@@ -1523,7 +1523,10 @@ impl CheckoutRuntime for CheckoutControllerRuntime {
                 return Err(cleanup_failed_checkout(&*runner, &staging_path, error).await);
             }
         };
-        if let Err(error) = runner.run("mv", &["-T", &staging_path, target_path], Path::new("/"), &ChannelLabel::Noop).await {
+        // Checkout actuation is serialized and recovery established that the
+        // target is absent. Plain `mv` is therefore an atomic sibling rename
+        // while remaining portable across GNU and BSD hosts.
+        if let Err(error) = runner.run("mv", &[&staging_path, target_path], Path::new("/"), &ChannelLabel::Noop).await {
             return Err(cleanup_failed_checkout(&*runner, &staging_path, error).await);
         }
         Ok(PreparedCheckout { commit, branch_provenance: CheckoutBranchProvenance::PreExisting })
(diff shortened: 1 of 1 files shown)


rjwittams:
Addressed all three findings in 997ca4f6:

- fresh-clone finalization now removes the deterministic staging sibling as well as the final target, with a regression test for deleting an interrupted clone;
- publication uses `mv -T` so an unexpectedly existing target fails closed instead of nesting the staging directory;
- the interrupted-clone test now asserts that failure cleanup removes staging immediately, before the retry sweep can mask it.

The focused fresh-clone suite, pinned formatting, daemon Clippy, and the sandbox-safe full workspace tests all pass locally.

rjwittams:
Fixed in d1a4b3d8. The publish step now uses portable plain `mv`, with a comment documenting the invariants that make it an atomic sibling rename: checkout actuation is serialized and `recover_existing_fresh_clone` established that the target is absent. This preserves macOS/BSD support while retaining the staging-directory idempotency fix.

All six fresh-clone tests, pinned formatting, and daemon Clippy pass locally after the change.

rjwittams:
Implemented the suggested fail-closed publish in 627cfffe. `create_fresh_clone` now uses `tokio::fs::rename` for the host-local sibling publish, preserving atomicity and GNU/BSD portability while failing if the destination invariant is violated. Publish failures still flow through staging cleanup with contextual error text.

All six fresh-clone tests, pinned formatting, and daemon Clippy pass locally.

rjwittams:
Confirmed. No further changes are needed after the atomic host-local rename update; all review findings are resolved and the final CI run is green.

PR #1222, opened 2026-07-28
fix: single-flight shared clone provisioning

### LINKED ISSUE #1218: Fresh-clone provisioning is not retry-idempotent: collides with its own partial directory

rjwittams:
## Summary

- single-flight base-repository clone actuation by target path across controller runtime restarts
- clone through a cleaned staging sibling, atomically publish, and adopt a matching clone that wins an external race
- keep operational clone failures in a visible retrying state instead of latching the clone, vessel, and convoy Failed
- cover concurrent duplicate requests, interrupted clone cleanup/retry, and fail-once reconciler convergence

## Root cause

#1220 made per-checkout fresh clones sequentially retry-idempotent, but shared base-repository clones use `CloneControllerRuntime`. Duplicate in-flight requests on that path both invoked `git clone` against the final target; one won and the loser reported an already-existing destination. `CloneReconciler` then made every runtime error terminal, so the transient loser wedged provisioning even when the filesystem had converged.

## Validation

Closes #1218

### DIFF
--- a/crates/flotilla-controllers/src/reconcilers/clone.rs
+++ b/crates/flotilla-controllers/src/reconcilers/clone.rs
@@ -69,7 +69,8 @@ where
         };
         Ok(match result {
             Ok(default_branch) => CloneDeps::Ready { default_branch },
-            Err(err) => CloneDeps::Retrying(err),
+            Err(err) if phase == ClonePhase::Pending => CloneDeps::Retrying(err),
+            Err(err) => CloneDeps::Failed(err),
         })
     }
 
--- a/crates/flotilla-controllers/src/reconcilers/clone.rs
+++ b/crates/flotilla-controllers/src/reconcilers/clone.rs
@@ -84,13 +84,6 @@ where
         let patch = if matches!(phase, ClonePhase::Pending | ClonePhase::Cloning) {
             match deps {
                 CloneDeps::Ready { default_branch } => Some(CloneStatusPatch::MarkReady { default_branch: default_branch.clone() }),
-                CloneDeps::Retrying(message)
-                    if obj.status.as_ref().is_some_and(|status| {
-                        status.phase == ClonePhase::Cloning && status.message.as_deref() == Some(message.as_str())
-                    }) =>
-                {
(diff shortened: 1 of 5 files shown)


rjwittams:
Addressed in 7387a03a.

- Operational failures now get one retry: the first failure records `Cloning`/retrying and schedules a one-second redrive; if the redrive fails again, the clone becomes `Failed` and propagates to the vessel/convoy. The persistent-auth test now verifies that bounded transition, while the fail-once test still proves transient convergence.
- Added `clone_runtime_adopts_a_matching_clone_that_wins_an_external_flight`, using independent flight registries so the loser reaches rename failure and adopts the winner.
- Rename failure and adoption failure context are now chained instead of replacing one another.
- Extracted shared origin verification for base clones and fresh-checkout clones.

Pinned formatting, full workspace Clippy, all focused tests, and the sandbox-safe full workspace suite pass after these changes.

rjwittams:
Fixed in a1742e2b. The unchanged-message dedup arm was indeed unreachable after bounding operational failures to one retry, so it has been removed. The focused clone reconciler suite, pinned formatting, and full workspace Clippy pass on the final cleanup.

rjwittams:
Confirmed. No further code changes are needed after the final-state review; the final head remains green locally and all prior findings are resolved.

PR #1223, opened 2026-07-28
fix: set PATH in cargo-sweep systemd unit

rjwittams:
The daily sweep unit invokes `cargo-sweep`, which spawns `cargo`. On hosts where the systemd user manager's PATH does not include `~/.cargo/bin` — stock Debian, LXC containers (observed on udder) — every scheduled sweep failed with `No such file or directory (os error 2)` while manual runs succeeded, because interactive shells do have the cargo path.

Carry the PATH in the unit itself so the schedule works without per-host `environment.d` configuration. Found installing the sweep schedule on udder (which had never had it — 23G target dir on a 50G rootfs).

### DIFF
--- a/config/launchd/org.flotilla.cargo-sweep-mtime.plist
+++ b/config/launchd/org.flotilla.cargo-sweep-mtime.plist
@@ -8,7 +8,7 @@
   <array>
     <string>/bin/sh</string>
     <string>-c</string>
-    <string>exec "$HOME/.local/libexec/flotilla/cargo-sweep-mtime.sh"</string>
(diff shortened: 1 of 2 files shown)


rjwittams:
Both review observations addressed:

- `/sbin:/bin` appended to the unit PATH for non-usrmerged hosts (udder itself is usrmerged, but the cost is zero).
- The launchd agent had the identical latent failure — and checking kiwi showed the schedule was never actually loaded there, so it had simply never fired to fail. PATH is now set in the plist's `sh -c` invocation the same way.

PR #1226, opened 2026-07-28
fix: mount daemon-matched cli into docker vessels

### LINKED ISSUE #1215: Contained vessels get the daemon socket but no flotilla CLI to speak to it
Verified 2026-07-28: docker vessels bind-mount the daemon socket at `/run/flotilla.sock` (Phase-D plumbing from #531, alive in the current path — confirmed via docker inspect of the smoke container) but the crew image contains **no `flotilla` binary** (`which flotilla` empty in `flotilla-crew:2026-07-26.1`). A contained crew cannot run `flotilla crew complete`, `flotilla hook`, or any CLI verb — the socket is a phone with no handset. This is the next containment wall after credential seeding (#1140) and would have blocked the smoke crew at completion time.

**Direction**: bind-mount the host's own binary (`flotilla` adjacent to the running `flotillad`) into the container at a PATH location, rather than baking it into the image — the image is built occasionally while the daemon moves daily, so a baked binary guarantees no-compat wire skew (#1195's disease inside every vessel); the host's binary is daemon-matched by construction. Verify glibc compat of the host binary inside the image's userland (both modern; static/musl build is the fallback if not). Also mount-note: the socket mount is invisible in the `flotilla.provisioned_mounts` label (workspace-domain only) — include infra mounts in whatever surface #1209 gives mounts, for legibility.

Chain: #1140 (seeding) → this → containment smoke round 4 completes end-to-end. Batches naturally with #1195 (same paired-CLI principle) in a future round.

rjwittams:
## Summary

- resolve the `flotilla` CLI adjacent to the running Linux `flotillad` binary
- bind-mount that exact binary read-only at `/usr/local/bin/flotilla` for Docker-provisioned vessels
- include daemon socket and requested infrastructure mounts in `flotilla.provisioned_mounts` metadata
- fail Docker provisioning clearly when adjacent binary injection is unavailable

## Verification

- verified the feta host binary is x86-64 glibc-linked and runs successfully inside `forgejo.lab.flotilla.work/image-builder/flotilla-crew:2026-07-26.1` (Ubuntu 24.04) as `flotilla 0.1.0`

Closes #1215

### DIFF
--- a/crates/flotilla-core/src/providers/environment/docker.rs
+++ b/crates/flotilla-core/src/providers/environment/docker.rs
@@ -72,13 +72,18 @@ impl EnvironmentProvider for DockerEnvironmentProvider {
     async fn create(&self, id: EnvironmentId, image: &ImageId, opts: CreateOpts) -> Result<EnvironmentHandle, String> {
         let container_name = format!("flotilla-env-{}", id);
 
-        let socket_str = opts.daemon_socket_path.to_string();
+        let mut provisioned_mounts = Vec::with_capacity(opts.provisioned_mounts.len() + 1);
+        provisioned_mounts.push(ProvisionedMount::new(
+            opts.daemon_socket_path.as_path().to_path_buf(),
+            CONTAINER_SOCKET_PATH,
+            ProvisionedMountMode::Rw,
+        ));
+        provisioned_mounts.extend(opts.provisioned_mounts);
         let env_id_str = id.to_string();
         let image_str = image.as_str().to_string();
         let label_val = format!("flotilla.environment={}", id);
         let mounts_label_val =
-            format!("flotilla.provisioned_mounts={}", serde_json::to_string(&opts.provisioned_mounts).map_err(|err| err.to_string())?);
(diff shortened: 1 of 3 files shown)


rjwittams:
Addressed the three concrete gaps in b10768c4:

- Requested mounts targeting `/run/flotilla.sock` or `/usr/local/bin/flotilla` now fail early with clear flotilla-level errors, before Docker/provider invocation, with regression coverage for both reserved paths.
- Added a fresh-state provisioning test (without `with_flotilla_binary_path`) that verifies the real adjacent-CLI resolution failure retains the `flotilla CLI unavailable for docker environment provisioning` context.
- `adjacent_flotilla_binary` now rejects files with no executable bits on Unix, with positive and negative tests.

The direct `std::env::current_exe()` call is intentional here: this is self-introspection of the running daemon binary rather than configurable/provider environment input, and each daemon therefore resolves its own per-host sibling CLI under multi-host operation.

The exact local CI gates all pass: pinned nightly format check, workspace/all-target clippy with warnings denied, and the locked full workspace test suite.

PR #1227, opened 2026-07-28
fix: extend heartbeat test timeout

### LINKED ISSUE #1197: Flaky: heartbeat_task_updates_host_status_without_socket_server times out under parallel test load
Fails with 'timed out waiting for host status' (runtime.rs:3163) in full `cargo test --workspace` runs on a loaded desk machine; passes reliably in isolation (`--lib heartbeat_task`). Seen twice on 2026-07-28 in otherwise-green runs. Likely a too-tight wait deadline in the test vs scheduler contention — wants a generous deadline or a virtual-clock rewrite (#1176's injectable-clock direction).

rjwittams:
## Summary

- add a timeout-aware variant of the runtime test polling helper
- give the spawned heartbeat status assertion a 30-second wall-clock budget
- preserve the existing five-second timeout for unrelated polling assertions

A paused Tokio clock is not appropriate for this test because the heartbeat performs a real `spawn_blocking` disk-space probe; virtual time could expire while that work is still queued under scheduler contention.

## Testing

- 20 repeated focused heartbeat test runs under concurrent desk load

Closes #1197

### DIFF
--- a/crates/flotilla-daemon/src/runtime.rs
+++ b/crates/flotilla-daemon/src/runtime.rs
@@ -3738,12 +3738,20 @@ mod tests {
             .expect("read maximum resource event rowid")
     }
 
-    async fn wait_until<F, Fut>(mut condition: F)
+    async fn wait_until<F, Fut>(condition: F)
     where
         F: FnMut() -> Fut,
         Fut: std::future::Future<Output = bool>,
     {
-        let deadline = tokio::time::Instant::now() + Duration::from_secs(5);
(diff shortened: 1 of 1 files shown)


rjwittams:
Thanks for verifying the scheduler-contention path and the unchanged default behavior. I’m leaving the peer-heartbeat test at its existing five-second budget because it has not exhibited the reported flake; the timeout-aware helper makes the same targeted adjustment straightforward if evidence emerges later.

Issue #1150, opened 2026-07-27
Landing convoys respond to conflicts: message the warm crew to rebase


rjwittams:
## Parent

Split from the #1149 grill (Robert: "we do need 'wake up to rebase' relatively soon, but it's not really part of this exact task").

## What to build

A convoy in `Landing` whose change request goes CONFLICTING (an unrelated merge landed) gets its warm crew messaged automatically with a rebase brief — the state-driven turn-messaging pattern: the lifecycle reconciler observes `Landing` + CONFLICTING on already-observed CR state, delivers a rebase instruction into the crew session (the `resume --prompt` path), and records the delivery so it never double-sends. No event/webhook plumbing required — this rides periodic CR observation.

This automates what the governor did by hand on PR #1134 (2026-07-27): resume the warm crew with "rebase onto main, resolve additively, verify checks". The delivered brief should carry the same contract: rebase, pinned CI gates, push same branch, no new PR.

## Acceptance criteria

- [ ] A Landing convoy whose PR turns CONFLICTING receives exactly one rebase message per conflict episode (idempotent via recorded delivery)
- [ ] The crew's successful rebase → PR MERGEABLE again with no human action
- [ ] A convoy already mid-rebase is not re-messaged; a second, later conflict is
- [ ] Works through the same delivery path the implement-review turn handoff uses (#1147 family)

## Blocked by

#1113 enactment (the `Landing` state itself — in flight as convoy-lifecycle-landing-landed).

rjwittams:
Design input from Robert for the eventual DAG-rebase side of this: rebases of a branch stack use **generations** rather than immediate force-pushes — new refs named `<dag-descr>-N/<branch-descr>` per generation. Approval for a stack rebase is correspondingly complex: the whole DAG must work *and* each branch must be coherent on its own. A generation rebase therefore declares **new refs** as its output — a different output structure from the single per-repo `target_ref` that #1155 introduces, so design it as its own output declaration rather than stretching `target_ref`.

Also folded in from #1081: "rebase obligation surfaced at settlement" (stack above a landed target moved) belongs here, not in the fork tracer — it needs stack/DAG declarations to be computable.

rjwittams:
## Ruling (Robert, 2026-07-28): mechanism rejected; behavior retained as the acceptance scenario

[...]

PR #1228, opened 2026-07-28
feat: observe change-request mergeability in checkout integrations

rjwittams:
## Summary

- persist change-request identity, state, mergeability, and target ref in checkout integration observations
- treat semantic change-request transitions as status changes while ignoring timestamp-only refreshes
- query forge mergeability during periodic checkout integration inspection, with coverage for conflicting observations

This is the mechanism-neutral observation slice needed by the future declarative workflow and state-event substrates in #1232 and #1233. Delivery semantics are intentionally out of scope here.

## Validation

Part of #1150

### DIFF
--- a/crates/flotilla-controllers/src/reconcilers/checkout.rs
+++ b/crates/flotilla-controllers/src/reconcilers/checkout.rs
@@ -169,10 +169,10 @@ where
         } else if obj.status.as_ref().is_some_and(|status| status.phase == CheckoutPhase::Ready) {
             match deps {
                 CheckoutDeps::Integration { status } => {
-                    Some(CheckoutStatusPatch::UpdateIntegration { integration: status.as_ref().clone() })
+                    Some(CheckoutStatusPatch::UpdateIntegration { integration: Box::new(status.as_ref().clone()) })
                 }
                 CheckoutDeps::Failed(message) => Some(CheckoutStatusPatch::UpdateIntegration {
-                    integration: CheckoutIntegrationStatus {
+                    integration: Box::new(CheckoutIntegrationStatus {
                         clean: flotilla_resources::IntegrationCondition::builder()
                             .value(flotilla_resources::ConditionValue::Unknown)
                             .details(vec![message.clone()])
@@ -189,7 +189,8 @@ where
                             .observed_at(now.to_rfc3339())
                             .build(),
                         landed_evidence: None,
-                    },
+                        change_request: None,
+                    }),
                 }),
                 CheckoutDeps::None | CheckoutDeps::Ready { .. } | CheckoutDeps::Waiting => None,
             }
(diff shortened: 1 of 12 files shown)


rjwittams:
Fixed both lifecycle races in 71df7df9.

- `LandingConflictObservation` now retains the current conflicts and the repositories whose observations are unknown. An active episode resolves against only its recorded repository/PR identities, so an unrelated unknown checkout cannot strand it or block a later conflict. The lifecycle regression now keeps a second checkout unknown while resolving and re-arming the recorded episode.
- Resolution now requires the terminal delivery acknowledgement to have been persisted. If the PR becomes mergeable first, reconciliation holds the episode without advancing until `delivered_at` is recorded; the regression covers that ordering too.

I left ambiguous multi-coder selection conservative: the default workflow has one eligible coder, and choosing arbitrarily when a future template supplies multiple candidates would risk routing a repository rebase to the wrong warm session.

rjwittams:
Confirmed. The re-review matches the intended post-fix invariants and reports no new blocking issue. I am keeping the unique-coder requirement conservative for this slice; if multi-coder routing becomes a supported workflow shape, surfacing that ambiguity should be designed alongside an explicit routing rule rather than selecting a session arbitrarily.

rjwittams:
Fixed in the follow-up commit. When an evidence-backed landing is latched, `UpdateIntegration` now retains an existing terminal (`Merged` or `Closed`) change-request observation if a later probe returns no change request. Open observations are not latched. The checkout status-patch regression now verifies that landed evidence and the merged observation survive together.

rjwittams:
Fixed in 919a13e4. Teardown-gate normalization now applies the same evidence-backed latch to the terminal change-request observation before the observation-equality check, so timestamp-only probes do not rewrite checkout status on every reconcile. Added a focused regression proving the normalized observation deduplicates against the persisted status.

Re-ran the exact local gates: nightly fmt check, workspace clippy with warnings denied, and the locked workspace test suite; all pass.

rjwittams:
Fixed in d74bcd6e. `verify_convoy_teardown_gate` now calls the same `latch_evidence_backed_integration` helper as adopted-checkout refresh, removing the duplicated partial latch entirely.

The regression now drives `verify_convoy_teardown_gate` end-to-end twice with a persisted terminal merged observation and a fresh zero-commit probe that cannot see the PR. It asserts the second probe leaves the checkout resource version unchanged, so the test catches repeated status-patch churn at the actual call site.

Re-ran the exact gates after this correction: nightly fmt check, workspace clippy with warnings denied, and the locked workspace test suite; all pass.

rjwittams:
Fixed in feb2ed97. `convoy_change_requests_settled` now normalizes the stale reprobe with the same `latch_evidence_backed_integration` helper before both persistence and its Landing→Landed decision, so all three observation consumers share one latch implementation.

Added an end-to-end regression that seeds stale evidence-backed `Merged` status, returns no PR from branch discovery, and asserts settlement remains true while the persisted terminal observation survives.

Re-ran the exact nightly fmt check, workspace clippy with warnings denied, and locked workspace test suite; all pass.

rjwittams:
Reviewed. Thanks for confirming the three call sites and persisted latch now agree and that no blocking issues remain. The suggested cross-crate consolidation is a reasonable future refactor, but it is intentionally outside this mechanism-neutral observation slice; no additional code change for this PR.

PR #1229, opened 2026-07-28
fix: enforce client wire generation handshake

### LINKED ISSUE #1195: Cross-host verbs must resolve a wire-compatible CLI via the daemon, not $PATH luck
## What happened (2026-07-28)

`flotilla attach --host feta --transient docker-crew-smoke` from kiwi failed with `daemon read error: failed to parse message: missing field 'command'`. The cross-host attach chain ssh-executes `flotilla` on the target host, which resolved to a **weeks-stale `~/.cargo/bin/flotilla`** — the deploy ritual rebuilds `target/debug` and bounces daemons but never updates cargo-bin on Linux hosts. Stale CLI + new daemon + no-compat wire = envelope decode error. Fixed by hand-copying binaries; the failure mode remains.

## The think-about (Robert): the daemon should determine which binary speaks to it

A client crossing hosts has no business resolving its peer executable from `$PATH`. Candidate shapes, roughly increasing in ambition:

1. **Version handshake + clear refusal**: every connection exchanges a wire-generation stamp; mismatch produces "client generation X, daemon generation Y — rebuild or use the daemon's paired CLI at <path>" instead of a serde error. Cheap, catches every skew instance (this session hit TUI-skew, CLI-skew, and replica-skew variants).
2. **Daemon-advertised paired CLI**: the Host resource already publishes `daemon_generation` and the daemon knows its own binary path (`flotillad` sits next to `flotilla` — `connect_or_spawn` already relies on adjacency). Remote invocation asks the daemon "exec your paired CLI" rather than naming a binary.
3. The #917 world (CI-built per-platform artifacts + deploy-to-fleet) makes the skew window small, but the handshake in (1) is wanted even then.

Related: #917 (artifact deploys), #1183 (routed verbs — routing through the daemon connection instead of ssh-exec-ing a remote binary dissolves this class entirely: one more argument for tunneled verbs over remote exec).

## Slice 1 contract (ruled, dispatchable)

Build **shape 1 only**: every client↔daemon connection exchanges a wire-generation stamp before any envelope decode; on mismatch the client gets a clear, actionable refusal ("client generation X, daemon generation Y — rebuild or use the daemon's paired CLI") instead of a serde error. Cover all three observed skew surfaces: CLI over socket, TUI over socket, and peer/replica connections (peers already exchange generations — reuse that stamp; the gap is client connections).

Acceptance:
- [ ] A deliberately stale client binary against a current daemon produces the legible refusal, not `failed to parse message`
- [ ] Matching generations add no perceptible handshake latency and no protocol churn for current clients
- [ ] Peer connections keep their existing generation behavior (no regression)

Shapes 2 (daemon-advertised paired CLI) and the #917/#1183 directions stay out of scope for this slice.

rjwittams:
## Summary

- require the existing Hello exchange and matching build generation before accepting any client RPC payloads
- report both client and daemon generations with an actionable rebuild/paired-CLI recovery
- route one-shot agent hooks through the same generation-checked client used by CLI and TUI connections
- remove the legacy request-first server path while preserving the peer protocol handshake

## Why

A protocol version only changes when someone remembers to bump it, so independently built client and daemon binaries could deserialize the same wire payload differently. The Hello exchange already carries each binary's build stamp; this makes that stamp the mandatory wire generation and refuses mismatches before dispatch.

## Validation

Closes #1195

### DIFF
--- a/crates/flotilla-client/src/lib.rs
+++ b/crates/flotilla-client/src/lib.rs
@@ -89,11 +89,22 @@ async fn do_client_hello_with_surface(session: &MessageSession, surface: Option<
         Some(Message::Hello { protocol_version, display_name, .. }) if protocol_version != PROTOCOL_VERSION => {
             let daemon_build = flotilla_protocol::hello_build_id(&display_name).unwrap_or("unknown");
             Err(format!(
-                "daemon protocol version mismatch: daemon has {protocol_version} ({daemon_build}), client has {PROTOCOL_VERSION} ({})",
-                BUILD_ID
+                "daemon protocol version mismatch: client generation {} uses protocol {PROTOCOL_VERSION}, daemon generation \
+                 {daemon_build} uses protocol {protocol_version} — rebuild or use the daemon's paired CLI",
+                BUILD_ID,
             ))
         }
-        Some(Message::Hello { display_name, .. }) => Ok(flotilla_protocol::hello_build_id(&display_name).map(str::to_owned)),
+        Some(Message::Hello { display_name, .. }) => {
+            let daemon_generation = flotilla_protocol::hello_build_id(&display_name).unwrap_or("unknown");
+            if daemon_generation != BUILD_ID {
+                return Err(format!(
+                    "wire generation mismatch: client generation {}, daemon generation {daemon_generation} — rebuild or use the \
+                     daemon's paired CLI",
+                    BUILD_ID,
+                ));
+            }
+            Ok(Some(daemon_generation.to_owned()))
+        }
         Some(other) => Err(format!("expected Hello reply, got: {other:?}")),
         None => Err("connection closed before Hello reply".into()),
     }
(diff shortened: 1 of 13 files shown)


rjwittams:
Addressed both points in 970fb1cd:

- wire-generation matching now fails closed when either stamp is `unknown`, with a protocol-level regression covering `unknown`/`unknown`
- removed the superseded `build_mismatch`/`warn_on_build_mismatch` helpers and their unreachable TUI/CLI call sites, while retaining the ordinary reconnect banner

Revalidated with the exact formatting, workspace clippy, and locked workspace test commands.

rjwittams:
Acknowledged. The fail-closed behavior is intentional: an exported-source build must supply `FLOTILLA_BUILD_ID` if it needs interoperable client/daemon binaries. The handshake error itself reports `client generation unknown, daemon generation unknown` and the rebuild/paired-CLI recovery, so it diagnoses the condition at the point of failure; I don't think an additional startup warning would improve this slice.

PR #1231, opened 2026-07-28
fix: make placement ordering legible

### LINKED ISSUE #1224: Placement: viable-but-not-selected candidates are invisible and prefer-local is hardcoded
## Observed (2026-07-28, three specimens)

After feta's daemon restart onto `07cfc9d6`, every subsequent dispatch (`rail-render-tests`, the four-convoy bug batch, `pr-adoption-slice1`) produced a `placement_decision` whose candidate set contains only kiwi-docker (refused: no codex adapter in image), udder host-direct (refused: no codex adapter), and kiwi host-direct (chosen). **Feta appears nowhere — not chosen, not refused.** All work piled onto kiwi despite feta being the preferred, healthier placement.

At decision time feta was demonstrably eligible:
- Host `d783f8a7` status: `ready: true`, heartbeat fresh (~20s), fleet diagnosis `agree`
- capabilities: `agent_adapters: ["claude-code", "codex"]`, `terminal_pools: ["cleat", "passthrough", "shpool"]` (after the env fix; even before it, `codex` — the adapter the workflow required — was present)
- `PlacementPolicy host-direct-d783f8a7` present in the store and byte-shape-identical to udder's `host-direct-b2aac222` (which **was** evaluated and refused): same keys, same spec shape, only host_ref differs

So the exclusion is not policy shape, not readiness, not capabilities. Whatever enumerates candidates never saw feta's placement.

## Hypothesis: mixed-version skew in placement advertising

Timing is exactly correlated with feta becoming the only host on `07cfc9d6` while the admitting daemon (kiwi) runs `3d85e9a5`. If candidate placements are advertised by each daemon (awareness/host events) rather than read purely from PlacementPolicy resources, a decode failure or shape change across the version gap would drop feta silently — the #1195 disease in the placement path. A kiwi/udder bump to `07cfc9d6` is in progress; result will be recorded here (returns → skew class, confirming #1195's handshake-refusal need covers resource/awareness channels too; persists → real enumeration bug).

Either way there is a distinct defect: **silent exclusion**. The legibility contract (#1209/#1217) that made this diagnosable should extend to non-candidacy — a healthy host with a policy in the store that produces neither a selection nor a refusal entry should be visible in the decision (e.g. `excluded: not advertised / decode failed / policy unresolvable`), not simply absent.

## Acceptance
- [ ] Root cause of feta's exclusion identified (skew vs enumeration logic), regression-tested
- [ ] `placement_decision` accounts for every host with a matching-kind policy in the store: chosen, refused-with-reason, or excluded-with-reason — silent absence is impossible

rjwittams:
## Summary

- add integer `priority` to `PlacementPolicy` and order viable candidates by descending priority before the existing local/kind/name fallback
- record every viable loser in `placement_decision.viable_not_selected` with the ordering reason
- expose viable alternatives in fleet CLI summaries and convoy/vessel TUI details
- cover manifest/resource round trips and an in-process dispatch with selected, refused, and viable-not-selected candidates

## Root cause

Candidate enumeration was working: feta passed admission and entered the viable set. The resolver sorted viable policies, selected the first with `into_iter().next()`, and discarded every remaining viable policy. The sort also unconditionally preferred the admitting daemon's local host, so kiwi consistently outcompeted feta without leaving evidence in the decision.

## Validation

Closes #1224

### DIFF
--- a/crates/flotilla-controllers/src/reconcilers/vessel.rs
+++ b/crates/flotilla-controllers/src/reconcilers/vessel.rs
@@ -1119,6 +1119,7 @@ mod tests {
             policy_name: "host-direct-test".to_string(),
             target_host: PlacementTargetHost { reference: "01HXYZ".to_string(), display_name: "kiwi".to_string() },
             refused_candidates: Vec::new(),
+            viable_not_selected: Vec::new(),
         };
 
         assert_eq!(
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -4,6 +4,7 @@
 //! and broadcasts events — all within the same process.
 
 use std::{
+    cmp::Reverse,
     collections::{BTreeMap, BTreeSet, HashMap, HashSet},
     panic::AssertUnwindSafe,
     path::{Path, PathBuf},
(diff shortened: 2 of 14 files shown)


rjwittams:
Addressed the duplication nit in 550906cf: the placement sort now calls `placement_host_ref`, keeping host extraction aligned with the ordering-reason path. The focused placement suite and workspace Clippy pass after the refactor. The CLI-count test gap is pre-existing and non-blocking as noted; this PR covers the new detailed viable-candidate rendering and the persisted decision end to end.

PR #1234, opened 2026-07-28
fix: surface agent adapter regressions

### LINKED ISSUE #1225: Host capabilities silently degrade with the daemon launcher's environment: ssh-restarted feta lost the claude-code adapter
## Observed (2026-07-28)

Feta's daemon, restarted via `ssh -f feta 'setsid ./target/debug/flotillad ...'`, published Host capabilities of `agent_adapters: ["codex"]` — no `claude-code` — because `claude` lives in `~/.local/bin`, which is on interactive-shell PATH but not the non-interactive ssh PATH the daemon inherited. The previous daemon generation (started from a fuller environment) had advertised claude-code all day. Restarting with `PATH=$HOME/.local/bin:...` prepended restored it.

Consequence: any workflow requiring claude-code would refuse placement on feta, and the degradation is invisible — no warning, just a quieter capabilities list on a Host resource nobody re-reads after bootstrap.

## Why this is a defect and not operator error

Discovery already goes through injected `EnvVars`/`CommandRunner` precisely because the daemon's own environment can be wrong. But the *fact base* still bottoms out in the launcher's PATH, so an operationally identical restart (same binary, same host) can silently shed adapters. The daily-driver deploy ritual restarts daemons over ssh constantly; this will recur.

Candidate directions (pick in-issue):
1. **Loud delta**: daemon compares its freshly-discovered capabilities against the previous generation's Host status and WARNs (or surfaces via fleet diagnosis) on any regression — cheap, catches every variant.
2. **Launcher-independent probing**: discovery consults a login-shell PATH (or a configured binary-search list in daemon.toml) instead of the inherited env, making capabilities a property of the host rather than of the spawn context.

(1) is wanted even if (2) lands. Related: the deploy ritual itself should stop being hand-rolled ssh one-liners — #917.

## Acceptance
- [ ] A daemon restart that would shed a previously-advertised adapter is loud (log + fleet-visible), not silent
- [ ] Test covers capability regression across generations

rjwittams:
## Summary

- compare freshly discovered agent adapters with the durable Host status from the previous daemon generation
- emit a structured warning naming adapters lost across a restart
- retain a fleet-visible `CapabilityRegression` Host condition for the new runtime, degrading the host until another generation establishes a non-regressed baseline
- cover the `claude-code + codex` to `codex` restart regression

## Root cause

Host heartbeat publication replaced the previous capability set without comparing generations. Because discovery bottoms out in the launcher's environment, a daemon restarted from a restricted SSH `PATH` could silently overwrite a previously advertised adapter set.

## Validation

Closes #1225

### DIFF
--- a/crates/flotilla-daemon/src/runtime.rs
+++ b/crates/flotilla-daemon/src/runtime.rs
@@ -211,45 +211,56 @@ impl RuntimeHealth {
     }
 }
 
-fn agent_adapter_regression_condition(
+struct AgentAdapterCapabilityAssessment {
+    baseline: BTreeSet<String>,
+    regression: Option<HostCondition>,
+}
+
+fn assess_agent_adapter_capabilities(
     previous: Option<&HostStatus>,
     current: &BTreeSet<String>,
     health: &DaemonHealthIdentity,
-) -> Option<HostCondition> {
-    let previous = previous?;
+) -> AgentAdapterCapabilityAssessment {
+    let Some(previous) = previous else {
+        return AgentAdapterCapabilityAssessment { baseline: current.clone(), regression: None };
+    };
+    let baseline = match &previous.agent_adapter_baseline {
+        Some(baseline) => baseline.clone(),
+        None => match previous.agent_adapters() {
+            Ok(adapters) => adapters,
+            Err(error) => {
+                warn!(%error, "cannot compare agent adapter capabilities with previous daemon generation");
+                return AgentAdapterCapabilityAssessment { baseline: current.clone(), regression: None };
+            }
+        },
+    };
     let same_daemon = previous.daemon_generation == health.generation && previous.daemon_started_at == Some(health.started_at);
     if same_daemon {
-        return None;
+        return AgentAdapterCapabilityAssessment { baseline, regression: None };
     }
-    let previous_adapters = match previous.agent_adapters() {
-        Ok(adapters) => adapters,
-        Err(error) => {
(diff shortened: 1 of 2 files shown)


rjwittams:
Fixed in `24b51051`. Host status now persists the last non-regressed adapter inventory separately from the truthful current capability set, so generation 3 under the same restricted PATH compares against the original `claude-code + codex` baseline, warns again, and remains degraded. The regression test now covers that repeated restart and also verifies that another heartbeat within generation 2 does not repeat the warning.

I kept the delta specific to agent adapters because that is #1225's explicit acceptance criterion; terminal-pool/docker baselines would need their own semantics and acknowledgement story rather than being folded into this adapter bug implicitly.

rjwittams:
Covered in `5a18b0a6`. The end-to-end restart test now clears generation 1's persisted `agent_adapter_baseline` before generation 2, simulating a Host status written by the pre-field daemon. Generation 2 must therefore seed from the legacy `agent_adapters` capability, and the existing generation-2/3 assertions prove that fallback preserves and reuses the non-regressed baseline.

`HostStatusPatch::Heartbeat` remains intentionally unchanged because it is not the production heartbeat publisher and should preserve status fields it does not own.

rjwittams:
Confirmed. `LocalProvisioningProfile` is built once before the runtime starts and cloned unchanged into heartbeat tasks, so an adapter regression cannot recover within the same process. Keeping the condition sticky for that runtime is therefore intentional; a restart re-probes adapters and creates fresh `RuntimeHealth`, which is the only recovery transition. No code change needed for the current invariant.

PR #1240, opened 2026-07-29
fix: preserve operator placement policy fields

### LINKED ISSUE #1236: Daemon policy re-registration stomps operator-authored spec fields (priority reverts to default)
## Observed (2026-07-29)

Applied `spec.priority: 10` to `PlacementPolicy host-direct-d783f8a7` (feta) via `flotilla resource apply` — verified stored (rv 5). Within a minute the policy was back to `priority: None` (rv 6), and a dispatch's `placement_decision` recorded feta as viable-not-selected via "fallback ordering preferred local" — i.e. admission compared equal priorities. The daemon's placement-policy registration loop re-published its canonical spec and silently dropped the operator's field.

Consequence: #1224's merged acceptance criterion — "fleet operators can express 'prefer feta, desk last' purely via policy resources" — is not actually achievable: the ordering input exists (#1231) but cannot survive the owner loop. (The legibility half of #1231 works and is what made this diagnosable: the viable-not-selected reason exposed the equal-priority comparison.)

## Direction

Ownership split within one resource, same pattern as the facts-merge rulings (#1179 family): the registration loop owns the fields it derives (host_ref, pool, checkout mode) and must **merge, not overwrite** — asserting its fields while preserving operator-authored ones (`priority`, future scheduling hints). Manifest-managed policies (`flotilla.work/managed-by`) must not be adopted/stomped at all, consistent with the manifest loop's refusal semantics.

Alternative considered and rejected: moving priority into daemon.toml config — that makes placement preference per-host config instead of fleet-visible data, and contradicts the resources-as-source-of-truth direction.

## Acceptance
- [ ] `apply` of `spec.priority` on a daemon-registered policy survives indefinitely across registration cycles and daemon restarts
- [ ] Registration still corrects drift in the fields it owns (e.g. a hand-edited pool reverts)
- [ ] Regression test: register → operator-apply priority → re-register → priority intact, owned fields asserted
- [ ] A dispatch after the fix lands on the higher-priority host and records the local host as viable-not-selected with the priority reason

rjwittams:
## Summary

- reconcile daemon-registered placement policies by merging daemon-owned topology fields into the live spec
- preserve operator-owned fields such as `priority` and Docker runtime configuration across registration cycles and restarts
- refuse to modify placement policies carrying `flotilla.work/managed-by`
- add SQLite restart coverage for priority preservation, owned-field drift correction, steady-state idempotence, and manifest collisions

## Root cause

Daemon registration treated its canonical placement policy spec as the whole resource definition. Re-publishing that spec could replace operator-authored fields with their defaults. Registration now starts from the stored spec and asserts only the fields it owns: pool, placement strategy, host reference, and checkout mode.

## Impact

Fleet-visible placement priorities now survive daemon registration and restart, so default dispatch can reliably prefer a higher-priority remote host while retaining the existing placement-decision explanation for lower-priority viable candidates.

## Validation

- `cargo test -p flotilla-daemon --locked --lib policy_registration -- --nocapture`
- `cargo test -p flotilla-core --locked --features test-support --test in_process_daemon bare_convoy_start_uses_priority_and_records_every_placement_candidate -- --nocapture`

Closes #1236

### DIFF
--- a/crates/flotilla-daemon/src/runtime.rs
+++ b/crates/flotilla-daemon/src/runtime.rs
@@ -4359,6 +4359,55 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn docker_policy_registration_preserves_runtime_configuration_and_corrects_owned_drift() {
+        let backend = ResourceBackend::InMemory(Default::default());
+        let profile = manual_profile("host-test", true);
+        ensure_default_policies(&backend, NAMESPACE, &profile).await.expect("initial policy registration should succeed");
+
+        let policies = backend.clone().using::<PlacementPolicy>(NAMESPACE);
+        let registered = policies.get(&profile.docker_policy_name()).await.expect("registered docker policy");
+        policies
+            .update(
+                &InputMeta::from(&registered.metadata),
+                &registered.metadata.resource_version,
+                &PlacementPolicySpec::builder()
+                    .pool("operator-edited-pool".to_string())
+                    .priority(20)
+                    .docker_per_vessel(DockerPerVesselPlacementPolicySpec {
+                        host_ref: "operator-edited-host".to_string(),
+                        image: "operator/image:latest".to_string(),
+                        pull_policy: flotilla_resources::DockerImagePullPolicy::Never,
+                        agent_adapters: BTreeSet::from(["codex".to_string()]),
+                        default_cwd: Some("/operator-workspace".to_string()),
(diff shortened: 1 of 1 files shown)


rjwittams:
Addressed the Docker coverage gap in `3e1b0c01`. The new regression applies operator edits to `priority`, image, pull policy, adapters, default cwd, and env, then verifies re-registration preserves them while correcting the daemon-owned pool, host reference, and checkout strategy.

Validated with the focused `policy_registration` tests, exact workspace clippy, pinned format, and the full locked workspace test suite.

PR #1241, opened 2026-07-29
fix: bound peer connection resource usage

### LINKED ISSUE #1239: Peer connection churn leaks fds until accept() death-spiral at ulimit
## Incident (2026-07-29)

While feta's daemon crash-looped (~every few seconds for ~1h, see the crash-loop issue), udder's daemon accumulated ~1000 open sockets — one per aborted peer dial — until it hit the stock `ulimit -n 1024`, at which point its accept loop spun hot on `Too many open files` (hundreds of ERROR lines per second) and the whole mesh degraded: kiwi showed both peers disconnected.

Two defects:
1. **Connections from peers that die mid-handshake/mid-session are not reaped.** Each of feta's aborted dials left a socket open on udder indefinitely. Connections should have handshake deadlines and be closed on error paths (worth auditing the #1229 wire-generation rejection path specifically for the same leak — a rejected client's socket must be closed after the refusal is written).
2. **accept() failure spins hot.** On `EMFILE` the accept loop retries instantly, burning CPU and flooding logs. It should back off, and ideally shed the oldest idle connections.

Mitigation applied operationally: daemon launchers now set `ulimit -n 8192` (kiwi/feta/udder) — worth making this a daemon-side rlimit raise or at least a documented requirement, since 1024 is the stock soft limit on most distros.

## Acceptance
- [ ] A peer that repeatedly dials and dies does not grow the victim daemon's fd count without bound (verify with a kill-loop test against a live daemon)
- [ ] Wire-generation-rejected clients leave no open socket behind
- [ ] accept() EMFILE path backs off instead of spinning; fleet diagnosis surfaces fd pressure before exhaustion

rjwittams:
Closes #1239

## What changed

- Bound accepted connections with 10-second preface and `Hello` handshake deadlines.
- Reap silent inbound peer sessions after 120 seconds; healthy peers remain active through the existing ping/pong traffic.
- Back off accept failures exponentially from 100 ms to 5 seconds, including `EMFILE`/`ENFILE`, instead of spinning.
- Raise the daemon's soft `RLIMIT_NOFILE` to 8192 where the hard limit permits.
- Publish a `FileDescriptors` degraded host condition once usage reaches 80%, so `fleet health` diagnoses pressure before exhaustion.
- Add focused lifecycle, generation-rejection, backoff, resource-pressure, and live-daemon fd-count regressions.

## Root cause

Accepted sockets could wait indefinitely before a complete peer handshake, and inbound peer sessions had no victim-side silence deadline. Repeated aborted SSH-forwarded dials therefore retained descriptors. Separately, the accept loop retried resource-exhaustion errors immediately.

## Validation

- Live-daemon churn regression holds 32 partial dials open and verifies all daemon-side socket descriptors are released at the handshake deadline while the client ends remain open.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1128,6 +1128,7 @@ dependencies = [
  "flotilla-transport",
  "futures",
  "indexmap",
+ "libc",
  "rusqlite",
  "serde",
  "serde_json",
--- a/crates/flotilla-daemon/Cargo.toml
+++ b/crates/flotilla-daemon/Cargo.toml
@@ -31,6 +31,9 @@ uuid = { version = "1", features = ["v4"] }
 futures = "0.3"
 visibility = "0.1.1"
 
+[target.'cfg(unix)'.dependencies]
+libc = "0.2"
+
 [dev-dependencies]
 flotilla-test-support = { path = "../flotilla-test-support" }
 indexmap = { workspace = true }
--- a/crates/flotilla-daemon/src/cli.rs
+++ b/crates/flotilla-daemon/src/cli.rs
@@ -9,7 +9,7 @@ use flotilla_core::{
 use tracing::info;
 use tracing_subscriber::{layer::SubscriberExt, util::SubscriberInitExt};
 
-use crate::{runtime::DaemonRuntime, server::DaemonServer};
+use crate::{resource_limits::raise_file_descriptor_limit, runtime::DaemonRuntime, server::DaemonServer};
 
 pub async fn run(socket_path: &Path, config_dir: &Path, state_dir: &Path, timeout_secs: u64) -> Result<(), String> {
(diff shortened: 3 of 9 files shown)


rjwittams:
Fixed in 0cee6fd6. Accept failures now arm a retry deadline in its own outer `tokio::select!` branch and disable only `listener.accept()` while the timer is active. The shutdown watch, SIGINT, and SIGTERM branches remain pollable throughout the full backoff window. The pinned format check, workspace clippy gate, and full locked workspace test suite all pass after the change.

rjwittams:
Reviewed. This follow-up confirms the blocking finding is fixed and reports no new correctness or security issues. The two remaining notes are explicitly non-blocking: the daemon is already Unix-specific, and the fd-pressure calculation/wiring has direct unit coverage. No further change is needed for this bug-fix PR.

PR #1242, opened 2026-07-29
feat: lease generic material pools into agent adapters

### LINKED ISSUE #1235: Credential seeding slice 1: codex pool slots leased into contained vessels
## Parent

#1140 (operator setup umbrella). Design ruled in ADR 0022 (as amended 2026-07-28: subscriptions primary, lease-located material) and `docs/superpowers/research/2026-07-28-multi-crew-agent-config-seeding.md`. Chain: this → containment smoke round 4 (#1215's CLI bind-mount is already merged).

## What to build

A contained (docker) vessel whose workflow requires the `codex` adapter gets a working, non-interactive codex login by **exclusively leasing one slot from the host's codex pool** for the vessel's lifetime:

- Pool location: `~/.config/flotilla/credentials/codex-pool/slot-N/` on the placement host (feta has slot-0, slot-1 minted and live). Each slot dir holds an `auth.json` (0600).
- **Lease semantics**: a slot is held by at most one vessel at a time (single-writer — codex refresh tokens are single-use; concurrent use of one login is a permanent `refresh_token_reused` failure). Lease acquired at vessel provisioning, released at teardown. Lease state must survive daemon restart without double-granting.
- **Delivery**: bind-mount the leased slot dir into the container and point `CODEX_HOME` at it, so codex's own token refresh writes back into the slot on the host — the slot stays current across leases.
- **Exhaustion is visible waiting, not failure**: when all slots are leased, the vessel waits in a legible state ("waiting for codex login slot; N in pool, all leased") rather than failing the convoy. Mint-on-exhaustion (interactive mint vessel) is a later slice — but the waiting state should already say that minting another slot is the remedy (it's a high-watermark-of-concurrency signal).
- **Do not overwrite existing state**: seeding only applies to vessels flotilla provisioned; never clobber a pre-existing `CODEX_HOME`/config in an already-set-up environment.
- Suppression seeds (skip-onboarding config so the agent gets to work instead of self-onboarding) per the research doc's codex section ride along in the same slot/mount.

## Acceptance criteria

- [ ] A contained vessel on a host with a free slot runs `codex exec`-style non-interactive work with zero interactive prompts (the operative test from the research doc)
- [ ] Two concurrent contained vessels get two different slots; a third waits legibly and proceeds when a slot frees
- [ ] Slot lease survives daemon restart: no double-grant, no orphaned-forever leases (teardown or restart recovery releases)
- [ ] Token refresh inside the container lands in the host slot dir (verify mtime/content change after a refresh)
- [ ] Never validated on macOS hosts (keychain leaks through isolation levers — Linux placement only for contained codex)

rjwittams:
## Summary

- add a host-local `MaterialPool` resource and CRD whose spec names opaque directory units and whose status durably records exclusive unit-to-holder leases
- preserve the existing single-writer, restart recovery, release, and exhaustion behavior in a generic material-pool manager with no Codex-specific naming or validation
- add an agent-material adapter seam, with Codex as the first specialization: it discovers and validates Codex login directories, leases one from its pool instance, mounts it read-write as `CODEX_HOME`, and runs the Codex login preflight
- keep explicit Codex API credentials and operator-provided `CODEX_HOME` authoritative; future CLI/auth/config consumers need only another pool instance plus adapter-side delivery logic, not another resource kind
- surface pool exhaustion through a typed `MaterialPoolExhausted` wait reason so the pending state is immediately visible without message-prefix coupling

## Resource shape

- `MaterialPool.spec.units`: named opaque host directory payloads
- `MaterialPool.status.leases`: exclusive unit leases keyed to typed holder resource references
- replication: host-local; payload bytes never enter the resource store
- persistence: lease status is stored in the durable resource backend and recovered against active holders after restart

## Validation

- `cargo test -p flotilla-core --locked --features test-support --test in_process_daemon` (106 passed)
- focused generic exclusivity, SQLite restart persistence, orphan recovery, recovery-scope, adapter specialization, writable mount, explicit override, exhaustion, cleanup, and visible-wait tests

Real Docker/Codex login refresh was intentionally not exercised on this macOS host; Codex material delivery remains Linux-gated and is covered with fake-provider contract tests.

Closes #1235


rjwittams:
Fixed in 5f7340e6. Codex slot admission now happens before registry login/pull, so an exhausted pool returns Waiting without creating or repeatedly replacing a registry credential cache. All fallible pre-container credential/lease paths now call the same environment cleanup, which releases any lease and removes any prior cache before returning.

Added `codex_slot_wait_skips_registry_preflight_and_leaves_no_credential_cache`, covering a private-registry credential combined with an exhausted one-slot Codex pool. It asserts zero registry-runner calls, no provider create, no on-disk credential cache, and the expected typed Waiting result. Formatting, strict workspace clippy, the full workspace suite, and the 106-test feature-gated daemon integration target all pass after the fix.

rjwittams:
Thanks for the re-review. I handled the three optional notes as follows:

- Fixed the unsupported-host state-file side effect in 140bbe70. `recover` and `release` now no-op when slot delivery is unsupported, with `unsupported_hosts_do_not_create_lease_state_during_recovery_or_release` covering both paths.
- Filed #1244 for typed provisioning wait reasons, since replacing the current message contract cleanly requires a resource/status shape change beyond this slot-delivery PR.
- Filed #1245 to centralize required agent-adapter resolution between workflow admission and per-vessel materialization.

After the additional fix, formatting, strict workspace clippy, the full workspace suite, and the 106-test feature-gated daemon integration target all pass.

rjwittams:
Thanks for the final pass and the additional crash-recovery/exclusivity verification. Confirmed: no further code changes are needed from this review, and the two non-blocking architecture notes remain tracked in #1244 and #1245.

rjwittams:
There is too much specificity in a "CodexSlot" here. 
This should be a general mechanism for making use of a pool of single-use-at-a-time 'authentication' or 'resources/config' with the codex agent adaptor making use of it and specialising it to this particular codex usage.

rjwittams:
Addressed in 482714cd. The new shape is:

- `MaterialPool` is the only resource kind. Its spec is a map of named opaque directory units; its status maps units to typed holder references. The generic resource and lease-manager layers contain no Codex naming, file conventions, or validation.
- The generic manager retains the existing semantics: exclusive acquisition with resource-version conflict handling, idempotent reacquisition by the same holder, durable lease status across SQLite/daemon restart, scoped orphan recovery, release, and an explicit exhausted outcome.
- `AgentMaterialAdapter` is the specialization seam. The first implementation is the Codex adapter, which alone owns the `codex-login` pool instance, discovers `slot-N` directories, validates protected non-empty `auth.json`, turns the leased directory into a read-write `CODEX_HOME` bind mount, and runs `codex login status`.
- Exhaustion is now carried as a typed `MaterialPoolExhausted` environment wait reason, so it is immediately visible without sniffing Codex message text.
- A future Claude setup-token mint, another CLI, or proxy/config payload adds another pool instance and adapter-side discovery/delivery implementation; it does not add a resource kind or alter the pool/lease machinery.

The exact requested format, clippy, and full workspace test gates pass, along with the 106-test feature-gated daemon integration target.

rjwittams:
Both findings from this pass are now fixed.

- `b6b3f2db` detects numeric slot aliases such as `slot-0` and `slot-00`, keeps the first discovered directory, and emits a structured warning naming both paths. `duplicate_numeric_codex_slots_keep_one_unit_and_warn_with_both_paths` covers the behavior.
- `b0b4dac8` removes the `deletion_timestamp` filter from startup material recovery. Every `Environment` still present in the resource store now counts as an active holder until its finalizer completes and removes it. `startup_recovery_retains_material_lease_for_environment_pending_finalization` creates an environment with a finalizer, marks it for deletion, seeds its lease, starts the daemon, and verifies a second holder still waits. The test reproduced the bug before the fix by granting the second holder the same unit.

The exact pinned format, strict workspace clippy, and full workspace test gates pass.

rjwittams:
Fixed in `b0b4dac8`. Startup recovery no longer filters out environments with `deletion_timestamp`: presence in the store is the active-holder criterion, so a mid-finalization container keeps its material lease. Fully orphaned holders remain releasable because their environment objects are absent from `list()`.

The new `startup_recovery_retains_material_lease_for_environment_pending_finalization` runtime test exercises the exact boundary. It creates an `Environment` with a finalizer, marks it for deletion, confirms the object remains with `deletion_timestamp`, seeds its `codex-login` lease, starts the daemon with controllers disabled, and asserts a second holder receives `Waiting`. Before removing the filter, that test failed because the second holder was granted the same unit.

I re-read and checked every finding across all five review passes:

- Credential-cache leakage and repeated registry preflight on material exhaustion/error: fixed by preflight ordering plus unified `forget_environment_state` cleanup, with `material_pool_wait_skips_registry_preflight_and_leaves_no_credential_cache` coverage.
- Unsupported-host lease-state side effect: fixed before the generic re-layering; the current Codex adapter rejects unsupported delivery before pool reconciliation, and generic leases live in the resource backend rather than a Codex state file.
- Fragile wait-message prefix coupling: fixed in this PR by typed `EnvironmentWaitReason::MaterialPoolExhausted`. #1244 exists and is still open, but the specific review finding it records is now implemented by this PR.
- Duplicated required-agent-adapter resolution: intentionally out of scope and verified still tracked in open issue #1245.
- Codex-specific resource design: fixed by the generic `MaterialPool`/lease layer plus adapter-side Codex specialization.
- Numeric slot-name collision: fixed in `b6b3f2db` with one-unit-plus-warning regression coverage.
- Mid-deletion startup recovery race: fixed in `b0b4dac8` with the failing-then-passing runtime regression described above.

No finding from the five passes is now unaddressed: correctness findings are fixed, the resolved typed-wait note remains recorded in #1244, and the one intentionally deferred design cleanup remains explicitly tracked in #1245. The exact pinned format, strict workspace clippy, and full workspace test gates pass on this head.

rjwittams:
Thanks for the full-history verification. Confirmed: the sixth pass finds no remaining correctness issue, and #1245 is the sole intentionally deferred design refactor from the prior rounds.

I am leaving the suggested `ENVIRONMENT_WAITING_REQUEUE_AFTER` extraction unchanged in this PR. It is explicitly cosmetic and non-blocking, while this final correction was intentionally scoped to restoring the material-lease safety invariant and its regression coverage; changing the existing one-use requeue literal would not affect that behavior.

The exact format, clippy, and workspace test gates, plus all GitHub checks, pass on `b0b4dac8`.

PR #1243, opened 2026-07-29
fix: quarantine decode failures and surface crash loops

### LINKED ISSUE #1238: Daemon exits on one undecodable stored object; crash-loop is silent when connect_or_spawn resurrects it
## Incident (2026-07-29, ~90 min fleet outage)

After the fleet bump to `24e09f29`, feta's daemon crash-looped at boot: `materialize whole-repository Project ... decode stored object: missing field 'source_ref'` — a pre-#1230 Convoy record in its store no longer decodes under the new schema (expected in the no-compat phase; the store should have been reset with the bump). Two systemic defects made this an outage instead of a nuisance:

1. **One stale object kills the whole daemon.** `materialize_tracked_repo_projects` `?`-propagates the decode error all the way out; the daemon exits. The per-item isolation lesson from the controller-budget incidents (#1206/#1208/#1210) applies here too: an undecodable stored object should be quarantined loudly (WARN + fleet-visible diagnosis naming kind/name and the decode error) while the daemon keeps serving everything else.

2. **The crash-loop is silent on hosts with local clients.** Kiwi hit the same class but every crew CLI call `connect_or_spawn`-resurrected the daemon, so it looked healthy while actually restart-cycling — the only symptom was `ProjectionParity` degradation (aggregator missing the two newest convoys) and mysteriously short daemon uptimes. A daemon that exited abnormally N times in a window should be loudly visible in `flotilla fleet` (e.g. "restarted 14× in 30m").

Secondary finding from the same incident: diagnosing this took hours partly because the operator (me) chased phantom "resurrection" paths — the legible quarantine in (1) would have named the poisoned object in one glance.

## Acceptance
- [ ] A stored object that fails typed decode is quarantined (skipped + WARN with kind/name/error + fleet diagnosis marker), daemon stays up and serves
- [ ] Abnormal-exit frequency is surfaced in fleet output (spawn-source-agnostic: setsid launch and connect_or_spawn alike)
- [ ] Regression test: store containing one undecodable object of each replicated kind → daemon boots, serves, reports

rjwittams:
## Summary

- quarantine undecodable SQLite resource rows instead of failing the whole typed list or daemon startup
- expose durable decode-quarantine details through fleet health and structured warning logs
- track abnormal `flotillad` exits across every launcher and diagnose restart frequency over a rolling 30-minute window
- cover all currently replicated resource kinds, lifecycle locking, clean/unclean exits, and restart diagnostics

## Root cause

SQLite typed listing decoded the full result set with `?`, so one malformed stored object aborted startup recovery. Separately, daemon launch paths had no shared durable lifecycle marker, so repeated abnormal exits were invisible to operators.

## Validation

Closes #1238

### DIFF
--- a/crates/flotilla-daemon/Cargo.toml
+++ b/crates/flotilla-daemon/Cargo.toml
@@ -30,6 +30,7 @@ tracing-subscriber = { version = "0.3", features = ["env-filter", "json"] }
 uuid = { version = "1", features = ["v4"] }
 futures = "0.3"
 visibility = "0.1.1"
+libc = "0.2"
 
 [target.'cfg(unix)'.dependencies]
 libc = "0.2"
--- a/crates/flotilla-daemon/Cargo.toml
+++ b/crates/flotilla-daemon/Cargo.toml
@@ -30,7 +30,6 @@ tracing-subscriber = { version = "0.3", features = ["env-filter", "json"] }
 uuid = { version = "1", features = ["v4"] }
 futures = "0.3"
 visibility = "0.1.1"
-libc = "0.2"
 
 [target.'cfg(unix)'.dependencies]
 libc = "0.2"
--- a/crates/flotilla-daemon/src/cli.rs
+++ b/crates/flotilla-daemon/src/cli.rs
@@ -9,9 +9,7 @@ use flotilla_core::{
 use tracing::info;
 use tracing_subscriber::{layer::SubscriberExt, util::SubscriberInitExt};
 
-use crate::{
-    resource_limits::raise_file_descriptor_limit, restart_history::DaemonLifecycle, runtime::DaemonRuntime, server::DaemonServer,
-};
+use crate::{resource_limits::raise_file_descriptor_limit, restart_history::DaemonLifecycle, runtime::DaemonRuntime, server::DaemonServer};
 
 pub async fn run(socket_path: &Path, config_dir: &Path, state_dir: &Path, timeout_secs: u64) -> Result<(), String> {
     let lifecycle = DaemonLifecycle::begin(state_dir)?;
--- a/crates/flotilla-daemon/src/cli.rs
+++ b/crates/flotilla-daemon/src/cli.rs
@@ -9,9 +9,12 @@ use flotilla_core::{
 use tracing::info;
 use tracing_subscriber::{layer::SubscriberExt, util::SubscriberInitExt};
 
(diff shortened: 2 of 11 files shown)


rjwittams:
Addressed in 6333cee7:

1. A successful same-identity create now deletes the active quarantine record in the same SQLite transaction; the shared upsert path does the same for valid updates/status writes. Added `recreating_a_quarantined_object_clears_its_decode_diagnosis`, which was red before the fix and now passes.
2. Moved `libc` under `[target.'cfg(unix)'.dependencies]`. The daemon itself is currently Unix-only because its transport directly uses Tokio Unix sockets, so this change follows the workspace dependency convention rather than claiming Windows support.
3. `RuntimeHealth::conditions` now performs the restart-history read through `spawn_blocking`, including a fleet-visible failure condition if the blocking task itself fails.

Revalidated with the pinned formatter, exact workspace clippy gate, and the full sandbox-safe workspace test command.

rjwittams:
Thanks — agreed that this is a deliberate scope boundary for #1238. The supported pre-existing-corruption path is swept across every registered SQLite kind before startup recovery, and typed lists quarantine any row encountered later. A point read targets one identity, so returning its decode error does not recreate the incident's whole-daemon/whole-list failure mode. Post-boot direct database mutation or a live schema downgrade is not a supported write path; replication validates typed objects before they reach this local table.

Also confirmed the abnormal-restart condition is intentionally loud for every abnormal exit in the rolling window: the acceptance asks for frequency visibility, and a clean stop never increments it. No further code change for these non-blocking observations.

rjwittams:
Confirmed this is intentional and kept in this PR because the exact post-rebase workspace gate exposed the flake. The removed reads count process-global `/dev/fd` entries, so unrelated tests running concurrently can move both the baseline and final count; the test passed in isolation and the same test binary passed serially.

The retained check is scoped to every one of the 32 connections: after advancing `HELLO_HANDSHAKE_TIMEOUT`, each client must read EOF. In the production path, `handle_client` exclusively owns the accepted `UnixStream`; a handshake timeout returns from `handle_client_session`, then `handle_client` returns and drops the session/stream. There is no half-close or retained stream owner on that path, so per-connection EOF directly demonstrates that each server endpoint was closed without inferring it from unrelated process-global descriptors.

After this change, the daemon lib suite passed with 16 test threads, all three exact local gates passed, and all four refreshed GitHub checks are green. No further code change here.

PR #1248, opened 2026-07-29
fix: preserve remote placement policy fields

### LINKED ISSUE #1236: Daemon policy re-registration stomps operator-authored spec fields (priority reverts to default)

rjwittams:
## Summary

- route local startup/default-policy registration and remote/announced host policy registration through one ownership-aware helper
- preserve operator-authored policy fields such as `priority` and future scheduling hints while still correcting daemon-owned `pool`, host, checkout, and strategy fields
- refuse to adopt or rewrite policies carrying `flotilla.work/managed-by`

## Second writer diagnosis

The second writer was `InProcessDaemon::materialize_peer_host_direct_placement`, invoked when a remote `HostSummary` is published and again by connected-peer heartbeat refresh. Its update path replaced the entire `PlacementPolicy` spec with the byte-canonical registration shape, removing `priority`. PlacementPolicy is not replica-enabled, so replica merge-back was not involved.

The regression test now drives the missed production sequence directly: remote announcement registers the policy, an operator applies `priority` plus owned-field drift, then a subsequent remote announcement preserves `priority` while restoring the announced pool/host/checkout. A companion test covers refusal of a manifest-managed collision.

## Verification

Closes #1236


rjwittams:
Verified. The conflict observation is pre-existing and both callers retry naturally on the next registration/heartbeat cycle, so no change is needed in this focused fix. The remote-registration regression and shared ownership helper directly cover the reported stomp.

PR #1249, opened 2026-07-29
fix: reconcile remote vessel placements locally

### LINKED ISSUE #1247: Remote placement broken: vessel reconciler resolves the placement host's Environment from the admitting store
## Two specimens (2026-07-29), all remote placements affected

`capability-shed-loud` r1 and `docker-crew-smoke-r4` both placed on feta (placement decisions correct — host-direct and docker-per-vessel respectively) and both immediately failed actuation with:

```
host-direct environment host-direct-d783f8a7-… not found
```

Store forensics: feta's primary store **has** `Environment/host-direct-d783f8a7-…`; kiwi has only its own environment and **no replica of feta's** — Environment is a non-replicated kind. Neither daemon logged the failure (it lands only in convoy status). Conclusion: the vessel reconciler for a remotely-placed convoy runs against the **admitting** store and cannot see a placement-host-local resource.

## Introduced by #1222

`reconcilers/vessel.rs` (the `needs_shared_clone()` path added with single-flight clone provisioning) resolves `host_direct_environment_name(strategy.host_ref())` via `self.environments.get(...)` before provisioning. Docker placements with `worktree_on_host_and_mount` and remote host-direct placements both take this path. Timeline fits: feta placements worked through yesterday afternoon (clone-single-flight) and broke exactly at the `07cfc9d6` bump that shipped #1222.

## Direction — follow the #1188 contract, don't widen replication ad hoc

The #1188 federation contract already rules this shape: **placement-host reconcilers watch replicated-in resources (placed-on-me filter) and actuate locally**. An admitting-side reconciler reaching for the placement host's Environment contradicts that ruling. Preferred fix: the shared-clone/vessel provisioning step that needs the placement host's Environment must run in the placement host's reconciler (which sees its own store), not on the admitting host. Replicating the Environment kind fleet-wide is the fallback only if the reconciler split is genuinely out of reach in one slice — if so, say why in the PR.

Also fix the legibility gap that made this expensive to diagnose twice: the failure message should name the store it looked in and state the kind is host-local (and it should be WARN-logged, not status-only).

## Acceptance
- [ ] A convoy dispatched from kiwi and placed on feta (host-direct AND docker) provisions end-to-end
- [ ] Regression test covering remote placement with the shared-clone strategy (two-store InProcessDaemon-level, no real multi-host needed)
- [ ] Actuation failures of this class are WARN-logged with store/host context

rjwittams:
## Summary

- project replicated-in Vessels whose placement decision targets the local host into locally authored actuator records, so the placement host performs shared-clone and vessel provisioning against its own resource store
- carry actuator provenance through Checkout and TerminalSession resources, and fold the placement host's replicated Vessel/Checkout facts back into the admitting Convoy's status
- replicate the control-plane PlacementPolicy dependency and Checkout fact needed by that flow while keeping Environment strictly host-local
- WARN when a host-local Environment is absent, with the Environment reference, placement host, and reconciler-store host in both logs and status

## Architecture

This follows the #1188 / ADR 0016 direction: replicated-in placement intent is filtered by `placed-on-me`, then actuation is authored and reconciled on the placement host. The admitting-side Vessel reconciler stands down for remote decisions. Environment replication is deliberately unchanged; the actuator reads only the placement host's local Environment.

The projected actuator record is an incremental bridge toward placement-authored Vessels: it preserves the admitting record as coordination input while giving the placement daemon a local-only primary that its existing controller can reconcile. Origin annotations make dependency lookup and returned facts source-qualified.

## Validation

- two-store InProcessDaemon regressions for remote host-direct and Docker worktree-on-host-and-mount placement
- focused projection idempotence and structured WARN-context coverage

Closes #1247

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1071,6 +1071,7 @@ dependencies = [
  "tempfile",
  "tokio",
  "tracing",
+ "tracing-subscriber",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1065,6 +1065,7 @@ dependencies = [
  "flotilla-manifest",
  "flotilla-protocol",
  "flotilla-resources",
+ "futures",
  "rstest",
  "serde_json",
  "sha2",
--- a/crates/flotilla-controllers/Cargo.toml
+++ b/crates/flotilla-controllers/Cargo.toml
@@ -23,3 +23,4 @@ flotilla-resources = { path = "../flotilla-resources", features = ["test-support
 tokio = { workspace = true, features = ["macros", "rt-multi-thread"] }
 rstest = { workspace = true }
 tempfile = "3"
+tracing-subscriber = { version = "0.3", features = ["fmt"] }
--- a/crates/flotilla-controllers/Cargo.toml
+++ b/crates/flotilla-controllers/Cargo.toml
@@ -15,6 +15,7 @@ bon = { workspace = true }
 tracing = { workspace = true }
 chrono = { version = "0.4", default-features = false, features = ["clock", "serde"] }
 dirs = "6"
+futures = "0.3"
 sha2 = "0.10"
 serde_json = { workspace = true }
 
--- a/crates/flotilla-controllers/src/reconcilers/mod.rs
+++ b/crates/flotilla-controllers/src/reconcilers/mod.rs
@@ -5,6 +5,7 @@ pub mod presentation;
 pub mod repository;
 pub mod terminal_session;
 pub mod vessel;
+pub mod vessel_placement;
 
 pub use checkout::{
     BranchPreservationReason, CheckoutReconciler, CheckoutRemoval, CheckoutRemovalOutcome, CheckoutRuntime, PreparedCheckout,
@@ -19,3 +20,4 @@ pub use presentation::{
 pub use repository::{ForgeDefaultBranchResolver, RepositoryReconciler};
 pub use terminal_session::{TerminalRuntime, TerminalRuntimeState, TerminalSessionReconciler};
 pub use vessel::{VesselDeps, VesselReconciler};
+pub use vessel_placement::{VesselPlacementProjector, VesselPlacementSync};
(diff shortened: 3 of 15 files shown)


rjwittams:
Thanks for tracing the invariants so carefully. I addressed the idle scan concern in 9038492b: the projector now establishes live Convoy/Vessel replica watches, performs one initial sync, and resyncs only for replica events. Local actuator events are ignored, so a single-host daemon stays idle instead of listing three kinds four times per second.

The source-qualified dependency lists are intentionally the existing typed read-only overlay API for this slice; an indexed `(origin_root, name)` resolver would be a useful store-level follow-up if fleet cardinality makes these scans material.

On collision ordering, `ReplicaReadResolver::list_sources` sorts equal names by provenance and replica `origin_root` before returning them, so after local rows are filtered the selected admitting origin is stable rather than arrival-ordered. The warning/degraded behavior remains deliberate because the colliding name is anomalous either way.

The focused projector test, both two-store regressions, formatter, clippy, and the full sandbox-safe workspace suite all pass after the watch change.

rjwittams:
Good catch. Fixed in 285e4a67 by establishing the federated secondary watch before taking and enqueueing the initial list. An event in that startup interval is now queued and may cause one harmless duplicate enqueue after the list, rather than being deferred to periodic resync.

The pinned formatter, workspace clippy with warnings denied, and full sandbox-safe workspace suite all pass after this change.

rjwittams:
Thanks for the independent re-check. I agree the two origin-scoped lookups are mechanically duplicated. I’m leaving that non-blocking DRY change out of this PR because the right shared seam is the read layer: an origin-qualified `get` on `ReplicaReadResolver`, which would also remove the O(n) list scans noted in the first review. Extracting the current list-and-scan body into a controllers utility would deduplicate syntax while entrenching the weaker API.

The substantive placement, idle-scan, and startup-race findings are all resolved on the current head. I’ll keep the source-qualified lookup improvement as the store-level follow-up if fleet cardinality makes it material.

PR #1250, opened 2026-07-29
docs: ADR 0024 — declared state machines, static field ownership, fold-based property engine

rjwittams:
Records the rulings from the 2026-07-29 grill on the week's state-transition bug classes, informed by the prior-art research (included): machines as Rust data, static field ownership with no force parameter enforced at a single write path, observe-then-enforce per-machine rollout, one generic fold/harness engine with per-resource property data, enrollment convoy-first. Class 3 (store authority) scoped to #1188 with the write-helper left extensible toward it.

Implementation chain filed separately (#TBD).


rjwittams:
Header fixed to the 0016+ convention (bold Status/Date order + Relates-to linking ADR 0003, #1188, and the #1176/#1214 harness). Thanks for the citation-level verification.

PR #1261, opened 2026-07-29
fix: replicate resources over inbound peer connections

### LINKED ISSUE #1257: Follower hosts cannot replicate resources in: replication waits for an outbound connection followers never make
## Observed (2026-07-29, blocks all remote placement)

After feta's store reset and the `9f3b3652` bump, feta's replica store stayed **completely empty** (`replica_objects` 0, `replica_cursors` 0) across multiple clean daemon boots with stable peers. Feta's structured log explains it precisely:

```
peer has no forwarded resource socket; replication waits for an outbound SSH connection  (peer 0fe74f4d… = kiwi)
```

Feta is `follower = true` with no peer list; kiwi (leader, `hosts.toml`) dials followers over SSH. Resource replication apparently requires the *puller* to own an outbound connection — which a follower, by definition, never makes. Result: Convoy/Vessel resources admitted on kiwi never reach feta, #1249's placed-vessel reconciler runs there syncing zero actuators, and every feta-placed convoy sits `Active`/crew-`Pending` forever (three currently stalled: docker-crew-smoke-r4, field-ownership-write-path, harness-transition-vocab).

Replication INTO feta demonstrably worked earlier today (~01:10, a fresh store pulled 10MB) — so either the leader's inbound connection previously carried a forwarded resource socket in both directions and a recent connection-lifecycle change dropped it (#1229 wire handshake / #1241 connection resource bounds / #1249 are the candidates — bisect), or generation churn from today's bounces wedged it. Second finding either way: `flotilla fleet` reported feta's replica sync as `current`/`agree` throughout — the diagnosis measures generation agreement, not replica content, so a completely empty replica store reads as healthy.

## Direction

Immediate fix: restore resource-socket forwarding over the leader-initiated connection (bidirectional), so followers replicate without dialing. Structural note for the ruling record: Robert has long intended to retire the follower setting ("we'll have to work out when we can get rid of the follower setting") — if the clean fix is symmetrical peering (followers get peer addresses and dial too), say so explicitly in the PR and this becomes the retirement's first slice rather than a workaround.

## Acceptance
- [ ] A follower with a fresh store bootstraps full replication from the leader without any outbound-dial config
- [ ] The three stalled feta convoys provision once deployed (or after re-dispatch)
- [ ] Fleet diagnosis reflects replica content: a host with zero cursors while peers hold replicable resources reads DEGRADED, not agree
- [ ] Regression: two-store harness case — leader admits, follower store starts empty, placed-on-me actuation proceeds

rjwittams:
Closes #1257

## Summary

- forward both daemon resource sockets over the existing leader-initiated SSH session
- mark connected peers with zero replica cursors as DEGRADED until replication bootstraps
- strengthen the two-store remote-placement regression so the follower begins with empty Convoy and Vessel replica cursors

## Cause

Resource replication only gave the outbound SSH owner a forwarded HTTP-over-UDS path. An inbound follower connection emitted no resource socket, so its replicators waited forever. The test-support transport routed watches in memory, which masked that production-only gap.

## Connection topology

This implements the #1257 ruling: the same leader-owned SSH process carries a local and reverse Unix-socket forward. Followers do not dial, gain a peer address book, or create a second SSH connection. Both sides derive the reverse socket path from the accepting daemon socket and dialing node identity during the existing handshake.

## Verification

The live feta convoy recovery remains a post-deployment acceptance check; this PR does not merge or deploy itself.

### DIFF
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -2374,6 +2374,14 @@ impl InProcessDaemon {
         self.host_registry.peer_connection_status(node_id).await
     }
 
+    pub async fn connected_peer_node_ids(&self) -> Vec<NodeId> {
+        let mut peers =
+            self.host_registry.connected_peer_summaries().await.into_iter().map(|summary| summary.node.node_id).collect::<Vec<_>>();
+        peers.sort();
+        peers.dedup();
+        peers
+    }
+
     pub async fn set_configured_peers(&self, peers: Vec<NodeInfo>) {
         let remote_counts = self.remote_host_counts().await;
         self.host_registry
--- a/crates/flotilla-daemon/src/peer/mod.rs
+++ b/crates/flotilla-daemon/src/peer/mod.rs
@@ -13,8 +13,8 @@ pub use manager::{
     ReversePathKey, RouteHop, RouteState,
 };
 pub use merge::merge_provider_data;
-pub(crate) use ssh_transport::peer_resource_socket_path;
-pub use ssh_transport::SshTransport;
+pub(crate) use ssh_transport::{peer_resource_socket_path, reverse_peer_resource_socket_path};
+pub use ssh_transport::{SshTransport, SshTransportPaths};
 pub use transport::{PeerConnectionStatus, PeerSender, PeerTransport};
 
 #[cfg(test)]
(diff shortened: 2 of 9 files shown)


rjwittams:
Thanks for tracing this against the topology ruling. Agreed both observations are non-blocking here: the cursor fan-out can be collapsed behind a backend query if profiling warrants it, and `spawn_embedded_peer_networking` does not consume peer-connected notices or run inbound-triggered replicators today. I’m keeping this PR scoped to the live `DaemonServer` path and #1257’s acceptance items.

PR #1279, opened 2026-07-30
feat: enforce declared resource field ownership

### LINKED ISSUE #1251: Field ownership: declared owner roles enforced at a single write path (ADR 0024 slice 1)
## What to build

Per ADR 0024 (§2, §3): every spec/status field of lifecycle-bearing resource kinds is statically assigned one owner role — operator-authored, loop-derived, or actuator-observed — in a per-resource table in `flotilla-resources`, next to the type. All writers route through one write helper that takes the writer's role and, for fields the role does not own, preserves the stored value. **No force parameter exists.** Violations are recorded as visible events (writer role, field, attempted value, rule) and surfaced in fleet diagnosis.

Rollout infra: per-machine observe/enforce mode. In observe mode violations are logged + fleet-visible but writes proceed; in enforce mode the write is refused with a typed error callers treat as a stale-view requeue.

First enrollee: **PlacementPolicy** — refactor the #1236 unified preserve-helper (in flight) into this declared form; `spec.priority` is the canonical operator-authored field, `host_direct`/`pool` loop-derived. Acceptance includes the #1236 bait scenario as a property: apply priority as operator, run local AND remote registration cycles, priority survives.

The helper's writer-identity parameter must be shaped so store-authority identity (#1188 single-writer) can extend it later without signature churn.

## Acceptance criteria

- [ ] Ownership tables exist for PlacementPolicy (full) with the table format documented for later enrollees
- [ ] Single write helper; direct `update()` on enrolled kinds from reconcilers is gone
- [ ] Observe mode logs + surfaces violations; enforce mode refuses with a typed error; per-machine flip
- [ ] Property test: operator-applied priority survives local and remote registration cycles indefinitely
- [ ] No force/override path exists anywhere in the API

## Blocked by

None — coordinate with the in-flight #1236 crew's helper (refactor, don't duplicate).

rjwittams:
Closes #1251

## Summary

- add reusable static field-ownership declarations and a single role-aware spec write path
- enroll PlacementPolicy in observe mode, preserving operator priority across local and remote registration
- persist/log ownership violations and surface them through fleet health diagnosis
- provide enforce-mode typed stale-view errors without a force/override API

## Verification

### DIFF
--- a/crates/flotilla-core/src/placement_policy.rs
+++ b/crates/flotilla-core/src/placement_policy.rs
@@ -1,4 +1,6 @@
-use flotilla_resources::{InputMeta, PlacementPolicy, PlacementPolicySpec, ResourceBackend, ResourceError, MANAGED_BY_LABEL};
+use flotilla_resources::{
+    InputMeta, PlacementPolicy, PlacementPolicySpec, ResourceBackend, ResourceError, WriterIdentity, MANAGED_BY_LABEL,
+};
 use tracing::warn;
 
 /// Reconcile a daemon-discovered placement policy without claiming
@@ -20,10 +22,14 @@ pub async fn reconcile_registered_policy(
                 return Ok(());
             }
 
-            let merged = merge_registered_policy_spec(&existing.spec, desired)?;
-            if merged != existing.spec {
+            if desired != &existing.spec {
                 policies
-                    .update(&InputMeta::from(&existing.metadata), &existing.metadata.resource_version, &merged)
+                    .write_spec(
+                        &WriterIdentity::reconcile_loop(),
+                        &InputMeta::from(&existing.metadata),
+                        &existing.metadata.resource_version,
+                        desired,
+                    )
                     .await
                     .map_err(|error| format!("reconcile registered placement policy {name}: {error}"))?;
             }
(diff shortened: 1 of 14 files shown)


rjwittams:
Fixed in ea8a82a9. Strategy presence is now a loop-owned structural field: legitimate loop switches skip descendant leaf checks, while an operator clear preserves the complete stored strategy and records a visible ownership violation instead of returning `Invalid`. Regression tests cover both directions. I also bounded retained violations to 1,024 in both stores, narrowed the status-path documentation, documented the HTTP diagnostic constraint, and replaced the one-off apply branch with reusable enrolled-kind dispatch.

rjwittams:
Fixed in e1cc881f. Manifest reconciliation now hashes the object returned by the ownership-aware apply path and, when filtering changed the persisted spec, corrects the last-applied annotation to that persisted hash. The regression test now performs a later manifest pass and proves it is not classified as drift and can still apply an operator-owned priority update while preserving the loop-owned pool.

rjwittams:
Fixed in 6a29785f. Registration now starts from the stored spec and overlays only loop-owned topology fields before calling the ownership-aware write path, so default values no longer manufacture intent for operator fields. The former synthetic direct-write stress test has been replaced with production registration tests: 256 alternating host-direct refreshes preserve operator priority with zero diagnostics, and a Docker refresh preserves all customized operator runtime fields with zero diagnostics.

rjwittams:
Fixed in 624725b8. Manifest reconciliation now projects an enrolled resource's desired spec through Operator and ReconcileLoop in separate ownership-aware writes: operator fields remain operator-authored, topology fields are applied by the manifest reconcile loop, and the general dynamic apply API remains operator-only. The regression changes both pool and priority, verifies external metadata and an empty violation store, then runs another pass and proves it is unchanged with no resource-version bump.

rjwittams:
Thanks — treating the suggested manifest strategy-bootstrap regression as non-blocking. The structural bootstrap behavior is intentional and is already exercised at the ownership merge layer in both directions; this capped shepherd cycle is otherwise clean. The unused `is_stale_view` note remains a pre-Enforce follow-up rather than a blocker for this Observe-mode slice.

PR #1280, opened 2026-07-30
test: add liveness harness transition vocabulary

### LINKED ISSUE #1252: Liveness harness transition vocabulary + the week's bugs as failing sequences (ADR 0024 slice 2)
## What to build

Per ADR 0024 (§4): extend the #1214 liveness harness from a straight-line reconcile loop to a driver over a transition vocabulary: `Reconcile`, `ExternalSpecWrite(field, value)`, `Delete`, `RestartController`, `AdvanceClock(δ)`, `DropActuation`, `DeliverActuation`, `PartitionStore(origin_root)`. Hand-written sequences for this slice (generation comes later).

Then encode this week's three dogfooding bugs as explicit sequences that **must fail against the pre-fix code shape and pass against current main**:
1. The #1236 stomp: operator `ExternalSpecWrite(priority)` interleaved with registration reconciles (local and remote writer roles).
2. The PR #1242 review's lease bug: holder gets `Delete` (deletion_timestamp set, finalizer not yet run), then `RestartController`, then a second holder's acquisition attempt — must NOT double-grant.
3. The #1202 ghost: `Delete` of a TerminalSession record mid-teardown, `RestartController`, recovery must not resurrect it.

Include sometimes-style coverage assertions so a sequence that never reaches the interesting intermediate state fails as vacuous.

## Acceptance criteria

- [ ] Vocabulary lands as harness API usable by any enrolled reconciler, alongside the existing bounded-convergence/quiescence checks
- [ ] The three sequences exist, are documented as regression properties, and pass on main
- [ ] Removing any of the three shipped fixes (locally, as an experiment documented in the PR) makes its sequence fail
- [ ] Coverage assertions fire on vacuous fixtures

## Blocked by

None - can start immediately.

rjwittams:
## What changed

- extend the reusable reconciler liveness harness with the ADR 0024 transition vocabulary: `Reconcile`, `ExternalSpecWrite`, `Delete`, `RestartController`, `AdvanceClock`, `DropActuation`, `DeliverActuation`, and `PartitionStore`
- add a generic transition driver that controls pending actuation delivery and composes with the existing bounded-convergence and quiescence enrollment
- add `sometimes` coverage assertions, including a harness self-test proving vacuous traces fail
- encode the three dogfooding regressions as hand-written transition properties:
  - local and remote placement-policy registration preserve an interleaved operator priority write (#1236/#1248)
  - daemon restart retains a material lease whose holder is pending finalization (PR #1242 review)
  - stale TerminalSession recovery after deletion/restart cannot resurrect an ownerless external session (#1202)

## Why

The straight-line liveness battery could verify convergence and quiescence, but it could not reach the interleavings that caused this week's field-stomp and restart/recovery bugs. This adds the disturbance vocabulary from ADR 0024 while keeping resource-specific operations behind fixture adapters.

## Re-break experiments

Each property was run against a temporary local removal of its shipped fix, then the production code was restored:

- changing placement registration to start from `desired` instead of the live spec failed with `coverage assertion \`operator priority survived a registration reconcile\` was never observed`
- filtering deletion-timestamped environments out of startup lease recovery failed with `coverage assertion \`second holder was refused after restart\` was never observed`
- removing the TerminalSession owner lookup caused the ghost sequence to call `ensure_session` and then fail its write with `resource not found: terminal-deleted-convoy-work-coder`

## Validation

The first full workspace run hit the existing transient fake-zellij `Text file busy (os error 26)` race; the isolated test passed immediately and the complete workspace rerun passed.

Closes #1252

### DIFF
--- a/crates/flotilla-controllers/tests/terminal_session_reconciler.rs
+++ b/crates/flotilla-controllers/tests/terminal_session_reconciler.rs
@@ -1,6 +1,9 @@
 use std::{
     collections::BTreeMap,
-    sync::{Arc, Mutex},
+    sync::{
+        atomic::{AtomicUsize, Ordering},
+        Arc, Mutex,
+    },
 };
 
 use async_trait::async_trait;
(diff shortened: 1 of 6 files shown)


rjwittams:
Verified; no code changes needed. These fixtures are sequence-driven, so their inert fixpoint predicates deliberately keep the existing enrollment triple without making a convergence claim. The clock and unsupported transition adapters similarly preserve the shared harness interface while each regression implements only its relevant disturbances. The atomic counter ordering is harmless test-only detail, as noted.

rjwittams:
Addressed in f34be55d. Reconcile and DeliverActuation now contain their inner error propagation so every failure reaches the shared transition index/variant wrapper. Added a regression test covering both paths, then reran format, clippy, and the full locked workspace suite.

rjwittams:
Reviewed; the re-review confirms f34be55d resolves the transition-context issue and raises no further changes. No additional action needed.

PR #1281, opened 2026-07-30
Run contained crews through interior cleat sessions

### LINKED ISSUE #1278: Contained crews launch their agent on the host, not in the environment (round-4 wall)
## Observed (2026-07-30, smoke r4 — abandoned as a false run)

The contained smoke convoy reached crew phase `Working` with everything *around* the agent correct — container up on feta, MaterialPool `codex-login` slot-0 leased to the environment, CLI bind-mount present — but:

- `docker exec … ps aux` inside the container: **only `sleep infinity`** (the keepalive). No agent, ever.
- On the **feta host**: a codex process with the crew-brief cmdline, `cwd=$HOME` (not any worktree), **respawn-looping** (fresh pid each ~minute — launched, dies, relaunched; crew phase stays `Working` throughout — a phantom-phase specimen for ADR 0024 class 2).
- The leased CODEX_HOME is mounted into the container and used by nothing; the host-side launches would see the host's own `~/.codex`.
- `flotilla attach --host feta --transient docker-crew-smoke-r4` from kiwi fails with `terminal pool cleat unavailable for environment env-docker-crew-smoke-r4-work` — attach resolves the pool *against the environment interior* (where no cleat exists; the crew image deliberately has none), while launch went host-side. Two paths, two different wrong answers.
- Related dormant evidence from round 3: `terminal_session` controller crash-looped with `provider registry unavailable for environment env-docker-crew-smoke-work` until budget exhaustion — the terminal-hosting-for-docker-environments gap was surfaced then and never closed.

## Direction

The agent process must run **inside** the environment; the terminal session hosting it should live **host-side**. Recommended shape: the crew engagement for a contained vessel creates a host cleat session whose command is `docker exec -it <container> <agent …>` with in-container cwd `/workspace` and env (`CODEX_HOME=<mounted slot path>`) — process containment preserved, recording/attach/watch stay on the host pool exactly as `pool: cleat` on the placement policy implies, and no cleat needs baking into the image. Attach then resolves the host session like any other crew. (If the engagement machinery wants a different cut, fine — but launch and attach must resolve through the *same* model, and the phantom-Working respawn loop must become a loud failure, not a stable lie.)

## Acceptance
- [ ] Contained crew's agent process is visible in `docker exec … ps` with cwd `/workspace` and `CODEX_HOME` pointing at the leased slot's in-container mount
- [ ] `flotilla attach` from another host reaches the crew session (controller or watch per cleat#178 semantics)
- [ ] A crew process that dies does not leave phase `Working`: either it restarts with the engagement recorded, or the convoy fails loudly
- [ ] Smoke r4 contract (#1246) passes end-to-end after this, probe report included

rjwittams:
## What changed

- make Docker-contained vessels request the `cleat` terminal pool inside the environment and remove silent fallback to a different pool for launch or presentation
- deliver the host cleat executable and its `libghostty-vt` runtime dependency into Linux containers as read-only mounts
- mount cleat's runtime root at `/var/lib/flotilla/cleat`, backed by host state under `contained-cleat/<environment>`, and launch sessions with recording enabled
- update the cleat adapter to the current `cleat launch` contract and make attach strict with `attach --no-create`
- make ordinary mounted CLI commands honor `FLOTILLA_DAEMON_SOCKET`, allowing contained crews to use the host daemon socket without an in-container `flotillad`
- document the interim bind-mounted delivery model and the preferred future baked-image model

## Root cause

Launch accepted an explicitly requested interior pool but silently fell back to the registry's preferred provider when that pool was absent. Attach resolved the named pool exactly. That split launch and attach across two models and let a host-side process appear `Working` while the container held only its keepalive.

The current cleat adapter also used the removed `create` command. Finally, the mounted Flotilla CLI ignored `FLOTILLA_DAEMON_SOCKET` for normal commands, so a correctly contained agent could not complete work through the mounted socket.

## Impact

Contained crew engagement now has one model: the agent is a child of cleat inside the container, in `/workspace`, with the leased `CODEX_HOME`. Attach tunnels to that same interior session. Missing sessions cannot be recreated by attach, and existing liveness reconciliation turns a disappeared session into stopped/interrupted work instead of a stable phantom `Working` state. Recordings survive container teardown on the host-mounted runtime root.

## Validation

Focused coverage includes exact-pool/no-fallback behavior, cleat asset resolution and mounts, current cleat launch/attach arguments, disappeared-session stopping, and vessel interruption/restart behavior.

## Contained smoke report

Ran `single-agent-contained` on feta with `forgejo.lab.flotilla.work/image-builder/flotilla-crew:2026-07-26.1` and placement `pool: cleat`.

While live:

- container process tree showed `/usr/local/bin/cleat ... serve` as PID 275, with the Codex node/native processes as its descendants
- the agent's `/proc/<pid>/cwd` resolved to `/workspace`
- agent environment contained `CODEX_HOME=/run/flotilla/codex`
- `CLEAT_RUNTIME_DIR=/var/lib/flotilla/cleat`; the live `.cast` was present there
- MaterialPool `codex-login` showed `slot-00000000000000000000` leased to the contained environment
- from kiwi, `flotilla attach smoke-1278-interior-cleat-3` routed through feta and executed `cleat attach --no-create terminal-smoke-1278-interior-cleat-3-work-coder`, displaying the live interior Codex UI at `/workspace`

The final r4 run landed without intervention in about 11 seconds. Its completion message reported `cwd=/workspace`, `CODEX_HOME=/run/flotilla/codex`, `CLI+socket=ok`, and slot 0. After teardown:

- the Docker container was gone
- MaterialPool leases were empty
- the recording remained at `~/.local/state/flotilla/contained-cleat/env-smoke-1278-interior-cleat-4-work/default/sessions/terminal-smoke-1278-interior-cleat-4-work-coder/session.cast`
- slot-0 `auth.json` mtime/content size were unchanged, so no token refresh occurred during the short run

Closes #1278

### DIFF
--- a/.flotilla/placement-policy.crew-image.yaml
+++ b/.flotilla/placement-policy.crew-image.yaml
@@ -6,7 +6,7 @@ metadata:
   annotations:
     flotilla.work/image-recipe: .flotilla/Dockerfile.crew
 spec:
-  pool: passthrough
+  pool: cleat
   docker_per_vessel:
     host_ref: d49f4c59-811f-44aa-a4ca-d4cac66cf2a3
     image: forgejo.lab.flotilla.work/image-builder/flotilla-crew:2026-07-26.1
--- a/crates/flotilla-controllers/src/actuators/mod.rs
+++ b/crates/flotilla-controllers/src/actuators/mod.rs
@@ -3,7 +3,10 @@ use std::{path::PathBuf, sync::Arc};
 use flotilla_core::{
     path_context::{DaemonHostPath, ExecutionEnvironmentPath},
     providers::{
-        environment::{CreateOpts, EnvironmentProvider, ProvisionedMount, ProvisionedMountMode},
+        environment::{
+            CreateOpts, EnvironmentProvider, EnvironmentTool, EnvironmentToolAsset, EnvironmentToolAssetAccess, EnvironmentToolAssetKind,
+            EnvironmentVariableUpdate, ProvisionedMount, ProvisionedMountMode,
+        },
         terminal::TerminalPool,
         vcs::{CloneInspection, CloneProvisioner},
     },
@@ -45,11 +48,19 @@ impl DockerEnvironmentActuator {
     pub fn build_create_opts(&self, spec: &DockerEnvironmentSpec) -> CreateOpts {
         CreateOpts {
             tokens: self.tokens.clone(),
-            daemon_socket_path: self.daemon_socket_path.clone(),
             working_directory: None,
             image_pull_policy: spec.pull_policy.into(),
             docker_config_dir: None,
             provisioned_mounts: spec.mounts.iter().map(provisioned_mount).collect(),
+            tools: vec![EnvironmentTool::new("flotilla-daemon-access", "/usr/local/bin/flotilla")
+                .with_asset(EnvironmentToolAsset::new(
+                    self.daemon_socket_path.as_path().to_path_buf(),
+                    "/run/flotilla.sock",
+                    EnvironmentToolAssetKind::UnixSocket,
+                    EnvironmentToolAssetAccess::SharedWritable,
+                    "the daemon socket",
+                ))
+                .with_environment(EnvironmentVariableUpdate::set("FLOTILLA_DAEMON_SOCKET", "/run/flotilla.sock", "the daemon socket"))],
         }
     }
 }
--- a/crates/flotilla-controllers/src/reconcilers/presentation.rs
+++ b/crates/flotilla-controllers/src/reconcilers/presentation.rs
@@ -252,7 +252,6 @@ impl<R> PresentationReconciler<R> {
             .terminal_pools
             .get(&session.spec.pool)
             .map(|(_, pool)| Arc::clone(pool))
-            .or_else(|| registry.terminal_pools.preferred().cloned())
             .ok_or_else(|| format!("terminal pool {} unavailable for environment {}", session.spec.pool, session.spec.env_ref))?;
 
         let session_cwd = ExecutionEnvironmentPath::new(&session.spec.cwd);
--- a/crates/flotilla-core/examples/environment_checkout.rs
+++ b/crates/flotilla-core/examples/environment_checkout.rs
@@ -52,19 +52,12 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     println!("\n--- Creating environment ---");
     let env_id = EnvironmentId::new("smoke-test");
 
-    // We need git inside the container — install it after creation
-    // Use a temp socket path (we don't need a real daemon socket for this test)
-    let temp = tempfile::tempdir()?;
-    let socket_path = DaemonHostPath::new(temp.path().join("fake.sock"));
-    // Create a dummy socket file so the mount doesn't fail
-    std::fs::write(socket_path.as_path(), "")?;
-
     let opts = CreateOpts {
         tokens: vec![],
-        daemon_socket_path: socket_path,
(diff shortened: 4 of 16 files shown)


rjwittams:
Addressed in 3a40437a:

- Removed the remaining preferred-pool fallback from crew follow-up delivery and added a regression test proving a missing named interior pool fails instead of delivering through another pool.
- Cached `libghostty-vt` discovery with `tokio::sync::OnceCell`.
- Made missing cleat assets a hard Docker provisioning error before container creation, with a regression test. Built-in Docker placement is now advertised only when both Docker and the cleat terminal pool are available.
- Removed the redundant hook-level socket environment lookup; `Cli::socket_path()` remains the single precedence implementation.

I intentionally did not add a passthrough fallback: issue #1278's updated direction requires launch and attach to share the one interior-cleat model.

rjwittams:
Addressed in 3a40437a:

- Removed the remaining preferred-pool fallback from crew follow-up delivery and added a regression test proving a missing named interior pool fails instead of delivering through another pool.
- Cached `libghostty-vt` discovery with `tokio::sync::OnceCell`.
- Made missing cleat assets a hard Docker provisioning error before container creation, with a regression test. Built-in Docker placement is now advertised only when both Docker and the cleat terminal pool are available.
- Removed the redundant hook-level socket environment lookup; `Cli::socket_path()` remains the single precedence implementation.

I intentionally did not add a passthrough fallback: issue #1278's updated direction requires launch and attach to share the one interior-cleat model.

rjwittams:
Acknowledged — thanks for the re-review confirming the prior findings are resolved at 3a40437a and the PR is ready from static review.

rjwittams:
It's weird to have so much cleat specific code in runtime.rs - surely we need a more general 'tool in environment ' mechanism that cleat is one instantiation of. And this should be written in such a way that it would be applicable for other 'environments providers' - no need to solve now, but don't preclude.

rjwittams:
Reworked in 8d3a9353 around a general tool-in-environment seam.

`CreateOpts` now carries provider-neutral `EnvironmentTool` descriptions: the interior executable path, host-sourced assets classified as files/directories/Unix sockets, read-only vs shared-writable access, and environment mutations (`Set` or path prepend). The daemon has an extensible factory list in `environment_tools.rs`; the Flotilla CLI + daemon socket and cleat are separate consumers. Cleat-specific binary/library/state discovery has moved out of `runtime.rs`.

The environment provider owns delivery. Docker lowers those descriptions to bind mounts and environment arguments, but contains no cleat-specific paths or logic. A k8s/Modal provider can instead upload files or expose sockets through its transport while receiving the same tool description. Launch/attach continue through the existing per-environment runner, discovery registry, and hop chain, using the tool's interior executable path.

I kept this honest rather than building hypothetical providers: the current source is still a daemon-host path and Docker is the only lowering implemented, but file delivery, shared state, and socket exposure are no longer encoded as Docker/cleat behavior in the controller runtime. Added a generic Docker lowering contract test and updated the provisioning docs. Exact fmt, clippy, and workspace test gates pass.

rjwittams:
Thanks for the focused re-review. Both observations are intentional scope boundaries for this refactor: asset kind is semantic metadata for providers whose delivery mechanism must distinguish uploadable files, directories, and socket exposure, while Docker currently lowers all three to bind mounts; and the compatibility constructors in core/controllers remain explicit because moving shared well-known Flotilla tool constants across the current crate dependency boundaries would broaden this change. The daemon production path now has the reusable factory seam, and a shared descriptor home can follow when those legacy constructors are consolidated. No code change needed here.

PR #1288, opened 2026-07-31
Expose resumable resource reads through client and CLI

### LINKED ISSUE #1286: Expose resource get/list/watch through the client (build-order step 0)
## Parent

Map #1262 → the adopted #1283 substrate exploration (prototypes/1283-workflow-substrate.md), build-order step 0. Shared prerequisite for `flotilla wait`, the Bosun-over-verbs experiment, and the post-reshape TUI. Zero decisions deferred — control-plane work.

## What to build

The typed resource watch already exists at the store layer (`flotilla-resources` backend: cursors, provenance) but is not the client surface. Expose **resource get / list / watch** through `flotilla-client` and the CLI:

- `flotilla resource get/list` already exist — bring them to parity (full envelope, `--json` stable for scripts) and add **`flotilla resource watch <kind> [name] --json`**: cursor-addressed stream of mutations, level-triggered start (current state first, then deltas), resumable via `--from-cursor`, working both locally and via `--host` routing.
- Cursor + ordered-slice semantics only — no store-isms in the surface (the #1264 materialization rider binds here).
- Script-friendliness is the acceptance bar: one JSON object per line, exit codes distinguish stream-end from error, provenance visible per record.

## Acceptance criteria

- [ ] `flotilla resource watch convoy --json` streams current-then-deltas with cursors; a script can resume from a recorded cursor without loss
- [ ] Works cross-host through the existing peer routing; wire-generation guard applies as everywhere
- [ ] get/list/watch share one envelope shape; documented for script authors
- [ ] An InProcessDaemon-level test drives a watch through create/update/delete including a resumed cursor

rjwittams:
## Summary

- add a typed `flotilla-client` resource get/list/watch API with opaque resumable cursors
- expose cursor-addressed `resource watch <kind> [name] --json` locally and through `--host`
- unify get/list/watch output into a documented envelope with ordered records and per-record provenance
- route watch requests across peers and update replica consumers for the unified stream shape
- cover current-state bootstrap, create/update/delete deltas, cancellation, and resume with `InProcessDaemon`

## Verification

- `cargo test -p flotilla-protocol --locked`
- `cargo test -p flotilla-client --locked`
- focused core, daemon routing, and CLI parser tests

`cargo test --workspace --locked` passes all affected suites locally, but this macOS runner cannot resolve a machine identity for two pre-existing daemon runtime tests whose fixtures omit `machine_id`; both reproduce in isolation before exercising this change. CI's Linux runner provides `/etc/machine-id`.

Closes #1286

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1164,6 +1164,7 @@ dependencies = [
 name = "flotilla-protocol"
 version = "0.1.0"
 dependencies = [
+ "base64",
  "bon",
  "chrono",
  "gethostname",
--- a/crates/flotilla-client/src/lib.rs
+++ b/crates/flotilla-client/src/lib.rs
@@ -19,6 +19,7 @@ use tokio::sync::{broadcast, oneshot, Mutex};
 use tracing::{debug, error, warn};
 
 pub mod reconnect;
+pub mod resource;
 pub const BUILD_ID: &str = env!("FLOTILLA_BUILD_ID");
 
 /// Std RwLock for local seq tracking — the critical sections are single HashMap
--- a/crates/flotilla-client/src/resource.rs
+++ b/crates/flotilla-client/src/resource.rs
@@ -1,0 +1,276 @@
+use std::sync::Arc;
+
+use flotilla_core::daemon::DaemonHandle;
+use flotilla_protocol::{Command, CommandAction, CommandValue, DaemonEvent, NodeId, ResourceCursor, ResourceReadEnvelope, StepStatus};
+use tokio::sync::broadcast;
+
+#[derive(Debug, Clone, bon::Builder)]
+pub struct ResourceListRequest {
+    pub kind: String,
+    #[builder(default = "flotilla".to_string())]
+    pub namespace: String,
+    pub node_id: Option<NodeId>,
+    #[builder(default)]
+    pub include_replicas: bool,
+}
+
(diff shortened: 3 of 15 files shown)


rjwittams:
Addressed both substantive points:

- Added an integration-test negative case that puts `ignored-convoy` beside the requested object and updates it before the requested object. The test now proves the `name` filter excludes unrelated records from both the initial slice and the streamed loop.
- Kept the collection-position-before-get ordering because it is the part that prevents a concurrent mutation from being hidden behind the returned cursor. The O(n) SQLite cost is real, so I filed #1289 for a backend-level cheap current stream-position primitive that preserves that guarantee without decoding the collection.

The minor observations match existing resource metadata invariants and feature gating, so no code change was needed for those.

rjwittams:
Fixed the cursor-ordering edge case. `QueryResourceGet` now returns the collection-list error immediately and does not perform the point read, so every successful response is guaranteed to carry the collection cursor captured before the object read. The focused list/get integration test and `flotilla-core` clippy pass.

I left the internal step-description casing unchanged because it is not part of the wire or CLI JSON contract and no existing consumer relies on it.

rjwittams:
Reviewed. This follow-up confirms the prior findings are resolved and raises no new actionable issue; no additional code change is needed.

Issue #1290, opened 2026-07-31
Two #1281 runtime tests fail on macOS: machine identity unresolvable under fake discovery runner


rjwittams:
## What happens

`cargo test --workspace --locked` on a macOS host fails in `flotilla-daemon`:

- `runtime::tests::local_profile_does_not_advertise_docker_without_interior_cleat`
- `runtime::tests::contained_terminal_session_never_falls_back_from_the_requested_interior_pool`

Both panic in `InProcessDaemon::new` with:

```
failed to resolve local node id: "Cannot determine machine identity. Set `machine_id` in daemon.toml."
```

## Why

Both tests (added with #1281) construct `InProcessDaemon::new` through `fake_discovery_with_provider_set`, whose `DiscoveryMockRunner` only answers `git --version`. Machine identity resolution (`host_identity.rs`) then falls through:

1. `machine_id` from daemon.toml — absent (fresh `TempDir` config)
2. `/etc/machine-id` — Linux-only, absent on macOS
3. `IOPlatformUUID` via `ioreg` through the injected runner — the mock doesn't answer it

So the tests pass on Linux CI (step 2 succeeds via the *real* `/etc/machine-id` — itself a mild injection leak) and fail on any macOS checkout. Surfaced by running the workspace suite on kiwi; pre-existing on `main` (`083b2366`), unrelated to any in-flight branch.

## Fix direction

Make identity resolution in these tests injected rather than host-dependent — either write a `daemon.toml` with `machine_id` into the temp config store in the two tests, or have `fake_discovery_with_provider_set` answer the `ioreg` probe (and consider whether `read_etc_machine_id` should also go through an injected collaborator so Linux CI isn't silently reading real host state per the discovery conventions in CLAUDE.md).

## Acceptance criteria

- [ ] Both tests pass on macOS and Linux without a real `/etc/machine-id` or `ioreg`
- [ ] No other `InProcessDaemon`-constructing test depends on real host identity state

PR #1291, opened 2026-07-31
feat: scope forgejo pr delivery for crews

rjwittams:
Extends the Forgejo credential adapter so contained crews can actually deliver PRs to the lab forge, per the lab-forks phase-1 runway (project-map `lab-forks-plan.md`).

## What changes

- `CredentialConsumer::Forgejo` gains a `username`; delivery now writes a **destination-scoped git credential helper** into the environment (gated to the exact HTTPS host:port, reads the token from `FORGEJO_TOKEN_FILE` at invocation time, never embeds the secret) and wires it via `GIT_CONFIG_KEY_0/VALUE_0`, alongside `FORGEJO_SERVER_URL`, `FORGEJO_API_URL`, `FORGEJO_USERNAME`, and `GIT_TERMINAL_PROMPT=0`. HTTPS is enforced; token file 0600, helper 0700; the curl preflight is retained.
- Crew, fork-stance, and shepherd briefs become **forge-aware**: the PR destination is the repository URL and target ref in `## Work context` (the issue source may be a different forge); push only to the existing remote matching that destination, never add or repoint remotes; no `gh` or GitHub-only shepherd helpers against a Forgejo destination; if the injected credentials are unavailable or rejected, fail the assignment rather than deliver to another forge.
- Adds a `curl` host detector so the credential's `placement.binaries` requirement is discoverable.

## Notes

- Schema note: `username` is a required addition to a persisted variant. No pre-existing Forgejo `CredentialSpec` exists anywhere in the fleet (verified: no decode quarantine on kiwi/feta/udder), so no store reset is needed.
- Local gates on macOS: fmt and clippy clean; workspace tests clean except two pre-existing `main` failures tracked in #1290 (macOS-only machine-identity resolution in two #1281 tests, unrelated to this change).


rjwittams:
All three findings addressed in e99616ea (comment-id: 5145767352):

1. **`shepherd.md` assignment contradiction** — the `assignment` block is now forge-aware: "with the `pr-shepherd` skill for a GitHub destination, or through the injected Forgejo API credentials for a Forgejo destination". The template test asserts the new phrasing and asserts the old unconditional "using the `pr-shepherd` skill" wording is gone.

2. **Silent clobber between same-adapter credentials** — `prepare()` now pre-collects specs and fails loudly (mirroring the `prepare_registry_pull` precedent) when two granted credentials share an adapter, before any material is resolved or written. Generalized to all fixed-env-name adapters rather than just Forgejo, since two `Gh` credentials had the identical failure mode; `docker-registry` stays exempt because it multiplexes per image with its own multiplicity check. When per-credential namespacing exists this can relax, but until then fail-closed is the contract.

3. **Validation-branch coverage** — four new tests: non-HTTPS rejection, unparsable-URL rejection, explicit-port agreement between `GIT_CONFIG_KEY_0` and the generated helper's `host:port` comparison, and the loud two-credential failure. The rejection-path tests also assert nothing is written to the environment.

The quoting nit is left as-is per your own assessment (fixed literal baked at generation time).

rjwittams:
Thanks — the `api_url`→`server_url` rename is agreed but deferred to #1294 (comment-id: 5151338982): the field name is persisted in kiwi's applied `CredentialSpec` and the project-map manifest, so the rename needs a coordinated manifest update + resource reapply at deploy time rather than riding this PR after convergence. No other action from this round.

PR #1300, opened 2026-08-01
fix: bound terminal session reconcile failures

### LINKED ISSUE #1297: Abandoned convoy leaks its TerminalSession, which fails reconcile every minute forever
## Observed (2026-08-01)

`docker-crew-smoke-r4` was abandoned on 30 Jul (#1278 false run). Two days later, feta's daemon log is dominated by, once per minute, unbounded:

```
WARN controller failed to reconcile object; other objects will continue
  resource_kind=TerminalSession resource=terminal-docker-crew-smoke-r4-work-coder
  err="provider registry unavailable for environment env-docker-crew-smoke-r4-work; ..."
```

The convoy's abandonment tore down (or never completed) the environment, but its `TerminalSession` resource survived and the controller retries it forever — no backoff cap, no terminal failure state, no garbage collection tied to the owning convoy's lifecycle. Round 3 of the smoke effort saw the same spam end in "budget exhaustion"; this specimen shows the budget either regressed or doesn't apply on this path.

## What should happen

- Abandoning/deleting a convoy must reap its subordinate resources (TerminalSession et al.) — ownership/finalizer chain, not best-effort.
- Independently, a reconcile that fails with the same error for N consecutive rounds should reach a terminal degraded state with escalating backoff, not a permanent 60s spin that buries every other WARN in the log (it materially slowed diagnosing #1295).

## Acceptance

- [ ] Abandoning a convoy removes or terminally parks its TerminalSessions
- [ ] Repeated identical reconcile failures back off and eventually stop retrying, with a visible degraded condition
- [ ] Existing orphan on feta is cleaned up by the fix (or a documented one-off)

rjwittams:
Closes #1297

## Summary

- reap TerminalSessions as soon as their owning convoy is abandoned, before provider lookup
- add opt-in per-object reconcile failure budgets with exponential backoff and terminal parking
- persist a visible TerminalSession degraded condition after five identical failures

## Verification

### DIFF
--- a/crates/flotilla-controllers/src/reconcilers/terminal_session.rs
+++ b/crates/flotilla-controllers/src/reconcilers/terminal_session.rs
@@ -100,6 +100,22 @@ impl<R> TerminalSessionReconciler<R> {
             .map(|source| source.object)
             .ok_or_else(|| ResourceError::not_found(convoy_ref))
     }
+
+    async fn session_owner_missing(&self, session: &ResourceObject<TerminalSession>) -> Result<bool, ResourceError> {
+        let convoy_ref = match &session.spec.source {
+            TerminalSessionSource::Agent { context, .. } => Some(context.convoy.as_str()),
+            TerminalSessionSource::Tool { .. } => session.metadata.labels.get(CONVOY_LABEL).map(String::as_str),
+        };
+        let Some(convoy_ref) = convoy_ref else {
+            return Ok(false);
+        };
+        match self.convoy_for_session(session, convoy_ref).await {
+            Ok(convoy) => Ok(convoy.metadata.deletion_timestamp.is_some()
+                || convoy.status.as_ref().is_some_and(|status| status.phase == ConvoyPhase::Abandoned)),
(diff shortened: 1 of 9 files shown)


rjwittams:
Fixed all three lifecycle concerns in 1d20ec36.

- The loop now parks only after the degraded status patch is persisted. Missing or failed persistence remains retryable.
- Failure state is scoped to the resource creation timestamp and cleared on NotFound, so deletion/recreation cannot inherit a stale budget.
- Degraded reconcilers can opt into lifecycle wakeups; TerminalSession uses that hook to re-check ownership, so later convoy abandonment still deletes an already-degraded session.

Added regression coverage for persistence-unavailable retry, same-name recreation, and abandonment after degradation. I also documented the stable-message assumption for identical-error counting.

rjwittams:
Confirmed. The follow-up review validates the lifecycle fixes and identifies no further actionable change; the duplicated status predicate is a pure, cheap read and can remain as-is in this patch.

PR #1301, opened 2026-08-01
fix: clean stale reverse peer sockets

### LINKED ISSUE #1295: Stale reverse peer socket permanently partitions the mesh after a daemon bounce
## Observed (2026-08-01, after the b5b7ad7a fleet bump — but armed since 30 Jul)

After bouncing all three daemons, kiwi could not re-establish peering with feta or udder: 13+ redial attempts over 10 minutes, every one failing with `peer closed before sending hello`; the accepting side logs `failed to read connection preface: early eof` at matching timestamps. Fleet showed both peers disconnected/STALE. Forensics showed the partition actually began **30 Jul evening** — two days of silent partition that fleet diagnosis reported only as STALE rows nobody was looking at.

## Mechanism

The SSH transport spawns per-attempt `ssh -N -L <local>:<remote-daemon> -R <remote>/.peer-<hash>:<local-daemon> -o ExitOnForwardFailure=yes -o StreamLocalBindUnlink=yes` (`crates/flotilla-daemon/src/peer/ssh_transport.rs`).

The trap:
1. A tunnel dies (daemon bounce, network drop). sshd does **not** unlink the `-R` streamlocal socket file it bound (`~/.config/flotilla/.peer-<node-hash>` on the accepting host).
2. On every subsequent dial, the `-R` bind fails because the path exists. `StreamLocalBindUnlink=yes` as a **client** `-o` option does not govern the server-side bind — that requires `StreamLocalBindUnlink yes` in the **accepting host's sshd_config**, which is not set (and shouldn't need to be a fleet prerequisite).
3. `ExitOnForwardFailure=yes` then kills the whole ssh child — after the `-L` local bind succeeded and the dialer's hello connection was accepted — so the dialer sees `peer closed before sending hello` and the acceptor sees a zero-byte connection (`early eof`).
4. The redial loop retries forever with the same doomed bind. The partition is permanent until someone hand-deletes the socket file.

Smoking gun: `.peer-8f640342b921ba37` (kiwi's node hash) dated 30 Jul on both feta and udder; removing both files healed the mesh within one redial cycle, no restarts needed.

## Impact of the two-day partition (all one root cause)

- Three crews on feta ran `flotilla crew complete` after delivering (#1279, #1280, #1281 all merged) and got `resource not found: <convoy>` — the authority was unreachable — leaving their convoys phase `Working` indefinitely (filed separately).
- Resource replication and manifest convergence silently limited to whichever hosts were reachable.

## Fix directions

1. **Dialer pre-cleans the remote path**: before spawning the tunnel, run `ssh <host> rm -f <derived .peer path>` (the path is already derived deterministically on both ends). Works regardless of server sshd config; adds one round-trip per redial attempt, not per message.
2. Consider the accepting daemon also unlinking any `.peer-*` files in its socket directory at startup (covers the bounce-while-dialer-quiet case cheaply).
3. **Loudness**: N consecutive `peer closed before sending hello` failures should escalate beyond a log line — a degraded fleet condition naming the probable stale-socket cause, since the symptom is otherwise indistinguishable from a network problem.
4. A liveness-harness sequence (ADR 0024 vocabulary): tunnel established → acceptor restart without socket cleanup → redial must recover. This partition class has now bitten twice (see also the capped-redial recovery work in #1053).

rjwittams:
## Summary

- remove the deterministic reverse peer socket over SSH before each tunnel dial
- clean orphaned `.peer-<node-hash>` socket entries when the accepting daemon starts
- surface cleanup failures with the remote host and stale path instead of an ambiguous hello EOF
- cover successful pre-clean ordering, failure diagnostics, shell quoting, and startup cleanup

## Testing

Closes #1295

### DIFF
--- a/crates/flotilla-daemon/src/peer/ssh_transport.rs
+++ b/crates/flotilla-daemon/src/peer/ssh_transport.rs
@@ -616,36 +616,39 @@ mod tests {
     #[test]
     fn resource_socket_forward_specs_are_bidirectional() {
         let config = RemoteHostConfig {
-            hostname: "feta.local".to_string(),
-            expected_host_name: "feta".to_string(),
+            hostname: "peer-a.example.invalid".to_string(),
+            expected_host_name: "peer-a".to_string(),
             expected_node_id: None,
-            user: Some("flotilla".to_string()),
-            daemon_socket: "/home/flotilla/.config/flotilla/flotilla.sock".to_string(),
+            user: Some("test-user".to_string()),
+            daemon_socket: "/home/test-remote/.config/flotilla/flotilla.sock".to_string(),
             ssh_multiplex: None,
         };
-        let local_node = NodeId::new("kiwi-root");
+        let local_node = NodeId::new("local-node");
         let transport = SshTransport::new(
             local_node.clone(),
-            "kiwi".into(),
-            ConfigLabel("feta".to_string()),
+            "local-host".into(),
+            ConfigLabel("peer-a".to_string()),
             config,
             None,
             uuid::Uuid::nil(),
             SshTransportPaths {
-                state_dir: Path::new("/home/kiwi/.local/state/flotilla"),
-                daemon_socket: Path::new("/home/kiwi/.config/flotilla/flotilla.sock"),
+                state_dir: Path::new("/home/test-local/.local/state/flotilla"),
+                daemon_socket: Path::new("/home/test-local/.config/flotilla/flotilla.sock"),
             },
         )
         .expect("valid transport");
 
         let (local_forward, reverse_forward) = transport.resource_forward_specs();
 
-        assert_eq!(local_forward, "/home/kiwi/.local/state/flotilla/peers/feta.sock:/home/flotilla/.config/flotilla/flotilla.sock");
+        assert_eq!(
+            local_forward,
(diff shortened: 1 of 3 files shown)


rjwittams:
Addressed both findings. The startup sweep now warns and continues for directory-entry, metadata, and removal failures, so an orphaned reverse socket cannot prevent the daemon's primary listener from starting; per-dial remote cleanup remains strict for the affected peer. I also filed #1303 for the user-visible consecutive-failure degradation and the full acceptor-restart/redial liveness harness.

rjwittams:
Verified. No additional code change is needed for this non-blocking note: the startup unlink is the intended recovery behavior from #1295, and the outbound peer loop should detect the daemon-side EOF, retire the old SSH child, and redial to recreate the reverse listener. #1303 now explicitly owns the real restart/redial liveness sequence so that timing and convergence are exercised end to end.

rjwittams:
Verified both non-blocking observations. No further change in this fixture-only update: the shared 10-second value intentionally bounds both setup phases, while separating or renaming the constant would be a readability-only follow-up; SSH multiplex wiring is pre-existing transport behavior outside #1295. The requested hostname cleanup is complete, and the fake SSH tests remain fully isolated from DNS and the network.

PR #1302, opened 2026-08-01
fix: preserve attach refusal visibility

### LINKED ISSUE #1299: flotilla attach: refusal errors invisible under alternate screen; no watcher fallback
## Observed (2026-08-01, Robert, live)

Attaching to a session whose controller seat is held fails with "already has a foreground client" — but the message is only discoverable by recording the terminal output, because the attach flow enters/clears the alternate screen and the teardown wipes the error before the user can see it. To the user, attach silently bounces back to a blank prompt.

## What should happen

1. **Errors must survive screen teardown**: restore the primary screen first, then print the failure to stderr as the last output. Nothing may be printed to the alternate screen that the user needs after exit.
2. **Refusal should name the holder and offer the watch seat**: cleat#178 gives sessions one controller plus read-only watchers. `flotilla attach` on a held session should say who holds control (host/pid or attach origin if available) and either suggest or take `--watch` — a read-only view is almost always what the second attacher wants (it is exactly what the governor's probe wanted against a working crew).

## Acceptance

- [ ] A refused attach leaves the error visible on the user's terminal after exit
- [ ] `flotilla attach --watch <ref>` attaches read-only to a controlled session
- [ ] The refusal message names the controlling attachment and mentions --watch

Related: #1298 (the seat in the observed case was held by a leaked dead client — fixing that removes this instance's cause, but refusal UX applies whenever a live controller legitimately exists).

rjwittams:
Closes #1299.

- buffer attach stderr and replay it only after restoring the primary screen
- pass cleat's default, --strict, and --take controller-seat policy through local, container, and remote attach plans
- preflight the selected cleat pool for the #178 capability surface and name stale binaries and versions
- rely on modern cleat's native watcher banner to announce a degraded watcher seat and identify the controller

Checks:

### DIFF
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -91,6 +91,7 @@ use crate::{
         },
         issue_tracker::{forge_issue_source, IssueProvider},
         ssh_runner::SshCommandRunner,
+        terminal::AttachSeat,
         ChannelLabel, CommandRunner,
     },
     refresh::RefreshSnapshot,
@@ -530,10 +531,16 @@ enum AttachTarget {
 }
 
 impl AttachTarget {
-    async fn resolve(&self, daemon: &InProcessDaemon, reference: &str, transient: bool) -> Result<ResolvedAttach, String> {
+    async fn resolve(
+        &self,
+        daemon: &InProcessDaemon,
+        reference: &str,
+        transient: bool,
+        seat: AttachSeat,
+    ) -> Result<ResolvedAttach, String> {
         match self {
             Self::Local(session) => {
-                let (plan, host) = daemon.attach_plan_for_session(reference, session).await?;
+                let (plan, host) = daemon.attach_plan_for_session(reference, session, seat).await?;
                 let labels = &session.metadata.labels;
                 let binding = AttachBinding::builder()
                     .host(host)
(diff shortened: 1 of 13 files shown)


rjwittams:
Addressed the stderr lifetime concern in 49df1855: child stderr is now teed live during the interactive attach, with only a bounded 64 KiB tail retained for post-teardown replay on failure. Successful attaches are not replayed, so diagnostics stay live without duplication or unbounded growth. The new regression test covers the live tee and retained bound.

The other notes referenced the earlier --watch implementation, which the revised issue scope removed. The branch now passes through cleat's native --strict/--take modes and uses capability preflight instead.

rjwittams:
Followed up on the current review in 8c4e6954:

- Default mode intentionally runs the capability preflight. The updated #1299 scope says the stale pre-#178 binary specifically breaks cleat's default degrade-to-watch behavior, so checking only explicit --strict/--take requests would miss the reported failure mode.
- The capability result is now cached once per selected terminal pool, removing the repeated subprocess cost.
- Attach arguments are now constructed in semantic order rather than splicing flags at a positional index.

The full pinned-format, workspace-clippy, and locked workspace-test gates all pass locally after this follow-up.

rjwittams:
Addressed the checkout-session ordering finding in ec7618c0. Capability preflight now runs before ensure_session, so a stale selected cleat fails without creating an orphaned transient checkout session. Added a regression test that injects a preflight failure and asserts no session was ensured.

Pinned formatting, workspace clippy with warnings denied, the focused regression, and the full locked workspace suite all pass locally.

rjwittams:
Followed up on the non-blocking cache observation in 5c70d309: capability successes remain cached, but failures are not. A cleat upgraded in place can therefore pass the next attach preflight without restarting the daemon. Added a stale-then-upgraded regression test.

All exact local gates pass on this head.

PR #1304, opened 2026-08-01
fix: make crew completion resilient to partitions

### LINKED ISSUE #1296: Failed crew completion strands the convoy in phase Working silently
## Observed (2026-08-01)

Three convoys (`field-ownership-write-path` #1251, `harness-transition-vocab` #1252, `contained-agent-launch` #1278) sat in phase `Working` with crews idle for two days **after their PRs (#1279/#1280/#1281) were delivered and merged**. Each crew had run its final act — `flotilla crew complete --message '<PR URL>'` — and received:

- `plane returned: resource not found: field-ownership-write-path`
- `resource no longer exists (resource not found: harness-transition-vocab)`
- `contained-agent-launch was not found`

Root cause of the *lookup* failure was the mesh partition in #1295 (crew on feta, convoy authority on kiwi, tunnel dead). This issue is about what happened next: **nothing**. The error existed only as text on the agent's terminal screen; the agent moved on; the convoy stayed `Working` — a stable lie, the same phantom-phase class as #1278's respawn loop (ADR 0024 class 2), now in the completion verb itself.

## What should happen

1. **Completion must be durable, not fire-and-forget.** A `crew complete` that fails on a transport/authority error should be queued/retried by the crew CLI or the local daemon (the local daemon knows the convoy exists — it has a replica; "resource not found" from a partitioned authority is not proof of nonexistence and must not be presented as such).
2. **A failed completion must be loud on the convoy**, not only on the crew's screen: a visible condition/event ("completion attempted, authority unreachable") so fleet diagnosis shows the convoy as completion-pending rather than healthily Working.
3. The error text itself is wrong: a follower that cannot reach the authority should say so (`authority unreachable for <convoy>`), never `resource not found` — three agents each reasonably concluded the convoy had been deleted and stopped trying.

This maps directly onto the workflow-substrate direction (#1283/#1285): completion is a verb whose attempt belongs in the turn log with an observed disposition, and phase should be reconciled from observations rather than left wherever the last successful write put it.

## Acceptance

- [ ] A completion attempted during an authority partition eventually lands without human intervention once the partition heals
- [ ] The convoy surfaces a visible completion-pending/degraded state in the interim
- [ ] Follower-side resolution failures name the authority as unreachable instead of claiming the resource does not exist

rjwittams:
Closes #1296

## Summary
- persist crew-completion intent on the local terminal session before routing to the convoy authority
- retry pending completion automatically after authority reachability returns and clear it only after acknowledgement
- surface completion-pending state in convoy/vessel health and report authority-unreachable routing errors
- route ambient crew commands through locally resolvable terminal-session context

## Validation

### DIFF
--- a/crates/flotilla-controllers/tests/common/mod.rs
+++ b/crates/flotilla-controllers/tests/common/mod.rs
@@ -358,6 +358,7 @@ pub async fn create_stopped_terminal(
             launch_command: Some(fixture.command),
             delivered_message_id: None,
             attention: None,
+            completion_pending: None,
         })
         .await
         .expect("terminal status update should succeed");
--- a/crates/flotilla-controllers/tests/vessel_reconciler.rs
+++ b/crates/flotilla-controllers/tests/vessel_reconciler.rs
@@ -2204,6 +2204,7 @@ async fn create_running_terminal(
             launch_command: Some(command.to_string()),
             delivered_message_id: None,
             attention: None,
+            completion_pending: None,
         })
         .await
         .expect("terminal status update should succeed");
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -6073,10 +6073,16 @@ impl InProcessDaemon {
     }
 
     pub async fn crew_complete_internal(&self, requested: &CrewCommandContext, message: Option<String>) -> Result<(), String> {
+        let routing = self.resolve_crew_routing_context(requested).await?;
         self.apply_crew_work_patch(requested, |context| {
             convoy_external_patches::mark_crew_completed(context.vessel.clone(), context.caller_role.clone(), chrono::Utc::now(), message)
         })
-        .await
+        .await?;
+        if let Some(session_name) = routing.session_name {
+            let namespace = routing.command_context.namespace.as_deref().expect("resolved crew routing context has a namespace");
+            self.clear_crew_completion_pending(namespace, &session_name).await?;
+        }
+        Ok(())
     }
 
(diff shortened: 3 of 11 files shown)


rjwittams:
Fixed in 20332912. The retry slot now stores the latest completion payload plus a generation, reads that payload before every dispatch, and only retires the slot when the successfully dispatched generation is still current. I also extended the partition lifecycle test to reissue completion with a corrected PR URL and assert that the healed route sends the corrected URL. Exact fmt, clippy, and full workspace test gates pass locally.

rjwittams:
Fixed in ed50e3ff. `crew_complete_internal` now resolves the caller terminal, applies the convoy completion patch, and only after that succeeds clears the matching durable `completion_pending` marker. The ambient local-completion test now seeds a pending marker (representing authority failover to this host) and asserts both completed crew work and marker retirement. Exact fmt, clippy, and full workspace tests pass against current main.

rjwittams:
Fixed in 74c85028. Authority `CommandValue::Error` is now treated as an explicit acknowledgement: the durable pending marker is cleared and the original command error remains visible, but no retry is scheduled. Transport/routing failures remain the only paths that persist and re-arm retry. The partition lifecycle test now asserts a permanent authority rejection clears pending state, produces no later retry, and still retains the successful-ack path for a fresh completion. Exact fmt, clippy, and full workspace tests pass.

rjwittams:
Resolved by ed50e3ff as described in the subsequent review: successful local completion now clears the durable marker after applying the convoy patch, with a regression that seeds and retires pending state.

rjwittams:
Thanks for the final verification. The two remaining observations are explicitly non-blocking and outside issue #1296\u2019s crew-completion scope; no further change is needed for this PR.

PR #1306, opened 2026-08-01
fix: reap cross-host attaches on client disconnect

### LINKED ISSUE #1298: Cross-host attach leaks the interior controller seat when the client dies
## Observed (2026-08-01, smoke r5)

`flotilla attach docker-crew-smoke-r5` from kiwi builds this chain: local client → feta daemon → `docker exec -it -w /workspace <container> /bin/sh` → interior `cleat attach --no-create <session>`, which takes the interior session's **controller seat**.

When the kiwi-side client died (the probe session hosting it was killed), the hangup propagated zero hops: an hour later both the feta-side `docker exec` client and the interior attach client were still alive, holding controller. Every subsequent attach (Robert, live) was then refused by cleat's one-controller rule — correctly by cleat's lights, but against a dead man's seat. Manual remediation required *two* kills, because docker compounds the problem: killing the `docker exec` client is documented **not** to kill the exec'd process, so the interior client had to be killed inside the container.

## What should happen

The daemon leg that spawns the exec owns the attach chain's lifetime:

- When the upstream client connection drops (for any reason — kill, network, terminal close), the daemon must tear down the interior attach client it created. Because `docker exec` client death does not kill the exec'd process, the teardown must be explicit — e.g. track the interior client's pid (or run it under a supervised wrapper) and `docker exec kill` it on disconnect.
- Idle/orphan insurance: the interior attach client should also die if its stdin/stdout PTY goes away (it survived its PTY's exec client dying — worth checking why the hangup didn't reach it).
- A liveness-harness sequence: attach, kill the outer client ungracefully, assert the controller seat frees within a bounded interval.

## Acceptance

- [ ] Killing the outermost attach client (SIGKILL, no cleanup) frees the interior controller seat within seconds
- [ ] No `docker exec` or interior `cleat attach` processes linger after client death
- [ ] A second attach immediately after ungraceful death of the first succeeds

rjwittams:
Closes #1298.

## What changed

- Add connection-scoped attach excursions: the daemon registers compensating actions before launch and drains them on explicit completion or client EOF.
- Compose cleanup through every transport boundary, including routing Docker cleanup back through the outer SSH hop.
- Tear down innermost-to-outermost: kill the leased interior process first, then remove the outer cleat bridge.
- Stamp Docker-interior attach processes with a unique lease and verify both PID and lease identity before killing, preventing PID-reuse kills.
- Bound each daemon cleanup action so a failed inner transport cannot indefinitely prevent outer teardown.

## Root cause

Multi-hop attaches launched persistent side effects, but ownership lived in client-local signal and panic hooks. SIGKILL and network loss bypassed those hooks. Docker also deliberately leaves an execd interior process alive when the docker exec client disappears, so the interior cleat attach continued holding its controller seat.

The daemon connection is now the lifetime boundary. When that connection disappears, its remaining attach excursions are reaped without relying on client code running.

## Validation

- cargo test -p flotilla-daemon client_disconnect_reaps_docker_exec_and_frees_interior_seat --locked -- --ignored --nocapture

The Docker liveness harness verifies that disconnecting the upstream client reaps the docker exec client and interior PID, then permits an immediate second seat acquisition.

### DIFF
--- a/crates/flotilla-client/src/lib.rs
+++ b/crates/flotilla-client/src/lib.rs
@@ -995,6 +995,24 @@ impl DaemonHandle for SocketDaemon {
         }
     }
 
+    async fn begin_attach_excursion(
+        &self,
+        excursion_id: flotilla_protocol::AttachExcursionId,
+        cleanup_actions: Vec<Vec<flotilla_protocol::arg::Arg>>,
+    ) -> Result<(), String> {
+        match into_success_response(self.request(Request::BeginAttachExcursion { excursion_id, cleanup_actions }).await?)? {
+            Response::BeginAttachExcursion => Ok(()),
+            other => Err(format!("unexpected response for begin attach excursion: {other:?}")),
+        }
+    }
+
+    async fn finish_attach_excursion(&self, excursion_id: flotilla_protocol::AttachExcursionId) -> Result<(), String> {
+        match into_success_response(self.request(Request::FinishAttachExcursion { excursion_id }).await?)? {
+            Response::FinishAttachExcursion => Ok(()),
+            other => Err(format!("unexpected response for finish attach excursion: {other:?}")),
+        }
+    }
+
     /// Execute a query command and return the result directly.
     ///
     /// The `session_id` parameter is ignored by `SocketDaemon` because cursor
--- a/crates/flotilla-core/src/daemon.rs
+++ b/crates/flotilla-core/src/daemon.rs
@@ -2,8 +2,8 @@ use std::collections::HashMap;
 
 use async_trait::async_trait;
 use flotilla_protocol::{
-    commands::CommandValue, Command, DaemonEvent, QueryCursor, RepoInfo, RepoSelector, RepoSnapshot, StatusResponse, StreamKey,
-    TopologyResponse,
+    arg::Arg, commands::CommandValue, AttachExcursionId, Command, DaemonEvent, QueryCursor, RepoInfo, RepoSelector, RepoSnapshot,
+    StatusResponse, StreamKey, TopologyResponse,
 };
 use tokio::sync::broadcast;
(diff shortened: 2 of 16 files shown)


rjwittams:
Both findings were valid and prompted a broader lifecycle correction in 0e87fc25.

1. Cleanup actions are now composed through outer hops. The SSH resolver rewrites existing inner cleanup actions as non-interactive SSH commands, and a Remote -> Docker -> Terminal test asserts that the Docker cleanup targets the remote host.
2. The detached client watchdog has been removed. Attach cleanup is registered as a connection-scoped daemon excursion before launch and is drained on explicit finish or client EOF. Docker cleanup verifies both the recorded PID and a unique lease value in /proc/<pid>/environ before sending SIGKILL, which closes the PID-reuse race.

Teardown is ordered innermost-to-outermost, each action is bounded, and a real Docker harness now verifies that socket disconnect reaps both docker exec and the interior process before a second seat acquisition.

rjwittams:
Verified after merging current main. Both remaining notes are intentional bounded tradeoffs rather than requested changes: attach cleanup is scoped to the daemon control connection, and best-effort teardown retains the existing five-second per-action timeout. The merge also preserves main's attach-refusal stderr visibility while keeping bridge and environment cleanup daemon-owned and ordered inside-out.

rjwittams:
Confirmed: `SocketDaemon` is the only production attach transport, while the other current `DaemonHandle` implementations are test/in-process fixtures. The default error is intentional fail-closed behavior: a non-trivial attach must not launch unless some owner has accepted its cleanup excursion, because graceful fallback would recreate the leak this PR fixes. If another daemon implementation becomes a production attach transport, excursion support is therefore part of that implementation's capability contract rather than an optional degradation path. No change or tracking issue is needed for the current scope.

PR #1307, opened 2026-08-01
fix: reap abandoned checkout worktrees

rjwittams:
Follow-up to #1300 for the checkout leak added to #1297 after the original PR merged.

## Summary

- prune the base clone after removing every managed checkout worktree
- keep checkout finalization pending if directory removal or worktree pruning fails
- verify abandoned convoys reclaim managed checkouts while adopted checkout resources remain as owner records

## Verification


rjwittams:
Addressed the test-coverage suggestion in 5e508785. The regression now creates a second registered worktree, removes its directory behind Git's back, confirms the stale base-clone admin record exists, then verifies checkout cleanup's `git worktree prune` removes that record. The exact format, Clippy, and full workspace test commands all pass after the change.

rjwittams:
Confirmed. The follow-up review has no blocking feedback, all four required checks pass, and the PR is ready for merge.

PR #1311, opened 2026-08-01
fix: decouple environment teardown from mount metadata

### LINKED ISSUE #1310: Mount-metadata schema change breaks teardown of pre-upgrade environments
## Observed (2026-08-01, after the six-fix bump)

feta's Environment reconciler failed permanently on the (pre-bump-provisioned) smoke-r5 environment:

```
WARN failed to parse provisioned mount metadata  container=flotilla-env-…  err="missing field `mode` at line 1 column 78"
WARN controller failed to reconcile object  resource_kind=Environment  resource=env-docker-crew-smoke-r5-work  err="failed to parse provisioned mount…"
```

The mount metadata is persisted on the docker container itself (labels), written by the pre-bump daemon without `mode`; the post-bump parser requires it. Because the reconciler cannot *parse* the environment, it cannot *tear it down* — the container, and everything downstream (TerminalSessions retrying against it, checkout reaping), stayed stuck until manual `docker rm -f` plus surgical resource deletion.

## Why the no-compat phase doesn't excuse this one

Store-reset bumps handle the resource store; **actuated-object metadata (docker labels) survives daemon bumps and has no reset mechanism**. Any required-field addition to that metadata orphans every environment provisioned before the bump — and the failure mode is the worst kind: the orphan is *unreconcilable*, so even deletion is blocked.

## What should happen

1. **Teardown must never require full metadata parse.** Destroy paths should work from the container name/identity alone (or a minimal always-stable subset), so an unparseable environment can always be reclaimed.
2. Mount-metadata parse failures on the *reconcile* path should degrade that environment loudly (condition on the resource) rather than retry-warn forever.
3. Convention: fields added to actuated-object metadata are always `#[serde(default)]`-tolerant on read, even during the no-compat phase — the writer may be strict, the reader must not be, because old writers' output outlives them on live containers.

## Acceptance

- [ ] An environment whose metadata fails to parse can still be torn down (liveness sequence: provision, corrupt/age the label, delete convoy → container removed)
- [ ] Parse failure surfaces as a visible degraded condition, not an eternal WARN loop

rjwittams:
Closes #1310

## What changed

- add an environment-provider teardown operation that accepts the stable container identity directly
- use that operation after daemon restart instead of listing and fully parsing container mount metadata
- default mount metadata written before the `mode` field existed to its historical read-only behavior
- surface environment finalizer failures through the resource's failed status and message

## Root cause

Restart teardown rediscovered a container through `EnvironmentProvider::list()`. That path deserialized the full `flotilla.provisioned_mounts` label before returning a handle, so a required-field schema change made the container impossible to reclaim.

## Validation

- focused legacy-label, restart-teardown, and finalizer-status regression tests

### DIFF
--- a/crates/flotilla-controllers/src/reconcilers/environment.rs
+++ b/crates/flotilla-controllers/src/reconcilers/environment.rs
@@ -122,4 +122,13 @@ where
     fn finalizer_name(&self) -> Option<&'static str> {
         Some("flotilla.work/environment-teardown")
     }
+
+    fn finalizer_error_patch(&self, obj: &ResourceObject<Self::Resource>, error: &ResourceError) -> Option<EnvironmentStatusPatch> {
+        let message = format!("environment teardown failed: {error}");
+        if obj.status.as_ref().is_some_and(|status| status.phase == EnvironmentPhase::Failed && status.message.as_deref() == Some(&message))
+        {
+            return None;
+        }
+        Some(EnvironmentStatusPatch::MarkFailed { message })
+    }
 }
--- a/crates/flotilla-controllers/tests/actuators.rs
+++ b/crates/flotilla-controllers/tests/actuators.rs
@@ -47,6 +47,10 @@ impl EnvironmentProvider for FakeEnvironmentProvider {
     async fn list(&self) -> Result<Vec<Arc<dyn ProvisionedEnvironment>>, String> {
         Ok(vec![])
     }
+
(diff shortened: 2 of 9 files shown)


rjwittams:
Reviewed. The notes are non-blocking and require no code change: the finalizer deduplication intentionally follows the established checkout-reconciler convention, while identity-only destroy intentionally delegates absent-container reporting to the provider so teardown never needs full actuated metadata parsing.

PR #1312, opened 2026-08-01
fix: quarantine undecodable replay events

### LINKED ISSUE #1309: One undecodable historical event permanently wedges peer replication
## Observed (2026-08-01)

After the #1291 schema change (`CredentialConsumer::Forgejo` gained required `username`), kiwi's event log held two ADDED events for the same credential: event 1 from an early old-schema apply (no `username`, fully superseded), event 2 current-schema. Feta's replicator, replaying the stream, hit event 1 and the **entire resource-HTTP connection failed** (`decode stored object: missing field 'username'`), retrying every 30s forever. Downstream, every kiwi-originated delete/teardown stopped propagating — abandoned/deleted convoys' containers and checkout directories sat unreaped on feta, which is how it surfaced.

Remediated surgically: snapshot + `DELETE` of the superseded event row on the origin; replication resumed immediately.

## The gap

#1238's decode quarantine covers stored **objects**; the **event-replay** path in the resource HTTP server/replicator instead fails the whole connection on one undecodable event. A single poison event is a permanent, silent partition of everything behind it in the stream — strictly worse than the object case because it also blocks unrelated resources' events.

## What should happen

1. **Quarantine at event granularity**: an undecodable event is skipped-and-recorded (kind, name, event_version, error) exactly like object quarantine — the stream continues. Superseded events (a later full ADDED/replace for the same resource) are safe to skip by construction; a quarantined *latest* event degrades that one resource, not the mesh.
2. Surface quarantined events in fleet diagnosis (same surface as object quarantine).
3. Compaction consideration: superseded events with undecodable bodies are prime compaction candidates — compacting them out retires the poison permanently.

## Store-reset rule amendment (operational, for the deploy ritual)

The pre-bump schema check must consider **event history**, not just live objects: "no old-schema objects exist" is insufficient when the kind has *any* events written under the old schema — those replay to peers. Until this issue lands, any no-compat change to a persisted kind's required fields needs either an event-log check (`SELECT ... FROM resource_events WHERE kind=... AND body lacks <field>`) or a store-reset bump.

## Acceptance

- [ ] Liveness sequence: write old-schema event, upgrade schema, replicate — stream completes, event quarantined and visible
- [ ] A poison event never blocks events for other resources
- [ ] Compaction removes superseded quarantined events

rjwittams:
Closes #1309.

## What changed

- Decode historical SQLite watch events at replay time and skip individual events that no longer match the current resource schema.
- Persist event quarantine diagnostics with kind, namespace, name, event version, error, body, and timestamp.
- Surface event quarantines through the existing fleet `ResourceStore/DecodeQuarantine` condition.
- Remove event quarantine records when retention compaction removes their corresponding historical rows.

## Root cause and impact

Stored event JSON was parsed during replay, but typed resource decoding happened later as each watch item was consumed. One old-schema item therefore reached the resource HTTP handler as an error, closed the entire watch connection, and permanently blocked every later event in that resource stream. Typed decoding now happens at the SQLite replay boundary, where one bad row can be recorded and omitted without terminating the stream.

## Validation

Regression coverage includes Unix-socket HTTP replication from a pre-upgrade cursor, an unrelated resource behind the poison event, fleet diagnosis visibility, and quarantine retirement through compaction.

### DIFF
--- a/crates/flotilla-daemon/src/runtime.rs
+++ b/crates/flotilla-daemon/src/runtime.rs
@@ -1224,26 +1224,60 @@ async fn resource_replication_content_condition(daemon: &Arc<InProcessDaemon>, n
 }
 
 fn resource_decode_quarantine_condition(diagnostics: Option<&flotilla_resources::ResourceStoreDiagnostics>) -> Option<HostCondition> {
-    let quarantines = &diagnostics?.decode_quarantines;
-    if quarantines.is_empty() {
+    let diagnostics = diagnostics?;
+    let object_quarantines = &diagnostics.decode_quarantines;
+    let event_quarantines = &diagnostics.event_decode_quarantines;
+    if object_quarantines.is_empty() && event_quarantines.is_empty() {
         return None;
     }
-    let identities = quarantines
+    let identities = object_quarantines
         .iter()
         .map(|quarantine| format!("{}/{}: {}", quarantine.kind, quarantine.name, quarantine.error))
+        .chain(
+            event_quarantines
+                .iter()
+                .map(|quarantine| format!("{}/{}@{}: {}", quarantine.kind, quarantine.name, quarantine.event_version, quarantine.error)),
+        )
         .collect::<Vec<_>>()
         .join("; ");
+    let (reason, message) = if event_quarantines.is_empty() {
+        (
+            "StoredObjectDecodeFailed",
+            format!(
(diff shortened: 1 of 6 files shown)


rjwittams:
Verified. A quarantined latest event is intentionally omitted without terminating the watch, so only that resource can remain stale; later events for the stream continue to flow. This matches #1309's explicit degradation boundary. No follow-up code change is needed for this PR.

PR #1313, opened 2026-08-01
fix: keep contained daemon sockets live across restarts

### LINKED ISSUE #1308: Contained environments lose the daemon socket across host daemon restarts
## Observed (2026-08-01, smoke r5, after the six-fix fleet bump)

After the host daemon bounce, the r5 crew's interior `flotilla crew complete` failed with:

```
spawn lock failed: lock open: Permission denied (os error 13)   (src/main.rs:794)
```

Decoded: the container bind-mounts the host daemon **socket file** (`FLOTILLA_DAEMON_SOCKET`). A Unix socket bind mount pins the *inode*; when the host daemon restarts it binds a fresh socket at the same path, and the container keeps the dead one. The interior CLI's connection fails, `connect_or_spawn` falls back to spawning a daemon inside the container, and the spawn-lock open fails on the read-only delivery path — a triply-wrong outcome (dead socket, wrong fallback, misleading error).

So every host daemon restart silently severs the control plane of all contained environments on that host. With the current bump cadence (twice today) that is a routine event, not an edge case.

## Fix directions

1. **Mount the socket's parent directory, not the socket file** — a directory bind mount sees the new socket file appear; the interior CLI reconnects to the same path transparently. This is the standard pattern for exactly this reason.
2. The interior CLI should never fall back to `connect_or_spawn` inside a delivered environment: if `FLOTILLA_DAEMON_SOCKET` is set, a failed connection is a hard, clearly-worded error ("host daemon socket stale/unreachable"), not a local-spawn attempt.
3. Consider a daemon-restart reconcile pass over provisioned environments to verify tool-delivery health (the environment records exist; a preflight `flotilla --version`-style probe through the mount would catch staleness proactively).

Related: the bind-mounted *binaries* have the same inode-pinning property (a rename-dance host upgrade leaves containers on old binaries) — acceptable for code (per docs/crew-image.md the durable fix is baked images per release, #917), fatal for sockets. The socket case is the one that must be fixed at the mount layer.

## Acceptance

- [ ] A contained crew survives a host daemon restart: interior CLI verbs work without reprovisioning
- [ ] Interior CLI with `FLOTILLA_DAEMON_SOCKET` set never attempts a local daemon spawn
- [ ] Liveness sequence: provision environment → restart host daemon → interior `flotilla crew list` succeeds

rjwittams:
Closes #1308

## Summary

- lower Unix socket tool assets to parent-directory bind mounts so socket inode replacement remains visible inside contained environments
- expose the mounted socket under a dedicated `/run/flotilla-daemon` directory while preserving custom socket filenames
- require an existing host daemon whenever `FLOTILLA_DAEMON_SOCKET` is set, with no spawn-lock or local-daemon fallback
- document the contained socket lifecycle and add regressions for mount lowering and connect-only behavior

## Testing

### DIFF
--- a/crates/flotilla-client/src/lib.rs
+++ b/crates/flotilla-client/src/lib.rs
@@ -404,7 +404,7 @@ async fn connect_required_host_daemon_with_optional_surface(
 ) -> Result<Arc<SocketDaemon>, String> {
     connect_existing_stateful(socket_path, surface.as_ref()).await?.ok_or_else(|| {
         format!(
-            "host daemon socket stale or unreachable at {}; FLOTILLA_DAEMON_SOCKET is set, so a local daemon will not be spawned",
+            "host daemon socket stale or unreachable at {}; this contained environment requires the host daemon, so a local daemon will not be spawned",
             socket_path.display()
         )
     })
--- a/crates/flotilla-client/src/lib.rs
+++ b/crates/flotilla-client/src/lib.rs
@@ -381,6 +381,35 @@ pub async fn connect_or_spawn_with_surface(
     connect_or_spawn_with_optional_surface(socket_path, config_dir, config_dir_override, socket_override, Some(surface)).await
 }
 
+/// Connect to the host daemon exposed inside a contained environment.
+///
+/// Unlike [`connect_or_spawn`], this never acquires a spawn lock, removes a
+/// stale socket, or starts a daemon. A missing listener means the host-owned
+/// control-plane mount is stale or unreachable and must be repaired outside
+/// the contained environment.
+pub async fn connect_required_host_daemon(socket_path: &Path) -> Result<Arc<SocketDaemon>, String> {
+    connect_required_host_daemon_with_optional_surface(socket_path, None).await
+}
+
+pub async fn connect_required_host_daemon_with_surface(
+    socket_path: &Path,
+    surface: SurfaceDeclaration,
+) -> Result<Arc<SocketDaemon>, String> {
+    connect_required_host_daemon_with_optional_surface(socket_path, Some(surface)).await
+}
+
+async fn connect_required_host_daemon_with_optional_surface(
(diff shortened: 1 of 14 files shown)


rjwittams:
Fixed in 4eff384d. Both the controller preflight and Docker provider now validate against the lowered parent-directory target for Unix socket assets. The regressions now mount directly onto `/run/flotilla-daemon` and verify that the collision is rejected before provider or Docker invocation.

rjwittams:
Fixed in 32195ae2. Contained tool delivery now sets a dedicated `FLOTILLA_CONTAINED_HOST_DAEMON` marker, and only that marker selects connect-only behavior; ordinary host-managed terminals continue to self-heal through `connect_or_spawn`. The default socket also moved to `${config_dir}/run/flotilla.sock`, so the parent-directory bind exposes a socket-only runtime directory rather than the config root. Added coverage for the dedicated default directory and marker propagation, then reran rustfmt, workspace clippy with warnings denied, and the full workspace suite.

rjwittams:
Addressed in 52ca56f6. Restored explicit reserved-file collision coverage alongside the socket-directory regressions in both the controller preflight and Docker provider layers. The implementation is unchanged; focused tests, pinned formatting, and workspace clippy with warnings denied are green.

rjwittams:
Fixed in 20064269. The diagnostic now explains that the contained environment requires the host daemon without incorrectly naming `FLOTILLA_DAEMON_SOCKET` as the policy switch. The focused connect-only regression and pinned formatting pass.

rjwittams:
Acknowledged. The PR is green and the remaining notes are non-blocking. I recorded the duplicated contained daemon-access construction and missing direct `CreateOpts` assertions as follow-up #1314. The `file_name()` expectation is safe for current callers because the daemon must already have successfully bound the configured socket file before contained provisioning can use that path.

PR #1317, opened 2026-08-01
fix: expose worktree git metadata to contained vessels

### LINKED ISSUE #1316: Contained checkouts: git worktree gitdir points at host-only path, breaking git inside the environment
Surfaced by smoke r6's PASS note (#1246): repository content is readable inside the container, but `git status` fails because the checkout is a **git worktree** whose `.git` file references the host's absolute common-dir path (`/home/robert/dev/flotilla/.git/worktrees/…`) — a path that does not exist inside the environment.

Crews doing real work under containment need functioning git (status, diff, commit, push are the delivery contract). Options, roughly in order of preference:

1. **Mount the worktree's common dir** at the same absolute path inside the container (simple, but leaks the whole host clone read-write and couples container paths to host layout).
2. **Provision contained checkouts as self-contained clones** (`git clone --reference`-style or full clone) instead of worktrees — containment gets an ordinary repo; the worktree optimization stays for host-direct crews. Costs disk/time per contained vessel.
3. Rewrite the gitdir/common-dir paths on provision to container-visible mounts (fiddly; `git repair`-class).

Decision interacts with the fork-stance remote rules (whatever is provisioned must preserve the exact remote set) and with teardown (self-contained clones make the abandon/delete reap simpler — no host worktree registration to prune).

## Acceptance

- [ ] `git status`, `diff`, `commit`, and `push` work inside a contained vessel's checkout
- [ ] Fork-stance remote invariants preserved by whatever provisioning shape is chosen
- [ ] Host `git worktree list` stays consistent (no stale registrations from contained vessels)

rjwittams:
## Summary

- mount each shared clone's `.git` directory into contained worktree vessels at its original absolute host path
- preserve the shared clone's remotes, refs, objects, and host worktree registrations
- cover single- and multi-repository contained mount plans

## Verification

Closes #1316

### DIFF
--- a/crates/flotilla-controllers/src/reconcilers/vessel.rs
+++ b/crates/flotilla-controllers/src/reconcilers/vessel.rs
@@ -684,6 +684,41 @@ impl Reconciler for VesselReconciler {
         } else {
             shared_clone_root.clone().unwrap_or_default()
         };
+        let mut worktree_git_common_dir_mounts = Vec::new();
+        if matches!(&strategy, PlacementStrategy::DockerWorktreeOnHostAndMount { .. }) {
+            // A linked worktree's `.git` file names its administrative directory
+            // beneath the shared clone with an absolute host path. Expose the
+            // clone metadata at that same path so Git can follow the pointer from
+            // the differently-mounted workspace inside the container. Keeping the
+            // shared clone authoritative also preserves its remote configuration
+            // and the host's worktree registrations.
+            let mut mounted_common_dirs = BTreeSet::new();
+            for checkout_name in checkout_refs.values() {
+                let checkout = self.checkouts.get(checkout_name).await?;
+                let CheckoutSpec::Worktree(checkout_spec) = &checkout.spec else {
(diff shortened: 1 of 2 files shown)


rjwittams:
Addressed all three points in 7e599fd1. Mount planning now reuses checkout data from the first pass and resolves clone metadata only in the environment-creation branch; missing clones map to an explicit vessel failure rather than bubbling a resource error. Added regression tests for both missing clone metadata and adopted/observed checkouts under the contained worktree policy, documenting that the latter is unsupported because it has no managed shared clone to mount.

rjwittams:
Evaluated; no further code change is needed. The re-review confirms all prior findings are resolved and identifies no blockers. The mount-overlap note is guarded by the current construction: clone metadata is `{shared_clone_root}/{clone_slug}/.git`, while workspaces are rooted under `{shared_clone_root}/{convoy_checkout_slug}/...`; changing those layout helpers would be the point to revisit validation. The contained/adopted incompatibility is intentional here and now has explicit regression coverage.

PR #1318, opened 2026-08-01
fix: derive remote peer socket path

### LINKED ISSUE #1315: Peer config pins the remote daemon socket path; socket relocations silently sever peering
## Observed (2026-08-01, deploying #1313)

#1313 moved the daemon socket to `~/.config/flotilla/run/flotilla.sock` (so the socket's *parent directory* can be bind-mounted into containers). Correct fix — but `hosts.toml` on every dialing peer statically pins `daemon_socket = ".../flotilla.sock"` (the old path). After the bump, kiwi's tunnels connected to the dead old socket file on feta/udder: `peer closed before sending hello` on every redial, with **zero log lines on the accepting side** (the connection never reaches the daemon), which makes it look identical to #1295's stale-reverse-socket failure while having a completely different cause. Fixed by hand-editing `hosts.toml` and bouncing.

This is the third distinct cause producing the same opaque symptom (stale reverse socket #1295, VT-dead cleat capability refusal, now config skew) — each cost a from-scratch diagnosis.

## What should happen

1. **The remote socket path should not be dialer configuration.** `hosts.toml` should name the host (and user); the daemon socket location is the *remote daemon's* contract. Either standardize the path (derive from the remote home, as the reverse-socket path already is — "both ends derive it… so no follower-side peer configuration is needed" is the right principle, applied asymmetrically today) or have the dialer resolve it with a cheap preflight (`ssh host cat ~/.config/flotilla/run/socket-path` or similar) before spawning the tunnel.
2. **Distinguish the failure**: when the forwarded connect dies before hello, the dialer can cheaply probe whether anything is listening at the remote path (`ssh host test -S <path> && ss -xl | grep <path>`-class check on repeated failure) and say "remote daemon not listening at configured path" instead of the generic hello error.
3. When a socket path changes again, the bump is not complete until every dialer's resolution agrees — if the path stays configured anywhere, that's a documented migration step in the PR.

## Acceptance

- [ ] A fresh checkout's peer config contains no daemon socket paths (or they are derived/probed)
- [ ] Moving the socket path on one host does not permanently sever peering after both sides restart
- [ ] Repeated pre-hello failures name the probable cause distinctly from #1295's class

rjwittams:
Closes #1315

## Summary

- remove daemon socket paths from `hosts.toml` peer configuration
- publish each daemon’s actual socket path and resolve it over SSH on every tunnel attempt
- probe the remote listener after a pre-hello close and report when the daemon is not listening at the derived path
- let fleet replica snapshots use the remote CLI’s own socket resolution

## Verification

### DIFF
--- a/crates/flotilla-core/src/config.rs
+++ b/crates/flotilla-core/src/config.rs
@@ -292,7 +292,6 @@ pub struct RemoteHostConfig {
     #[serde(default)]
     pub expected_node_id: Option<NodeId>,
     pub user: Option<String>,
-    pub daemon_socket: String,
     pub ssh_multiplex: Option<bool>,
 }
 
@@ -311,7 +310,6 @@ struct RawRemoteHostConfig {
     #[serde(default)]
     expected_node_id: Option<NodeId>,
     user: Option<String>,
-    daemon_socket: String,
     ssh_multiplex: Option<bool>,
 }
 
@@ -332,7 +330,6 @@ impl<'de> Deserialize<'de> for HostsConfig {
                     expected_host_name,
                     expected_node_id: host.expected_node_id,
                     user: host.user,
-                    daemon_socket: host.daemon_socket,
                     ssh_multiplex: host.ssh_multiplex,
                 })
             })
--- a/crates/flotilla-core/src/config/tests.rs
+++ b/crates/flotilla-core/src/config/tests.rs
@@ -489,12 +489,10 @@ hostname = "desktop.local"
 expected_host_name = "desktop"
 expected_node_id = "1b4d1d6b-f7b5-4c1c-8f61-6f2d8e4c91ab"
 user = "robert"
-daemon_socket = "/run/user/1000/flotilla/daemon.sock"
 
 [hosts.cloud]
 hostname = "10.0.1.50"
 expected_host_name = "cloud"
-daemon_socket = "/home/robert/.config/flotilla/daemon.sock"
 "#;
     let config: HostsConfig = toml::from_str(toml).unwrap();
     assert_eq!(config.hosts.len(), 2);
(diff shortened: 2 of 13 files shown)


rjwittams:
Fixed the configuration override regression. The daemon now publishes its actual socket path at a stable `$HOME/.config/flotilla/run/socket-path` contract independent of `FLOTILLA_ROOT`, `XDG_CONFIG_HOME`, `--config-dir`, and the socket override; a regression test keeps the discovery, config, and socket paths deliberately distinct. I also made publication atomic with a temporary file plus rename and replaced the duplicated pre-hello string with a shared constant.

rjwittams:
Kept the discovery round trip intentionally: resolving the remote daemon-owned contract on each attempt is what lets socket moves heal without follower-side configuration. I did address the timeout tuning ambiguity by splitting `REMOTE_COMMAND_TIMEOUT` from `FORWARDED_SOCKET_TIMEOUT`, then reran the exact format, clippy, and full workspace test gates.

rjwittams:
Fixed in 74e05e62. Remote discovery now emits and parses a tagged socket-path line, so login banners and MOTD output are ignored; the SSH command tests cover output before and after the tagged line. Remote command invocations also use `-T` and `BatchMode=yes`. I additionally centralized the discovery relative path and documented why its stable location deliberately ignores config-directory overrides.

rjwittams:
Acknowledged. A daemon socket directory relocation can leave the old deterministic reverse-socket inode orphaned, but it does not affect reconnection because each attempt discovers and forwards to the new directory. I’m keeping historical-directory garbage collection out of this narrowly scoped fix; the current PR handles the functional socket-move and diagnostic behavior requested by #1315.

Issue #1283, opened 2026-07-30
Prototype: workflow definition syntax — glanceability candidates over the ruled semantics


rjwittams:
Part of #1262

## Question

The ruled model is semantically right but ~30 lines of when/subject/guards/expecting YAML per rule — one-sentence rules should read like one sentence (Robert: "we want glanceability here, though ofc we can have derived views"). Produce the same workflows from #1269 in candidate surfaces, for reaction:

1. **Terse rule strings inside YAML** — e.g. `rule: "Landing & change-request.went-conflicting -> coder: rebase-brief until change-request.went-mergeable"` with the verbose form as the parsed normal form;
2. **KDL** — portfolio precedent (andamento config); designed for glanceable nested config;
3. **Status quo + derived views only** — store form stays verbose, glanceability lives in `flotilla workflow show` renderings (the null hypothesis, argued honestly).

Constraints: identical semantics to the ruled model (no expressiveness smuggled in via syntax); whatever surface wins must round-trip to the resource form (the store shape is the contract, per the manifest loop); include the rebase-on-conflict one-liner and the full implement-review loop in every candidate so both simple and looping cases are judged. HITL at resolution: closes on reaction to the artifacts.

rjwittams:
**Resolved (2026-07-31).** The eight-round HITL exploration falsified the ticket's premise — the ruled-model YAML was not the substrate to make glanceable. Resolution recorded in `prototypes/1283-workflow-substrate.md` (PR #1285, landed): **the substrate is the verb surface plus the turn log; workflow definitions are any deterministic driver of the verbs (script, rules function, restricted agent, human at the CLI), all resuming by reading the log.** Glanceability's answer: restraint lives in the capability surface, not a grammar — bash reads like bash. Robert adopted the four ADR-carry candidates as written (2026-07-31): #1265 dotted transitions → condition leaves + named definitions; #1268 'no imperative scripts' → 'no durable hidden control state competing with the machine'; #1266 rules-only survives structurally (condition table = the one surface; rows also written by wait parkings and wake registrations); #1263 Temporal-signals-middle → store-is-the-durability. Build order 0–4 stands; #1270 (exit) reconciles the reframe with the surviving grill rulings into the ADR cut.

Issue #1263, opened 2026-07-29
Research: workflow-model prior art for crewed, conversational steps


rjwittams:
Part of #1262

## Question

Survey the workflow-orchestration prior art with one lens: **what fits steps that are conversations with agent crews, not containers?** Primary sources: Argo Workflows (DAG/steps, artifacts, exit handlers) and ArgoCD (sync waves/hooks — the "argocd-ish" that was mooted), GitHub Actions (`on:` trigger model, reusable workflows, environments/approvals as HITL gates), Temporal (durable execution, signals — the closest thing to "wake a long-lived thing with an event"), AWS Step Functions (state-machine-as-data, callbacks). For each: how steps are declared, how external events re-enter a running workflow, how human gates work, what breaks when a "step" is a stateful crew session rather than an idempotent pod. Map onto our existing WorkflowTemplateSpec (vessels, stances, crew roles) — extension or replacement? Deliverable: docs/superpowers/research/ file, comparison table + a recommended shape sketch for G-tickets to grill against.

rjwittams:
Scope addition (Robert flagged https://golem.cloud/): include **Golem** in the survey as the durable-execution family member that enforces determinism at the WASM boundary (op-log + replay) instead of via SDK discipline (Temporal). Decipher its claims against primary docs (oplog/replay mechanics, what 'exactly once' actually guarantees at external boundaries, the suspend/rehydrate model), and treat it as the strongest statement of the durable-imperative paradigm vs our level-triggered reconciliation — the #1268 grill needs that contrast stated crisply. Also note (secondary) its agent-sandbox/capability/tool-middleware layer as prior art adjacent to our vessel containment + credential leases, and agents-as-MCP-servers as a distribution idea.

rjwittams:
**Resolved.** Research merged as `docs/superpowers/research/2026-07-29-workflow-model-prior-art-for-crewed-steps.md` (PR #1272). Recommended shape: **extend WorkflowTemplateSpec, not replace it** — 'Argo CD outside, Temporal Signals in the middle, Golem-like durable actors inside': level-triggered reconcilers evaluating declared state (outside), durable addressed event delivery creating idempotent crew *engagements* (middle), vessel/role retaining conversation+workspace identity across engagements (inside — today's warm terminal, tomorrow's suspend-to-zero). The #1268 grill resolves against this.

PR #1319, opened 2026-08-01
docs: adopt the workflow substrate ADR (0027)

### LINKED ISSUE #1270: Grill: ADR cut + implementation contracts (map exit)
Part of #1262

## Question

The exit ticket, now reconciling two layers: given the four resolved grills AND the adopted #1283 reframe (prototypes/1283-workflow-substrate.md — verb surface + turn log substrate, ADR-carry table, build order 0–4), decide the ADR cut and write the ADR(s): likely one substrate ADR (verbs, turns/claims/dispositions, observations with three-valued honesty, postdating, suspend ladder, condition leaves + named definitions) with the surviving grill rulings folded in (pinning-at-admission, episode identity/ladder, machine authority, condition-table-as-one-surface) and the amended rulings recorded as amendments, not silent replacements. Then produce the dispatchable implementation contracts following the exploration's build order: step 0 (unified reads — already filed) → Bosun-over-verbs experiment → flotilla wait (block-mode, full-ladder contract) → extraction round. The three proving scenarios and the inside-out shepherd are acceptance cases across the chain. The map closes when the ADR(s) and contracts exist.

rjwittams:
Closes #1270's ADR half. Compresses the adopted #1283/#1285 exploration into ADR 0027, with the four prior rulings (#1263/#1265/#1266/#1268) recorded as amendments rather than silently replaced, and the surviving rulings (pinning-at-admission, machine authority, episode identity/ladder, one condition table, no second bus) folded in.

Grilled with Robert before drafting; the two rulings from that session are reflected:
- **Turns are not a new concept** — a turn is one round of an agent loop as it already works (brief in, completion out); the record is what we already keep (`crew_work`, event history, transcripts). No Turn resource kind.
- **Schema footprint is exactly two additions**: brief-declared completion dispositions (one word from a declared list; free-form stays legal) and `--step` keys on mutating verbs. Richer completion cargo is deferred to a future *artifact* concept, explicitly out of scope.

Also fixes the ADR numbering collision: two ADRs shared 0025; the later-dated fleet-releases ADR is renumbered to 0026 (no references existed outside the directory).

Remaining #1270 work (implementation contracts for Bosun-over-verbs and `flotilla wait`) follows as separate tickets per the build order.


rjwittams:
Follow-up commit for the density observation (seconded by Robert as a real trend worth countering): the four heaviest terms now carry plain-English glosses at first use in the ADR itself — level-triggered, condition leaf, postdating ("never judge work by evidence older than the work"), episode identity ("one firing, one engagement; a re-fire is a new escalated round") — and the durable vocabulary (Brief, Disposition, Observation, Condition Leaf) is promoted into CONTEXT.md's glossary with the standard _Avoid_ lines, so future documents can reference rather than re-explain. The working rule going forward: a term not worth a glossary entry is not worth coining.

rjwittams:
Good catch — fixed. The two entries are reconciled into the original: **Brief** keeps its durable-file framing and the `_Avoid_: Prompt` line (which my duplicate directly violated), gaining only what the ADR needs — each engagement round is started by a brief (the existing "accumulated briefs are the script as executed" already carried the per-round sense), and a brief may declare the valid Dispositions. ADR 0027's wording is corrected to match: the brief is the statement of what the round is for; the prompt is only its delivery mechanism. One bolded **Brief** entry remains in the glossary.

rjwittams:
Resolved by using the word we already ruled: the turn-ending report is ADR 0017's **settlement claim**, and this PR's drift into lowercase "completion" was exactly that — drift. Applied throughout: the glossary gains a **Settlement Claim** entry (with `_Avoid_: Completion`), **Disposition** and **Condition Leaf** now reference claims, the admission-time **Completion** entry gains a disambiguating `_Avoid_` cross-reference, and ADR 0027's prose says claim everywhere (the `crew complete` verb keeps its name; verbs and nouns needn't match). Pleasant side effect: the three primitives are now visibly ADR 0017's title — claims, conditions, observations — generalized from convoy settlement to all workflow, plus the machine. ADR 0017 added to Relates-to accordingly.

rjwittams:
Fourth catch, same fix rule: "engagement" and "turn" were the same unit seen from two sides — the machinery that starts it and the agent living it — so the plain word wins everywhere. The ADR's machinery prose now says turn (one open turn per condition source/convoy/vessel/role; resume is a new turn; Interrupted ends the turn), the Brief entry says "each turn is started by a brief", and "engagement" survives only as a verb ("a rule that engages crew") describing condition-table rows, which needs no noun coinage. Zero occurrences of the noun remain in either document.

PR #1333, opened 2026-08-02
fix: cascade convoy checkout cleanup

### LINKED ISSUE #1332: Convoy deletion orphans Checkout resources: cleanup finalizer never fires, checkouts accumulate unboundedly
## What happened

After a normal day of dispatches (2026-08-02), feta had **22 orphaned Checkout resources** and ~230G of checkout directories under `~/dev/flotilla-repos/` — every one belonging to a convoy that had already been deleted (landed convoys removed with `flotilla convoy delete`, plus stragglers from earlier waves back to `docker-crew-smoke-r4`). Each Checkout resource was healthy-looking: no `deletionTimestamp`, finalizer `flotilla.work/checkout-cleanup` present, path pointing at a live directory.

`convoy delete` tears down environments/terminals but does **not** cascade to the convoy's Checkout resources. Since nothing ever deletes them, the cleanup finalizer never fires, and the disk grows forever (individual checkouts reach 30-45G once a crew has built the workspace). This is the third leak corridor in this family, after #1297 (abandon leaks, fixed by #1300/#1307 for the abandon path) — the delete-after-landed path has the same gap.

Manual remediation performed today: `flotilla resource delete checkout <name>` on the authority host fired the finalizers correctly and reclaimed the space — so the cleanup machinery itself works; it is only the cascade that is missing.

## What to do

- Make convoy teardown (the `flotilla.work/convoy-teardown` finalizer path) delete the convoy's Checkout resources, so their own cleanup finalizers run. Ownership metadata should make membership unambiguous — if Checkouts don't currently carry an owner reference to their convoy, add one at admission rather than matching by name convention.
- Cover both exits: `convoy delete` of a Landed convoy and `convoy abandon` (verify #1307's path also removes the *resource*, not just the directory).
- A checkout whose directory is already gone must delete cleanly (no wedged finalizer) — same decoupling principle as #1311.
- Behavior test through injected collaborators per the testing philosophy: convoy with a checkout → delete convoy → checkout resource gone and cleanup invoked; plus the already-deleted-directory case.

## Acceptance criteria

- [ ] Deleting a Landed convoy removes its Checkout resources and their directories
- [ ] Abandon path verified to do the same
- [ ] Missing-directory checkouts delete without wedging
- [ ] Tests cover the cascade at the reconciler level (no real filesystem/subprocess needed)

rjwittams:
Fixes #1332.

Crew-authored fix (convoy checkout-cascade) salvaged by the governor: the implementation completed in the vessel as commit 536edb88, but delivery failed — the contained vessel reported `gh` unauthenticated at push time (feta suspended mid-run; possible credential-lease expiry after resume, under investigation separately). Branch pushed from the host; the commit is the crew's, unmodified.

Makes convoy teardown cascade to the convoy's Checkout resources so their `checkout-cleanup` finalizers actually fire, covering both the delete-after-landed and abandon exits, with missing-directory checkouts deleting cleanly.

### DIFF
--- a/crates/flotilla-controllers/src/reconcilers/checkout.rs
+++ b/crates/flotilla-controllers/src/reconcilers/checkout.rs
@@ -15,6 +15,7 @@ const CHECKOUT_PROVISIONING_REQUEUE_AFTER: Duration = Duration::from_secs(1);
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum CheckoutRemoval {
     Worktree { clone_path: String, branch: String, target_path: String },
+    OrphanedWorktree { target_path: String },
     FreshClone { target_path: String },
 }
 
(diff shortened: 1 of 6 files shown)


rjwittams:
Review round processed: the one non-blocking note (vessel teardown's stale independent-cleanup claim) is filed as a follow-up. All six checks green on the only commit; the review independently confirmed the root cause (vessel-owned vs convoy-owned checkout meta) and that the missing-directory criterion falls out of the idempotent removal. Merge-ready.

PR #1336, opened 2026-08-02
fix: key sleep inhibition on local placement

### LINKED ISSUE #1334: Sleep inhibition keys on convoy authority, not placement: feta suspended under two active crews
## What happened

2026-08-02 ~17:00: feta suspended while hosting two Active convoys (checkout-cascade, contained; cleat-daemon-autostart, trusted). The cleat crew's shepherd loop stalled mid-review; the checkout-cascade crew failed delivery after resume. feta's fleet-health row showed `sleep: not_required` throughout, while kiwi showed `held` — for a convoy placed on kiwi, but kiwi also holds for convoys it dispatched onto other hosts.

## Root cause

`sleep_inhibitor.rs` builds `ActiveConvoys` from `convoys.list()` + `watch` — **local-authority resources only**. Convoy resources live with the dispatching daemon (kiwi), so:

- feta's inhibitor never sees convoys placed on feta by kiwi → `required() == false` → machine sleeps under active vessels.
- kiwi inhibits for any non-terminal convoy it authored, including ones whose vessels run entirely elsewhere → kiwi stays awake for work it isn't doing.

Inhibition keys on the wrong side of the placement relationship.

## What to do

Key inhibition on **work placed on this host**. Two candidate shapes — pick with reasoning in the PR:

1. Placement-filtered convoys: watch convoys including replicas and require `placement_decision.target_host == self` (plus non-terminal phase).
2. Locally-authored work resources: key on this host's own Environments/TerminalSessions belonging to live convoys — no replica dependency, and it tracks what is *physically* running here (vessels enable work; the host with the vessel stays awake).

Either way: a host with no locally-placed work must not inhibit just because it dispatched something, and a host with locally-placed work must inhibit even when the convoy record is remote. Behavior tests through the existing injected-inhibitor harness covering both directions of the placement split, plus replication-lag honesty (an Unknown/lagging replica should fail toward holding inhibition, not toward sleeping — sleeping under active work is the expensive error).

## Acceptance criteria

- [ ] Host with a remotely-dispatched, locally-placed active convoy holds inhibition
- [ ] Host that dispatched a convoy placed elsewhere does not hold for it
- [ ] Terminal phases release as today
- [ ] Uncertainty (replica lag / watch gap) errs toward holding
- [ ] Injected-inhibitor behavior tests for all four cases

rjwittams:
Closes #1334.

## What changed

- inhibit sleep for the union of locally-authored vessel resources and non-terminal convoys placed on this host
- read convoy replicas so remotely dispatched local placement is visible
- acquire defensively during startup and on watch failure, and close the overlay list/watch gap by establishing the watch before listing
- add injected-inhibitor behavior coverage for both placement directions, terminal release, unknown placement, and vessel-held teardown lag

## Why

Convoys are authoritative on the dispatching host, so a local-only convoy watch caused placement hosts to sleep under active work while dispatch hosts stayed awake for work placed elsewhere. Per the governor ruling on #1334, local vessels are the primary physical-work signal; placement-filtered convoys cover the admission window before a vessel exists.

## Validation

- focused sleep inhibitor tests: 11 passed
- workspace suite: affected tests pass; 1,219 `flotilla-core` tests passed, with one environment-only failure because the container has no readable machine ID (`machine_scoped_dir_falls_back_to_etc_machine_id`)


rjwittams:
Addressed the review's dynamic coverage suggestion in `3745b4f5`: `unknown_placement_holds_until_replica_catches_up` now drives `placement_decision` from unknown to `other-host` through the live watch and asserts inhibition releases. The other noted items are intentional fail-toward-holding tradeoffs required by #1334's ruling.

PR #1338, opened 2026-08-02
refactor: leave checkout cleanup to convoy teardown

### LINKED ISSUE #1337: Vessel teardown no longer cleans checkouts: remove the dead path / stale claim
From PR #1333 review: with all placement strategies now creating convoy-owned checkouts (CONVOY_LABEL + Convoy owner ref), checkout cleanup is fully delegated to the convoy-level cascade. The vessel teardown path still suggests (in code/comments) that vessel teardown independently cleans its checkouts — no longer true for any strategy. Remove the dead path or correct the claim with a comment pointing at the convoy cascade, so a future reader doesn't rely on the vessel finalizer for checkout cleanup.

Not a functional bug — latent confusion/dead code only.

rjwittams:
Closes #1337

Removes the stale vessel-finalizer checkout deletion path now that every managed checkout is convoy-owned. The finalizer comment and lifecycle tests now make the ownership boundary explicit: vessel teardown cleans terminal sessions and environments, while the convoy cascade cleans checkouts.

## Verification

### DIFF
--- a/crates/flotilla-controllers/src/reconcilers/vessel.rs
+++ b/crates/flotilla-controllers/src/reconcilers/vessel.rs
@@ -1051,7 +1051,8 @@ impl Reconciler for VesselReconciler {
         let selector = BTreeMap::from([(VESSEL_REF_LABEL.to_string(), obj.metadata.name.clone())]);
 
         delete_lifecycle_owned_matching(&self.terminal_sessions, &selector).await?;
-        delete_lifecycle_owned_matching(&self.checkouts, &selector).await?;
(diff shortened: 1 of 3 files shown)


rjwittams:
The non-blocking secondary-watch observation is valid and separate from the vessel-finalizer cleanup in this PR. I filed #1339 to verify adopted-checkout implications and remove the stale vessel-label watch if no lifecycle depends on it.

rjwittams:
Handled as follow-up #1339 after verifying managed checkout creation only uses convoy ownership. The follow-up explicitly calls out checking adopted-checkout lifecycle behavior before removing the watch.

Issue #1342, opened 2026-08-03
checkout integration no-CR early return keys on the checkout's spec ref — a PR opened from a different branch is invisible to settlement


rjwittams:
## What to fix

`inspect_checkout_integration` declares `landed=True` when the checkout's HEAD has no commits beyond `base_ref`, as an early return **before any forge lookup** (`crates/flotilla-core/src/checkout_integration.rs:249-266`), keyed on the checkout's own spec ref (`checkout_branch_from_spec`, `:35-41`). A change request opened from a different head/branch than the checkout's spec ref — e.g. a crew that created its own branch inside the checkout and pushed from it — is invisible to the Landing→Landed condition: the spec ref sits at base, the early return fires, and the convoy settles under an open PR.

Found while tracing #1341 (grill #1340); this is a second, narrower hole in the same condition.

## Fix

The zero-commits-beyond-base check alone must not be sufficient for `landed=True`. Only return `True` on the no-CR path when the forge has also been consulted and reports no open change request associated with the convoy (not merely with the spec ref). If the forge cannot be consulted, the result is `Unknown` (holds Landing), consistent with the existing unknown/unprobeable handling (`checkout_integration.rs:407-419`).

## Acceptance criteria

- [ ] A checkout whose spec ref is at base but whose convoy has an open CR from another branch holds `Landing`, with a replay-fixture or injected-collaborator test
- [ ] No-CR-anywhere checkout with spec ref at base still reaches `Landed`
- [ ] Forge-unreachable ⇒ `Unknown`, unchanged

Relates: #1341, ADR 0021, #1340.

Issue #1341, opened 2026-08-03
Landing→Landed condition and teardown gate are vacuously true on empty checkout lists; daemon impl discards the federated checkout list


rjwittams:
## What happened

2026-08-02: two convoys (vessel-teardown-claim → PR #1338, ssh-transport-seam → PR #1329) reached `Landed` and vessel teardown at the moment of the crew's settlement claim while their PRs were still open. ADR 0021 forbids exactly this: the claim writes only `Active → Landing`, and only the lifecycle reconciler's condition ("no change request remains outstanding") may write `Landed`. Traced 2026-08-03 (grill #1340): the claim edge, the reconciler, and the teardown keying are all ADR-compliant — the divergence is inside the condition's evaluation.

## Root cause

Two compounding defects, same shape — **vacuous truth over an empty checkout list**:

1. **The federated checkout list is discarded.** The convoy reconciler resolves the convoy's checkouts through `including_replicas`/`federated_children` (`crates/flotilla-resources/src/convoy/reconcile.rs:254-257`, `:187-215`) and hands them to `ConvoyTeardownRuntime::no_change_request_outstanding` — whose daemon impl (`crates/flotilla-daemon/src/runtime.rs:87-93`) ignores the `_checkouts` argument and re-lists from the **local, non-federated** backend (`crates/flotilla-core/src/in_process.rs:6127`). A convoy whose authority record and vessel/checkouts sit on different hosts sees an empty local list on the controlling daemon.
2. **Empty ⇒ settled.** The evaluation iterates `for checkout in checkout_list { … }` and falls through to `Ok(true)` (`in_process.rs:6133-6180`); the trait default (`reconcile.rs:44-46`, `.all()` over an empty slice) and the no-runtime fallback (`reconcile.rs:274-276`) share the shape. `Adopted` checkouts are skipped (`in_process.rs:6134-6136`), so an adopted-PR convoy's list can be effectively empty too. The teardown gate `verify_convoy_teardown_gate` (`in_process.rs:6183-6300`) walks the same empty local list, collects no refusals, returns `Ok(())` (`:6293-6297`) → `reclaim_eligible` → `DeleteVessel`.

Result: condition true with zero evidence → `Settle` → `Landed` → teardown under an open PR. Reproduces both incidents (the vessel-teardown-claim convoy was authoritative on kiwi while its vessel ran on feta).

## Fix

[...]

Issue #1340, opened 2026-08-02
Grill: split turn-exit from convoy-exit — settlement, vessel lifetime, and workflow-level observed exits


rjwittams:
## Question

In today's machinery, a sole crew member's settlement claim IS the workflow's completion condition, so claim → convoy Landed → vessel teardown are one event. The 2026-08-02 evidence (first Bosun run, #1321): the coder's claim tore down its own session while PR #1338 sat unmerged — any review round, merge, or revert after the claim arrives into a reaped vessel. Same conflation behind ssh-transport-seam settling under an open review and the stalled cleat shepherd.

Grill the split, with ADR 0027 as the frame (claims end turns; observations end convoys):

1. **Workflow-level exit conditions**: how does a workflow definition (pinned at admission) declare its exit as an observation (e.g. `cr.merged`) rather than the latest claim? What do today's single-agent workflows declare instead, and what is the migration?
2. **Vessel lifetime**: vessels live for the workflow's horizon, not the latest claim's (per the vessels-enable-work ruling). What holds a vessel open between turns — and what releases it (observed exit, abandon, explicit reap)? Interaction with sleep inhibition (#1336, vessel-union) falls out of this.
3. **Ensure-flavored brief delivery**: a new turn against reaped/idle crew re-provisions instead of failing on a dead session name. What does re-provisioning restore (checkout, credentials, agent context), and what is legitimately lost between turns?
4. **What Landed means afterwards**: if convoys stay open awaiting observations, what phase does a filed-claim-but-unmet-exit convoy sit in, and what does the fleet view show?

**Visualizations wanted** (prototype-grade, per Robert): timeline/state diagrams of today-vs-proposed — turn, claim, convoy phase, vessel lifetime, observation stream on one time axis, using the #1338 run and the ssh-transport-seam incident as the two worked examples.

Relates to: ADR 0027, #1321 (evidence source), #1322 (`flotilla wait` — its leaves are the exit-condition language), ADR 0017/0024.

rjwittams:
## Working frame (ruled 2026-08-03, provisional)

Robert accepted the three-state ladder as the experimental frame for this grill — "very hard to know if it will work out in practice, but the right shape to experiment with for now":

1. **Claimed-out** — every crew turn has exited via a settlement claim. Claims end turns, nothing more.
[...]

Issue #1322, opened 2026-08-02
flotilla wait: level-triggered which-condition select over condition leaves


rjwittams:
## Parent

ADR 0027 build-order step 2; part of wayfinder map #1262. Companion experiment: #1321 (Bosun-over-verbs).

## What to build

The synchronization keystone of the workflow substrate:

```
flotilla wait --for <leaf> [--for <leaf>]… [--fresher-than <ref>] [--timeout <s>] --json
```

- **Level-triggered**: evaluates current state before blocking; a condition that already holds returns immediately.
- **Which-condition select**: multiple `--for` leaves OR together; the result names *which* leaf fired and carries a snapshot of the evaluated state. AND-chains, counters, loops, and budgets stay in the caller's language — deliberately not in the leaf grammar.
- **Three-valued honesty**: an Unknown observation never triggers a leaf and never closes one.
- **Postdating**: `--fresher-than <ref>` refuses to fire on evidence older than the referenced record (e.g. the settlement claim a driver would otherwise judge with stale observations).
- **Ships block-mode only, but the contract is written for the full suspend ladder** (ADR 0027): plain-shell block / parked crew script (daemon condition-table row + SIGSTOP) / evicted crew script (killed, re-run from top, step keys make it sound) / agent wake (turn ends; resume is a new turn carrying the which-condition digest). Park/evict/wake must be compatible deepenings, not redesigns. Discipline stated in the contract: `wait` is a script's only blocking point against flotilla state.

Plus the two ADR-ruled schema additions, which land here because wait's leaves read them:

- **Dispositions**: a brief may declare its answer vocabulary (e.g. `merged | changes-pushed | blocked`); `crew complete` records the chosen disposition beside the free-text claim; the `latest claim disposition` leaf reads it. Briefs that declare nothing keep free-form claims.
- **Step keys**: mutating verbs accept `--step <key>` (author-named, e.g. `round-2`); re-running a driver finds the recorded result instead of re-issuing the action.

## Design work this contract owns (deferred here by ADR 0027)

Resolve these *before* implementation, as a proposal comment on this issue flagged for ruling — each has a stated default; validate against the six scenario files in `prototypes/1262-scenarios/`:

[...]

PR #1344, opened 2026-08-03
docs: adopt ADR 0028 — world-terminal exits, vessel-as-cache, leaf engine

rjwittams:
Records the #1340 grill rulings (2026-08-03) as ADR 0028, amending ADR 0021 and ADR 0027 jointly:

- **Exit tables**: `Landed` records world terminals only (`cr.merged`, `cr.closed`); budget → delivery policy, staleness → attention flag, give-up → explicit reap; `exit: claim` declarable for artifact-less work. Migration reading: the hardwired "no CR outstanding" condition is the implicit default table.
- **The vessel is a cache**: no substrate semantics may depend on vessel identity across turns; park depth is a spectrum (warm → suspended → archived logs) that changes latency, never meaning.
- **The claim is a durability fence** (flag, don't hard-reject) and **ensure-flavored delivery** walks a ladder with fresh-agent-plus-self-sufficient-brief as the never-fails floor; rung observability is first-class, rung heuristics are experiments.
- **The leaf engine**: one event-delivery mechanism, two watcher kinds (hanging `wait` callers, parked-convoy reconcilers); #1322 re-scoped to build it; resident drivers demoted to stopgap, the Bosun's judgment surviving as wakeup-delivered driver turns.
- **Vocabulary**: `Landing` keeps its name ("claimed-out" is the mechanism gloss); `Anchored` survives; stuck = Landing + escalation.

Also adds six glossary entries to CONTEXT.md (World Terminal, Exit Table, Durability Fence, Delivery Ladder, Leaf Engine, Park Depth).

Context: the 2026-08-02 teardown-at-claim incidents were traced to a machinery bug (#1341, dispatched; #1342/#1343 filed), not a design gap — ADR 0021 already rules the skeleton; this ADR layers the grill's generalizations on top. Full ruling trail: #1340.

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -509,6 +509,50 @@ are layered data, not new objects.
 _Avoid_: Rule (that is a leaf bound to a wake or crew turn), expression
 language.
 
+**World Terminal**:
+An **Observation** about a workflow's artifact that makes the work genuinely
+over (`cr.merged`, `cr.closed`) — the only legal right-hand side of an
+**Exit Table** (ADR 0028). Facts about the convoy's own effort (budget,
+staleness) are never world terminals: budget is a delivery policy, staleness
+an attention flag, give-up an explicit reap.
+_Avoid_: Completion (the admission-time stage), done.
+
+**Exit Table**:
+A workflow definition's declared exit: named **Dispositions** each bound to
+one **World Terminal** leaf; the first to fire writes `Landed` and records
+its name (ADR 0028). `exit: claim` is the declarable degenerate case for
+artifact-less work.
+_Avoid_: Success criteria, completion condition.
+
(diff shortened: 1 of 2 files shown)


rjwittams:
Good catch — fixed rather than defended. The tension was real: the budget bullet read as the substrate silently taking over 0027's caller-owned budgets. The reconciliation is in 0027 itself: its episode-identity obligation already gives the machinery an escalation ladder. The bullet now distinguishes the two cases — caller-owned budgets stand unchanged wherever a caller exists (scripts, resident drivers); wakeup-delivered turns have no caller, so their bound is the machinery-owned episode escalation ladder, surfaced as delivery refusal plus 0027's own loud `convoy hold`. Added the corresponding row to the amendments table.

Issue #1335, opened 2026-08-02
Contained crews on GitHub-backed projects have no push credentials: no gh CredentialSpec/grant exists, and the gh adapter never wires git


rjwittams:
## What to build

The `Gh` credential consumer adapter delivers only a `GH_TOKEN` env var into contained crew sessions (`crates/flotilla-daemon/src/credential.rs:288-302`). That authenticates the `gh` CLI but leaves `git push` over https with no credential path at all — the observed failure is `could not read Username for 'https://github.com'` from crews whose work was otherwise complete (#1333's crew, #1341's crew → salvaged as PR #1346).

Bring the gh adapter to parity with the Forgejo adapter, which already stages a generated `git-credential-forgejo` helper plus `GIT_CONFIG_COUNT`/`GIT_CONFIG_KEY_*`/`GIT_TERMINAL_PROMPT=0` wiring at `/run/flotilla/credentials/<name>/` inside the vessel (`credential.rs:314-352`): stage the equivalent github.com credential-helper material (helper script sourcing the token, or `gh auth setup-git`-style `credential.helper=!gh auth git-credential` config env), so that both `gh` and plain `git push`/`git fetch` against github.com work in a contained session with only the granted credential.

Keep the existing mandatory preflight, and extend it to verify the git path too (e.g. `git ls-remote` against an https URL of the granted scope), so a mis-wired helper fails provisioning loudly instead of surfacing as a dead crew claim.

## Acceptance criteria

- [ ] A contained vessel with a granted gh credential can `git push` to a github.com https remote and use `gh` — no ambient host auth involved
- [ ] `GIT_TERMINAL_PROMPT=0` (or equivalent) so a missing credential fails fast instead of hanging a session
- [ ] Preflight covers the git credential path; failure discards the environment loudly (existing `runtime.rs:1840-1845` path)
- [ ] Unit coverage through the adapter seam mirroring the existing Forgejo adapter tests

## Out of scope

Creating the GitHub CredentialSpec/grant themselves — that is host/resource configuration pending a token-provenance ruling (dedicated fine-grained PAT vs `IssueCommand: gh auth token`), tracked in the comments here. The related feta sleep-inhibitor polkit failure is tracked separately.

## Context

Full corrected diagnosis in the 2026-08-03 comment below the original (wrong) suspend hypothesis. Relates: #1291 (Forgejo adapter precedent), #1156 (credential grants/adapters), #1140.

rjwittams:
[...]

PR #1346, opened 2026-08-03
fix: require federated checkout settlement evidence

### LINKED ISSUE #1341: Landing→Landed condition and teardown gate are vacuously true on empty checkout lists; daemon impl discards the federated checkout list

rjwittams:
Fixes #1341.

Crew-authored fix (convoy `landing-vacuous-settle`, contained on feta), delivered by hand: the vessel had no working GitHub auth — second occurrence of #1335, this time in a fresh vessel created after host resume — so the branch was pushed from the host and this PR opened by the governor. Crew's claim message: implementation committed as 90bd3edc with fmt, clippy, and the focused regression suites passing.

Makes convoy settlement and the teardown gate consume the federated checkout list the reconciler already resolved, and bans vacuous truth: an empty effective checkout list no longer evaluates as settled/reclaimable.

### DIFF
--- a/crates/flotilla-controllers/tests/liveness_contract.rs
+++ b/crates/flotilla-controllers/tests/liveness_contract.rs
@@ -277,7 +277,11 @@ impl ConvoyTeardownRuntime for LandingRuntime {
         Ok(!self.contradictory)
     }
 
-    async fn verify_reclaim(&self, _convoy: &ResourceObject<Convoy>) -> Result<(), String> {
+    async fn verify_reclaim(
+        &self,
+        _convoy: &ResourceObject<Convoy>,
+        _checkouts: &[ResourceObject<flotilla_resources::Checkout>],
+    ) -> Result<(), String> {
         Err("not terminal".to_string())
     }
 }
--- a/crates/flotilla-controllers/tests/vessel_reconciler.rs
+++ b/crates/flotilla-controllers/tests/vessel_reconciler.rs
@@ -38,7 +38,11 @@ struct AlwaysEligible;
 
 #[async_trait]
 impl ConvoyTeardownRuntime for AlwaysEligible {
-    async fn verify_reclaim(&self, _convoy: &flotilla_resources::ResourceObject<Convoy>) -> Result<(), String> {
+    async fn verify_reclaim(
+        &self,
+        _convoy: &flotilla_resources::ResourceObject<Convoy>,
+        _checkouts: &[flotilla_resources::ResourceObject<flotilla_resources::Checkout>],
+    ) -> Result<(), String> {
         Ok(())
     }
 }
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -6132,18 +6132,22 @@ impl InProcessDaemon {
             .await
             .map_err(|err| err.to_string())?
             .items;
-        self.convoy_change_requests_settled_for_checkouts(&checkout_list).await
+        let convoys = self.resource_backend.clone().using::<ResourceConvoy>(namespace);
+        let convoy = convoys.get(name).await.map_err(|err| err.to_string())?;
+        self.convoy_change_requests_settled_for_checkouts(&convoy, &checkout_list).await
     }
 
     pub async fn convoy_change_requests_settled_for_checkouts(
         &self,
+        convoy: &ResourceObject<ResourceConvoy>,
         checkout_list: &[ResourceObject<ResourceCheckout>],
     ) -> Result<bool, String> {
-        let Some(first_checkout) = checkout_list.first() else {
-            return Ok(false);
-        };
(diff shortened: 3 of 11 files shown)


rjwittams:
The failing Test job exposed an important distinction that the original blanket non-empty check missed. Fixed in `65434b0f`: the convoy now derives its expected checkout names from durable work-placement `checkout_refs` plus declared adopted refs, and compares those against convoy-labeled federated observations. Artifact-less/tool-only convoys can settle on their completion claims; when any checkout is expected, every expected name must be observed and carry landed evidence, with the same presence rule enforced at the reclaim gate. Added regression coverage for expected-but-unobserved, artifact-less, and expected+observed+landed cases. The exact format, clippy, and full workspace test gates pass locally.

On the two non-blocking notes: adopted checkouts remain covered by the shared expected-set gate and their existing landed requirement. The free-standing `reconcile()` path is intentionally claim-only for artifact-less inputs; it has no checkout dependency surface, while `ConvoyReconciler` is the path that compares expected refs with observed checkout resources.

rjwittams:
Both carried-over items closed by code (the crew's session was torn down by the very bug this PR fixes — third specimen — so the governor finished the round): (1) added `adopted_checkout_with_unlanded_change_request_refuses_reclaim_gate`, the gate-side twin of the settle-condition test, asserting the refusal names the checkout and `Landed=False`; (2) the free-standing test-support `reconcile()` now derives `no_change_request_outstanding` from `expected_checkout_refs(convoy).is_ok_and(is_empty)` with a doc comment stating the artifact-less contract — the vacuous-truth door is closed even for future direct callers. Full workspace tests, pinned fmt, and clippy green locally.

PR #1349, opened 2026-08-03
fix: wire gh credentials into git

### LINKED ISSUE #1335: Contained crews on GitHub-backed projects have no push credentials: no gh CredentialSpec/grant exists, and the gh adapter never wires git

rjwittams:
Closes #1335

Configure Git credential lookup for github.com through gh auth git-credential under the granted GH_TOKEN, and disable terminal prompting in contained vessels.

Extend the mandatory credential preflight to validate both gh API authentication and the Git credential-fill path. Add adapter-seam unit coverage mirroring the Forgejo credential delivery test.

Checks:

### DIFF
--- a/crates/flotilla-daemon/src/credential.rs
+++ b/crates/flotilla-daemon/src/credential.rs
@@ -26,6 +26,7 @@ pub(crate) struct CredentialStore {
     state_dir: PathBuf,
     prepared: Mutex<BTreeSet<(String, String)>>,
     materials: Mutex<BTreeMap<(String, String), String>>,
+    git_config_fragments: Mutex<BTreeMap<String, BTreeMap<String, GitConfigFragment>>>,
     registry_configs: Mutex<BTreeMap<String, PathBuf>>,
 }
 
@@ -40,6 +41,7 @@ struct GitCredentialContribution {
     preflight: Option<GitCredentialPreflight>,
 }
 
+#[derive(Clone)]
 struct GitConfigFragment {
     target: String,
     content: String,
@@ -54,9 +56,9 @@ struct PendingGitPreflight {
     preflight: GitCredentialPreflight,
 }
 
-fn render_gitconfig(fragments: &[GitConfigFragment]) -> String {
+fn render_gitconfig<'a>(fragments: impl IntoIterator<Item = &'a GitConfigFragment>) -> String {
     fragments
-        .iter()
+        .into_iter()
         .map(|fragment| format!("[credential \"{}\"]\n\thelper = {}\n", fragment.target, fragment.content))
         .collect::<Vec<_>>()
         .join("\n")
@@ -128,6 +130,7 @@ impl CredentialStore {
             state_dir,
             prepared: Mutex::new(BTreeSet::new()),
             materials: Mutex::new(BTreeMap::new()),
+            git_config_fragments: Mutex::new(BTreeMap::new()),
             registry_configs: Mutex::new(BTreeMap::new()),
         }
     }
@@ -186,7 +189,7 @@ impl CredentialStore {
             }
         }
         let mut env = BTreeMap::new();
-        let mut git_config_fragments = Vec::new();
+        let mut new_git_config_fragments = BTreeMap::new();
         let mut git_config_owner = None;
         let mut pending_git_preflights = Vec::new();
         let mut prepared_cache_keys = Vec::new();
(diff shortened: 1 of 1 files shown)


rjwittams:
Design ruling (Robert, 2026-08-03), superseding the slot-counter direction from the earlier round: no numbered-env multiplexing — `GIT_CONFIG_COUNT`/`KEY_n`/`VALUE_n` is git's own env API but it is a flat array namespace that composes badly across independent adapters (the collision the review found is intrinsic to it). Instead each git-hosting adapter contributes a fragment to a single generated gitconfig file staged in the vessel (`/run/flotilla/credentials/gitconfig`, same move as the staged Forgejo helper), delivered via one single-valued env var `GIT_CONFIG_GLOBAL`; the Forgejo adapter's existing `GIT_CONFIG_*` env entries migrate into fragments too. `GIT_TERMINAL_PROMPT=0` remains a plain env var. Regression test: Gh + Forgejo granted together → one gitconfig, both helper stanzas, both hosts resolve. Round brief updated in the crew session.

rjwittams:
Fixed. Git-hosting adapters now contribute credentials-internal GitConfigFragment values with a target and helper content. prepare() renders all fragments once into /run/flotilla/credentials/gitconfig, delivers GIT_CONFIG_GLOBAL plus one GIT_TERMINAL_PROMPT=0, and runs each host credential-fill preflight against the final shared file. The Forgejo fixed-slot wiring was migrated to the same model.

Added a Gh + Forgejo regression proving the shared file contains both host-specific stanzas and that both preflights use GIT_CONFIG_GLOBAL. The pinned format check, workspace clippy gate, and full locked workspace test suite all pass.

rjwittams:
Fixed the reachable cross-call bug. CredentialStore now retains GitConfigFragment values per environment_ref, merges each preparation into that accumulated set under one mutex, renders the full file, and only commits the new fragment state after the shared-file write and Git preflights succeed. forget_environment removes the accumulated fragments with the material/preflight cache.

Added a regression that prepares Gh and Forgejo in two disjoint calls for the same environment and verifies the final gitconfig retains both stanzas. The existing one-call composition test remains. Format, workspace clippy, and the full locked workspace suite pass.

The host-direct collision described in the comment is not reachable for this credential delivery path: these granted credentials are contained-stance-only by design, so the shared path is inside each contained environment's filesystem rather than the host-direct runner. The per-environment accumulation covers the repeated contained-runtime calls identified in the review.

rjwittams:
Reviewed; no further code change is needed for these non-blocking observations. The candidate file must exist before git credential fill can exercise that exact GIT_CONFIG_GLOBAL content. Preparations are serialized under the fragment mutex, the in-memory map commits only after every preflight succeeds, and any later preparation re-renders the authoritative accumulated map, so a failed candidate cannot become durable bookkeeping state. The blank line is valid stanza separation in Git config, and the noted partial material-cache eviction does not change retry correctness or delivered authorization.

Thanks for independently verifying the contained-only placement invariant and the cross-call regression path.

PR #1350, opened 2026-08-03
fix: surface sleep inhibition failure as host condition

### LINKED ISSUE #1347: Sleep-inhibitor acquisition failure is silent — surface it as a host attention condition
## What to build

`flotilla-daemon/src/sleep_inhibitor.rs` warn-loops when the `systemd-inhibit` child exits nonzero ("system sleep inhibitor process exited; reacquiring...") and never escalates. On feta this has been failing on *every* acquisition since the beginning: polkit denies `org.freedesktop.login1.inhibit-block-sleep` to the daemon's non-graphical ssh-spawned context ("interactive authentication required" — reproduced 2026-08-03 with an env replica). Result: the vessel-union predicate from #1336 is correct but moot; feta suspends on KDE's idle timer under active crews — including mid-crew-turn on 2026-08-03 at 12:40 local, and every prior feta-asleep-under-crews incident.

Detect persistent acquisition failure (N consecutive nonzero exits while `required()` holds) and surface it as a host-level condition visible in fleet health / the fleet view — "sleep inhibition required but not held (reason)" — rather than a DEBUG/WARN log line nobody sees (the current daemon runs with default log filtering and the loop is invisible).

## Acceptance criteria

- [ ] Repeated child-exit while required ⇒ a host condition/attention flag observable via `host list` / fleet view, carrying the child's stderr (the polkit denial text)
- [ ] Condition clears when acquisition succeeds
- [ ] Unit tests through the existing injected-runner seam covering fail-loop → condition → recovery

## Out of scope

The host-side fix itself (polkit rule granting the inhibit action to the daemon's user), which lives in the fleet repo.

Relates: #1336, #1334.

rjwittams:
Closes #1347.

## Summary

- publish persistent sleep-inhibitor acquisition failure as a degrading `SleepInhibition` host condition
- preserve that condition across heartbeat status refreshes and clear it only after acquisition succeeds
- carry inhibitor stderr into fleet-health diagnosis and cover fail-loop, recovery, heartbeat persistence, and fleet projection

## Tests

### DIFF
--- a/crates/flotilla-daemon/src/runtime.rs
+++ b/crates/flotilla-daemon/src/runtime.rs
@@ -41,7 +41,7 @@ use flotilla_resources::{
     InputDefinition, InputMeta, PlacementPolicySpec, Presentation, Project, Regard, ReplicationClass, Repository, ResourceBackend,
     ResourceError, ResourceObject, Stance, TerminalSession, TerminalSessionSource, Vessel, VesselRequirement, WorkflowTemplate,
     WorkflowTemplateSpec, AGENT_ADAPTERS_CAPABILITY, CREDENTIAL_REFS_ENV, CREDENTIAL_REF_SESSION_TAG, HELD_CREDENTIALS_CAPABILITY,
-    MANAGED_BY_LABEL, REGISTERED_RESOURCE_KINDS,
+    MANAGED_BY_LABEL, REGISTERED_RESOURCE_KINDS, SLEEP_INHIBITION_CONDITION_TYPE,
 };
 use serde_json::json;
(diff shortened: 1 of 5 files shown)


rjwittams:
Acknowledged. Preserving the existing degrading condition while the fresh in-memory counter rebuilds is intentional: an `Acquiring` observation is not evidence of recovery, so clearing or refreshing the condition at that point would briefly make the host appear healthy. A successful `Held` observation clears it, while the third fresh failure refreshes the message and timestamp. No code change needed for this non-blocking display staleness.

PR #1352, opened 2026-08-03
fix: consult forge before checkout settlement

### LINKED ISSUE #1342: checkout integration no-CR early return keys on the checkout's spec ref — a PR opened from a different branch is invisible to settlement

rjwittams:
## Summary

- require a successful forge lookup before a zero-commit checkout can count as landed
- associate crew completion PR URLs with their convoy repository so PRs opened from a different head branch are queried by ID
- preserve Unknown when the forge is unavailable and retain no-CR checkout settlement after the forge confirms absence

## Root cause

Checkout integration returned Landed before contacting the forge whenever HEAD had no commits beyond the base. It also searched by the checkout's provisioned ref when no change request was bound, so a crew-delivered PR from another branch was invisible to settlement.

## Validation

Closes #1342

### DIFF
--- a/crates/flotilla-core/src/checkout_integration.rs
+++ b/crates/flotilla-core/src/checkout_integration.rs
@@ -246,24 +246,11 @@ async fn inspect_landed(
     change_request_id: Option<&str>,
     observed_at: &str,
 ) -> (IntegrationCondition, Option<LandedEvidence>, Option<ChangeRequestObservation>) {
-    // Git evidence first: a branch with no commits beyond its base has
-    // nothing to land, so it counts as landed without consulting the forge at
-    // all. This keeps untouched checkouts landable in environments where `gh`
-    // is absent or unauthenticated.
+    // Git evidence alone cannot prove that no change request remains
+    // outstanding. In particular, a crew may publish its work from a branch
+    // other than the checkout's provisioned ref. Always consult the forge;
+    // failure to do so leaves the condition unknown.
     let comparison = compare_branch_to_base(runner, checkout_path, base_ref).await;
-    if change_request_id.is_none() {
-        if let BaseComparison::Counted { base_ref, count: 0 } = &comparison {
-            return (
-                IntegrationCondition::builder()
-                    .value(ConditionValue::True)
-                    .details(vec![format!("branch has no commits beyond {base_ref}")])
-                    .observed_at(observed_at.to_string())
-                    .build(),
-                None,
-                None,
-            );
-        }
-    }
     let args = match change_request_id {
         Some(id) => vec!["pr", "view", id, "--json", "number,state,mergedAt,baseRefName,mergeable"],
         None => {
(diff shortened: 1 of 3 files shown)


rjwittams:
Fixed in e3b6eb6c.

- Checkout-only inspection can no longer prove the absence of a convoy-associated CR; it records Unknown when no bound ID is available.
- Settlement and teardown use a convoy-aware inspection entry point, and fresh Unknown observations are re-probed with convoy context instead of trusted.
- Completion PR extraction now scans prose and Markdown for matching GitHub `/pull/` and Forgejo `/pulls/` URLs, while rejecting URLs for other repositories.
- Added direct parser tests and a regression proving checkout-only lookup cannot report Landed=True without convoy association.

The exact format, clippy, and locked workspace test gates pass after these changes.

rjwittams:
Added doc comments distinguishing checkout-only and convoy-aware inspection. Filed the periodic status visibility follow-up as #1354: https://github.com/flotilla-org/flotilla/issues/1354. Exact formatting, clippy, and locked workspace tests pass.

rjwittams:
Fixed in 871e9d56 by storing the resolved RepositorySpec's canonical remote identity in direct convoy repository snapshots instead of the raw `--repo` input. This also canonicalizes adopted-checkout snapshots through the same path. Updated the direct-admission regression to supply a mixed-case `.git` URL and assert the canonical HTTPS URL, and updated the adopted-checkout assertion. Exact formatting, workspace Clippy, and locked workspace tests pass.

rjwittams:
Acknowledged; no further action required. The final exact local CI-equivalent format, workspace Clippy, and locked workspace test run is green.

PR #1358, opened 2026-08-03
fix: guard checkout peer replication

### LINKED ISSUE #1355: Checkout records are not replicated to peer hosts — cross-host convoys can never observe settlement evidence
## What happened

With #1346 deployed, two kiwi-authority convoys with feta-placed vessels (`github-app-minting`, `specref-blind-settlement`) held `Landing` indefinitely after their PRs (#1353, #1352) merged 2026-08-03T17:57Z. Diagnosis: kiwi's federated view contains feta's **vessels** but **zero checkout records** (`flotilla resource list checkouts` on kiwi: empty; on feta: both present). The reconciler's expected-vs-observed guard therefore correctly refuses to settle — expected refs are recorded on the convoy, observed evidence never arrives.

This is the other half of #1341's root cause: the daemon impl discarding the federated list was fixed, but the federated list is *empty anyway* because checkout records never replicate. Pre-#1346 the same gap produced vacuous settlement (teardown under open PRs); post-#1346 it produces visible stuck-in-Landing. Same defect, opposite symptom — the guard converted silent-wrong to loud-stuck as designed.

## Fix

Include Checkout in the peer replication set so the authority host's reconciler can observe them (the resource layer already supports it — #1346's regression tests drive a two-backend federated scenario via `replica_writer`; it is the daemon peer replication that omits the kind). Verify with a cross-host convoy reaching `Landed` on its own after a real merge.

## Acceptance criteria

- [ ] Checkout records created on a vessel host appear in the authority host's federated view
- [ ] A cross-host convoy in `Landing` transitions to `Landed` via the reconciler after its change request merges, no manual intervention
- [ ] InProcessDaemon-level test for the replication set including checkouts

Note: full end-to-end correctness also needs the #1352 probe fix deployed (feta's local records carry `landed: true` latched from before the crew pushed — #1342's hole + #1343's latch) — the latch ruling (#1343) matters here too.

Relates: #1341/#1346, #1342/#1352, #1343, ADR 0021/0028, ADR 0025 (store authority/federation).

rjwittams:
Closes #1355.

## Summary

- exercise checkout replication through a connected pair of `InProcessDaemon`s
- verify the authority host's federated resolver observes the vessel-host checkout as a replica
- verify landed settlement evidence survives replication without re-authoring the checkout locally

## Context

Checkout is already declared as a home-bound runtime resource and the daemon spawns peer replication from the registered resource set. The missing piece was a daemon-level contract proving that `Checkout` remains in that set. Existing coverage asserted Convoy, Vessel, and TerminalSession replication but could not catch checkout omission, which is exactly the failure mode that leaves cross-host convoys stuck in Landing.

## Validation

### DIFF
--- a/crates/flotilla-daemon/tests/overlay_replication.rs
+++ b/crates/flotilla-daemon/tests/overlay_replication.rs
@@ -4,7 +4,8 @@ use flotilla_core::{config::ConfigStore, in_process::InProcessDaemon, providers:
 use flotilla_daemon::server::test_support::spawn_in_memory_request_topology;
 use flotilla_protocol::{HostName, PeerConnectionState};
 use flotilla_resources::{
-    watch_resource_kind_replica_sources, Convoy, ConvoySpec, Host, HostSpec, HostStatus, InMemoryBackend, InputMeta, Project, ProjectSpec,
+    watch_resource_kind_replica_sources, Checkout, CheckoutPhase, CheckoutSpec, CheckoutStatus, ConditionValue, Convoy, ConvoySpec, Host,
+    HostSpec, HostStatus, InMemoryBackend, InputMeta, IntegrationCondition, ObservedCheckoutSpec, Project, ProjectSpec, RepositoryKey,
     ResourceBackend, ResourceProvenance, SqliteBackend, TerminalSession, TerminalSessionSource, TerminalSessionSpec, Vessel, VesselSpec,
 };
 use futures::StreamExt;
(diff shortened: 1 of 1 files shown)


rjwittams:
Thanks for checking this. No code change needed for the non-blocking import nit; the fully qualified one-off status type is valid under the repository's import convention and keeps the fixture's less central construction detail local.

The acceptance path is split across focused contracts: `federated_open_checkout_holds_landing_on_authority_host` verifies that `ConvoyReconciler` consumes federated checkout state, while `landing_with_settled_change_request_becomes_landed` verifies the Landing-to-Landed transition from settled evidence. This PR adds the previously missing daemon transport seam between them: two connected `InProcessDaemon`s preserve landed evidence in the authority's federated resolver. Keeping that test focused makes a future omission from peer replication fail at the responsible boundary.

PR #1360, opened 2026-08-03
fix: guard daemon socket ownership

### LINKED ISSUE #1357: A failed connect_or_spawn daemon unlinks the live daemon's socket file, breaking all host-local CLI
## What happened

2026-08-03 ~18:04Z on feta: the daemon (started 17:26Z) kept serving the mesh, but its Unix socket *file* vanished from `~/.config/flotilla/run/` while the bound listener lived on (visible in `ss -xl` as an unlinked socket). Every subsequent host-local CLI call failed with "timed out waiting for daemon to start (10s)": connect fails (no file) → `connect_or_spawn` spawns a competing daemon → that daemon loses the daemon lock/lifecycle and exits — and on its way out **removes the socket path belonging to the live daemon**. First trigger appears to be a CLI read racing the daemon (the run-2 Bosun's `flotilla resource get`); after that the host is wedged until a manual daemon restart (which is what recovered it).

## Fix

The socket file must be owned by exactly the process that bound it: a daemon that fails to acquire the lifecycle/lock must not unlink the socket path (only remove paths it created itself — e.g. bind to a temp path and rename on lock acquisition, or gate cleanup on having won the lock). Also worth a CLI-side guard: if connect fails but the lock holder is alive, report that instead of spawning.

## Acceptance criteria

- [ ] Losing spawner never removes a live daemon's socket; regression test at the lifecycle/server level
- [ ] CLI distinguishes "daemon alive but socket missing" (a diagnosable error) from "no daemon"
- [ ] The race (connect fails during restart window → spawn → lock loss) leaves the host recoverable without manual intervention

Relates: #1313 (socket lifecycle), the 2026-08-03 feta incident during Bosun run 2 (#1321).

rjwittams:
Closes #1357

## Summary

- refuse client-side daemon spawning when the lifecycle lock proves an existing daemon is alive but its socket is missing
- make server shutdown unlink only the exact socket inode the server bound
- cover lifecycle lock loss and the restart/replacement socket race with regressions

## Verification

### DIFF
--- a/crates/flotilla-client/src/lib.rs
+++ b/crates/flotilla-client/src/lib.rs
@@ -365,20 +365,22 @@ fn spawn_daemon(config_dir: &Path, config_dir_override: Option<&Path>, socket_ov
 pub async fn connect_or_spawn(
     socket_path: &Path,
     config_dir: &Path,
+    state_dir: &Path,
     config_dir_override: Option<&Path>,
     socket_override: Option<&Path>,
 ) -> Result<Arc<SocketDaemon>, String> {
-    connect_or_spawn_with_optional_surface(socket_path, config_dir, config_dir_override, socket_override, None).await
+    connect_or_spawn_with_optional_surface(socket_path, config_dir, state_dir, config_dir_override, socket_override, None).await
 }
 
 pub async fn connect_or_spawn_with_surface(
     socket_path: &Path,
     config_dir: &Path,
+    state_dir: &Path,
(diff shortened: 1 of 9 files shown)


rjwittams:
Reviewed—thank you. No code change needed: the residual race and metadata failure are explicitly non-blocking, and resolving `PathPolicy` per reconnect preserves the current process environment at the moment a new daemon may be spawned. All acceptance-criteria regressions and all CI checks pass.

PR #1365, opened 2026-08-03
feat: add leaf evaluator core and blocking wait

### LINKED ISSUE #1362: Leaf engine slice 1: evaluator core, subscription table, and blocking wait over machine/claim leaves
## Parent

#1322 (the leaf-evaluation engine). Contract: the six rulings recorded there 2026-08-03; ADR 0027 (condition leaves, single subscription surface, suspend ladder) and ADR 0028 (the engine as shared substrate) frame it.

## What to build

The engine core with no new observations: a `LeafSubject` seam (kind + field path → three-valued value) and evaluator over a **closed admitted vocabulary** covering machine leaves (`convoy.phase`, `vessel.phase`, `work.phase`) and claim leaves (`work.latest-claim.disposition`, `.claimed-at`); the single subscription table (row = leaves OR-set with literals bound/frozen at subscription time, watcher, freshness demand, created-at, reserved episode-key fields); and the **WaitCaller** watcher: `flotilla wait --for <leaf> [--for <leaf>…]` blocks until one admitted leaf evaluates True, then reports which fired. Block-only per ADR 0027's suspend ladder — the row dies with the connection; no persistence.

Leaf form is `(record address, field path, operator, bound literal)` → True | False | Unknown; Unknown (absent record, unsatisfied freshness) triggers nothing. Unknown *path* is a loud admission error at subscribe time, never a silent row. Evaluation is event-driven off store writes (watch), not scanning. Implement the seam with a hand-written per-kind accessor table for now — shape it so a descriptor-backed impl (facet research, prototypes/1322-facet-leaf-evaluator.md when it lands) can replace it behind the unchanged trait.

## Acceptance criteria

- [ ] `flotilla wait --for "convoy/<name> .status.phase == Landed" --timeout <s>` blocks and unblocks correctly (multiple `--for` = OR, reports which leaf fired; timeout is a clean nonzero exit)
- [ ] Claim leaves work: wait on a work's latest-claim disposition
- [ ] Unknown-path subscription is rejected loudly at admission with the admitted vocabulary named
- [ ] Absent record / freshness-unsatisfied evaluates Unknown and does not fire
- [ ] Daemon restart drops WaitCaller rows; a blocked caller errors with a clean "daemon restarted"
- [ ] Behavior tests through the in-memory backend (evaluator is pure over stored records — no replay fixtures needed); contract tests shared across store backends per the testing philosophy

## Blocked by

None — can start immediately. (The facet research informs internals but does not gate; the seam holds either implementation.)

rjwittams:
## Summary

- add the typed `LeafSubject` seam and closed leaf vocabulary for convoy, vessel, and work state
- add an ephemeral, event-driven subscription table with shared in-memory and SQLite backend contract tests
- add `flotilla wait --for ...`, including OR conditions, freshness constraints, timeouts, and clean daemon-restart errors
- preserve and expose crew completion dispositions for work claim leaves

## Testing

- `cargo clippy --workspace --all-targets --locked --features flotilla-daemon/skip-no-sandbox-tests -- -D warnings`

Closes #1362

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1073,6 +1073,7 @@ name = "flotilla"
 version = "0.1.0"
 dependencies = [
  "bon",
+ "chrono",
  "clap",
  "color-eyre",
  "flotilla-client",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,3 +65,4 @@ serde_yml = "0.0.12"
 libc = "0.2"
 uuid = { version = "1", features = ["v4"] }
 humantime = "2"
+chrono = { version = "0.4", default-features = false, features = ["clock"] }
--- a/crates/flotilla-client/src/lib.rs
+++ b/crates/flotilla-client/src/lib.rs
@@ -11,8 +11,8 @@ use std::{
 use async_trait::async_trait;
 use flotilla_core::daemon::DaemonHandle;
 use flotilla_protocol::{
-    Command, ConnectionRole, DaemonEvent, Message, NodeId, QueryCursor, QueryId, ReplayCursor, RepoIdentity, RepoInfo, RepoSnapshot,
-    Request, Response, ResponseResult, StatusResponse, StreamKey, SurfaceDeclaration, TopologyResponse, PROTOCOL_VERSION,
+    Command, ConnectionRole, DaemonEvent, LeafFire, Message, NodeId, QueryCursor, QueryId, ReplayCursor, RepoIdentity, RepoInfo,
+    RepoSnapshot, Request, Response, ResponseResult, StatusResponse, StreamKey, SurfaceDeclaration, TopologyResponse, PROTOCOL_VERSION,
 };
 use flotilla_transport::message::{connect_unix_message_session, MessageSession};
 use tokio::sync::{broadcast, oneshot, Mutex};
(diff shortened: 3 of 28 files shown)


rjwittams:
Thanks — I addressed the disposition footgun in fee22766. A completion with an omitted `--disposition` now preserves the previously settled disposition and timestamp, and `crew_completion_updates_only_the_calling_agent` covers completion-with-disposition followed by a message-only completion.

The freshness behavior is deliberate in this slice. #1362 scopes this as the engine core “with no new observations” and explicitly requires freshness-unsatisfied evaluation to remain `Unknown`. Convoy/vessel subjects currently have no observation timestamp to prove the demand; inventing one from phase transition fields would not represent observation freshness, while rejecting the request would contradict the specified three-valued freshness behavior. The later refresher/observation slices can supply a real `observed_at` behind the unchanged `LeafSubject` seam.

rjwittams:
Addressed both substantive findings in ef0bd678:

- Admission now rejects `<`, `<=`, `>`, and `>=` for text-valued leaves; ordering remains available only for `.latest-claim.claimed-at`. A regression test covers loud rejection.
- Wait fires now have a dedicated client broadcast channel, so unrelated daemon/replay traffic cannot displace the one-shot event. `run_wait` also treats any pathological lag on that dedicated stream as a loud error rather than silently continuing.

I also added the suggested `--fresher-than` help text documenting that this slice only has observation timestamps for work claim leaves. Exact fmt, full clippy, and the sandbox-safe workspace suite all pass on this revision.

rjwittams:
Thanks for the final pass. The remaining notes are follow-up-shaped and outside this slice; no further code changes are needed here.

Issue #1321, opened 2026-08-02
Bosun-over-verbs experiment: an agent driving the workflow verb surface


rjwittams:
## Parent

ADR 0027 build-order step 1 (after #1288, delivered). Closes out the exploration recorded in `prototypes/1283-workflow-substrate.md`; part of wayfinder map #1262 (closed).

## What to build

Run the first proving experiment for the workflow substrate: a **Bosun** — an agent crew member whose brief is to *drive a workflow*, not to code — operating with nothing but the ruled verb surface:

- **Reads**: `flotilla resource get/list/watch --json` (delivered by #1288), plus transcripts. No privileged driver API.
- **Manipulation**: the existing verbs — deliver/handoff a brief, complete, hold (loud, with reason), escalate, delete/abandon.
- **No `flotilla wait`** yet: the Bosun polls or is nudged. Wasteful is fine; this step exists to test *"is an agent with verbs enough?"* before synchronization ships.

**Placement (ruled 2026-08-02):** the Bosun is a crew member *inside* the target convoy — membership is notional/logical, physical deployment is flexible. For this experiment it rides in the convoy's sole vessel as a full-fat agent alongside the coder. Own-vessel is a requirements question for later: vessels are capability and security boundaries (platform, has-a-gpu), not workflow structure — crew do the work, vessels enable it. Because the Bosun is inside the convoy, **no scope widening is needed**: default convoy-scoped verbs suffice.

Concretely: author a Bosun brief for the **inside-out shepherd** scenario (exit condition is the observed world-fact `cr.merged`, not the agent's judgment of doneness; exhaustion is a loud `convoy hold` at a round budget, not agent fatigue) and run it against a real convoy working a real issue. The Bosun delivers round briefs to the coder crew member, watches PR state via reads, and ends the loop only on observation.

## What the findings must cover (deferred here by ADR 0027 and the placement ruling)

- **Verb gaps**: which verbs were missing or awkward; where polling hurt; where the agent leaked judgment back into the loop.
- **Scope-widening proposal (forward-looking, for #650)**: the experiment needs none, but a standing governor over many convoys does. Propose the dispatching-principal grant shape — what a multi-convoy driver's principal grants and how it is granted at dispatch — informed by what the Bosun actually touched.
[...]

PR #1366, opened 2026-08-03
feat: observe change requests for leaf waits

### LINKED ISSUE #1363: Leaf engine slice 2: observed ChangeRequest records, demand-driven refresher, cr.* leaves
## Parent

#1322 — rulings 1, 4, and 6 recorded there (evaluator/refresher split; v1 vocabulary; cross-host observation authority).

## What to build

The observation refresher and the first-class observed ChangeRequest record. Subject-keyed identity (`cr/<service>/<scope>/<number>`), fields: `state (open|merged|closed)`, `head_sha`, `checks (pass|fail|pending)`, `review.actionable_at_head`, `mergeable`, all with `observed-at`, Unknown permitted. Refresh is **demand-driven**: a subscription on a `cr.*` leaf causes the record's refresher to run (per-kind default cadences as tunables: CR state ~60–120s, checks faster while pending; tightened by the max freshness demand among live subscriptions); no subscriber → no polling. The first demanding host creates the record and is its observing authority; other hosts read replicas (replication-set invariant from the #1322 rulings — every leaf-addressable kind replicates; see #1355). Authority capability loss → staleness → dependent leaves Unknown.

`cr.*` paths join the admitted vocabulary; `flotilla wait --for "cr/github.com/flotilla-org/flotilla/<n> .state == merged"` works end-to-end.

## Acceptance criteria

- [ ] Subscribing a cr leaf starts observation; unsubscribing all stops it (verifiable via refresher state/records)
- [ ] `wait` on cr.merged unblocks on a real merge (replay-fixture coverage for the gh interactions; record/replay per the testing philosophy)
- [ ] Freshness demand tightens cadence; stale record → Unknown → no fire
- [ ] Cross-host: a replica-reading host's evaluator sees the authority's record (in-process federated test)
- [ ] Demo: a Bosun-shaped script drives the #1321 decision table with `wait` instead of sleep loops (run-3-lite — polling-cost comparison recorded on #1321)

## Blocked by

Slice 1 (the engine core and admission surface).

rjwittams:
Closes #1363.

## Summary

- add replicated, subject-keyed observed `ChangeRequest` resources with per-field observation timestamps and structural Unknown values
- admit and evaluate `.state`, `.head-sha`, `.checks`, `.review.actionable-at-head`, and `.mergeable` leaves
- start GitHub observation on first CR-leaf demand, tighten cadence for pending checks/freshness demand, and stop polling after the last subscription ends
- garbage-collect the observed record after its last demand is released (and clear orphaned local observations on daemon startup)
- coalesce observations by field value before persistence, so identical polls do not write status or advance observation timestamps
- keep evaluation store-only and let replica-reading hosts consume the observing authority record without becoming a second authority
- add the #1321 Bosun decision-table run-3-lite script using `flotilla wait` instead of sleeps

## Design-round riders

**Thin history:** `ChangeRequest` uses the new `ReplicationClass::Observations`. Embedded stores retain only the latest watch-handoff event for an observation stream while still notifying live watchers. A peer behind that one-event window receives watch expiry and relists the current snapshot instead of paying to replay observation history.

**Bound subjects only:** CR leaves accept only an exact `cr/<service>/<scope>/<number>` subject. Collection, wildcard, and query-shaped CR addresses are rejected; browsing remains an Aggregator concern and cannot materialize observed records.

## Verification

Focused coverage includes replayed GitHub OPEN→MERGED observation, subscribe/materialize → last-unsubscribe/delete, N identical polls → zero status writes, one-event observation-history bounds in both embedded stores, exact-subject address admission, freshness cadence tightening, stale-as-Unknown, and replica-reading evaluation.

### DIFF
--- a/crates/flotilla-core/src/change_request_observer.rs
+++ b/crates/flotilla-core/src/change_request_observer.rs
@@ -1,0 +1,295 @@
+use std::{collections::HashMap, path::Path, sync::Arc, time::Duration};
+
+use async_trait::async_trait;
+use chrono::{DateTime, Utc};
+use flotilla_protocol::LeafAddress;
+use flotilla_resources::{
+    change_request_record_name, ChangeRequest, ChangeRequestReviewObservation, ChangeRequestSpec, ChangeRequestStatus, InputMeta,
+    Observation, ObservedChangeRequestState, ObservedChecks, ObservedMergeability, ResourceBackend, ResourceProvenance,
+};
+use tokio::{sync::Mutex, task::JoinHandle};
+
+use crate::providers::{run, CommandRunner};
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct ChangeRequestRef {
+    pub namespace: String,
+    pub service: String,
+    pub scope: String,
+    pub number: u64,
+}
+
+impl ChangeRequestRef {
+    pub fn from_address(namespace: &str, address: &LeafAddress) -> Option<Self> {
+        let LeafAddress::ChangeRequest { service, scope, number } = address else { return None };
+        Some(Self { namespace: namespace.to_string(), service: service.clone(), scope: scope.clone(), number: *number })
+    }
+
+    fn record_name(&self) -> String {
+        change_request_record_name(&self.service, &self.scope, self.number)
+    }
+}
+
+#[async_trait]
+pub trait ChangeRequestObservationSource: Send + Sync {
+    async fn observe(&self, subject: &ChangeRequestRef) -> Result<ChangeRequestStatus, String>;
+}
+
+pub struct GhChangeRequestObservationSource {
+    runner: Arc<dyn CommandRunner>,
+}
+
+impl GhChangeRequestObservationSource {
+    pub fn new(runner: Arc<dyn CommandRunner>) -> Self {
+        Self { runner }
+    }
+}
+
+#[async_trait]
+impl ChangeRequestObservationSource for GhChangeRequestObservationSource {
+    async fn observe(&self, subject: &ChangeRequestRef) -> Result<ChangeRequestStatus, String> {
+        if subject.service != "github.com" {
+            return Err(format!("change request observation service `{}` is not available on this host", subject.service));
+        }
+        let number = subject.number.to_string();
+        let output = run!(
+            self.runner,
+            "gh",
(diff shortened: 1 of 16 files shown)


rjwittams:
Fixed the classic StatusContext correctness bug in 37105f52. `check_pending` now treats only a present, null `conclusion` as an unfinished CheckRun, and `classic_success_status_without_check_run_fields_passes` covers a successful `{ "state": "SUCCESS" }` entry with no CheckRun-only fields.

The concurrent first-demand authority race noted in the lower-priority section is the authority election/handoff case explicitly deferred by the #1322 ruling; this slice preserves first-visible-demand semantics. The namespace scan is retained here for the correctness-first replica check and can be optimized if observed CR cardinality makes it material.

rjwittams:
Fixed in 46ab15a6. Both first-demand creation and the lower-severity publish fallback now use the crate's create-or-get-on-Conflict idiom. `concurrent_first_demands_converge_on_one_authority_record` exercises two same-daemon demands and verifies one record plus two live demands.

Also confirmed the review semantic: an explicit GitHub `reviewDecision: null` means no changes are requested, so it now records `actionable_at_head = false`; only an absent field remains structural Unknown. `explicit_null_review_decision_is_not_actionable` covers that case.

rjwittams:
Fixed in e63ade9d. Each active subject now owns a `Notify`; a late freshness demand signals it, the refresh loop races that signal against its current sleep, and re-observes immediately instead of waiting out the old state cadence. `late_freshness_demand_preempts_existing_state_cadence_sleep` starts with a 90-second cadence, joins after 5 seconds, and verifies the second observation without advancing the remaining 85 seconds.

I also factored the duplicated record creation path into `get_or_create_record`, shared by initial demand and publication.

rjwittams:
Fixed in 16aaadb1. Startup orphan collection now asks the embedded backend for every namespace containing locally authored ChangeRequest records, then deletes the observations in each namespace. The namespace enumeration is covered on both in-memory and SQLite stores, and startup_garbage_collection_covers_non_default_namespaces verifies that an orphan under ops is removed rather than surviving restart.

rjwittams:
Acknowledged. This pass found no blocking issues; the two explicitly non-blocking efficiency notes remain outside the design-round riders and do not affect correctness.

rjwittams:
Acknowledged. The sixth pass confirms the blocking riders and prior review fixes hold on the current head, with no new blocking issues.

PR #1370, opened 2026-08-03
Fix cross-host landing integration observation

### LINKED ISSUE #1367: Cross-host landing evaluation probes from the wrong host — integration observation must run on the checkout's authority host
## What happened

Post-#1358 deploy (2026-08-03 ~21:45Z): kiwi's reconciler now *sees* feta's replicated checkouts (replication works) and probes them each Landing pass — but all four merged-PR convoys (`checkout-replication`, `daemon-log-silence`, `leaf-engine-core`, `socket-unlink-guard`) hold `Landing` indefinitely. From `~/.local/state/flotilla/log/flotillad.jsonl` (readable at last, thanks #1359):

- Every pass warns `failed to persist checkout integration conditions during landing evaluation … resource not found` — the status writeback targets the local store where the checkout is a non-writable replica (`crates/flotilla-core/src/in_process.rs:6257-6266`).
- The hold itself: the fresh probe (`inspect_convoy_checkout_integration` via `runner_for_resource_checkout`) does not yield `landed=True` for a remote checkout — the convoy-authority host cannot meaningfully probe a checkout whose path, environment, and forge credentials live on another host. And when the runner errors, `convoy_change_requests_settled_for_checkouts` returns `Ok(false)` **silently** (`in_process.rs:6247-6250` — no log).
- Feta, which could probe these checkouts trivially, never does: the convoy is not its authority, so nothing on feta triggers integration observation post-merge; its replica-visible evidence stays latched at pre-push `landed=true` values (see #1343).

## Contract (aligned with #1322 ruling 6 — per-host observation authority)

**Integration observation runs on the checkout's authority host; the convoy reconciler consumes replicated evidence.**

- The checkout's authority daemon observes integration (probe + forge consult per #1352) for any of its checkouts whose owning convoy — authority or replica — is in `Landing`/terminal-verification, on the existing cadence/TTL. Trigger source: the replicated convoy record's phase.
- Status writeback is then always authority-local (the current `apply_resource_status_patch` works unchanged there); the replica-write path on the convoy-authority host is removed, not patched.
- `convoy_change_requests_settled_for_checkouts` on the convoy-authority host evaluates **only** stored evidence (fresh-within-TTL `landed`, per the existing `LANDING_EVIDENCE_TTL` branch); the remote-probe fallback is removed. Missing/stale evidence ⇒ hold, with a log line naming which checkout's evidence is missing — never a silent `Ok(false)`.
- The teardown gate follows the same rule for replicated checkouts.

## Acceptance criteria

- [ ] Two-backend federated test: convoy authority on A, checkout authority on B; B observes post-merge, A transitions `Landing → Landed` with no cross-host probing from A
- [ ] Runner/evidence failures during evaluation are logged, never silent `Ok(false)`
- [ ] The four real convoys named above transition autonomously after deploy (fleet verification recorded on this issue)
- [ ] Stale-latched `landed=true` from pre-push observations cannot satisfy the fresh-evidence branch (interacts with #1343 — do not resolve the latch ruling here, just ensure TTL freshness is required)

Relates: #1355/#1358 (replication — the necessary first half), #1352, #1343, #1322 ruling 6, ADR 0025 (federated actuation), ADR 0021/0028.

rjwittams:
## Summary

- move Landing and teardown integration evaluation to fresh stored checkout evidence only
- make checkout-authority reconcilers react to local or replicated owning convoys and probe with convoy change-request context
- log named missing/stale evidence and reject stale latched landed results
- add a two-backend regression proving the checkout authority probes while the convoy authority lands from replicated evidence

## Verification

Fleet verification of the four affected production convoys is intentionally post-deploy and remains to be recorded on #1367 after this change lands.

Closes #1367

### DIFF
--- a/crates/flotilla-controllers/src/reconcilers/checkout.rs
+++ b/crates/flotilla-controllers/src/reconcilers/checkout.rs
@@ -2,10 +2,14 @@ use std::{sync::Arc, time::Duration};
 
 use async_trait::async_trait;
 use chrono::{DateTime, Utc};
+use flotilla_core::checkout_integration::{
+    checkout_observation_lacks_convoy_association, convoy_change_request_id_for_checkout, LANDING_EVIDENCE_TTL,
+};
 use flotilla_resources::{
-    controller::{ReconcileOutcome, Reconciler},
+    controller::{ReconcileOutcome, Reconciler, ReplicaConvoyCheckoutWatch, SecondaryWatch},
     Checkout, CheckoutBranchProvenance, CheckoutIntegrationStatus, CheckoutPhase, CheckoutSpec, CheckoutStatus, CheckoutStatusPatch, Clock,
-    Clone, ClonePhase, IntegrationCondition, ResourceBackend, ResourceError, ResourceObject, SystemClock, TypedResolver,
+    Clone, ClonePhase, Convoy, ConvoyPhase, IntegrationCondition, ReplicaReadResolver, ResourceBackend, ResourceError, ResourceObject,
+    ResourceProvenance, SystemClock, TypedResolver, ACTUATOR_SOURCE_ROOT_ANNOTATION, CONVOY_LABEL,
 };
 use tracing::warn;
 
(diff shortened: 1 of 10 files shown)


rjwittams:
Addressed the retry concern in 5b4b6da1: a missing CR observation after a failed forge lookup now respects the Landing TTL; only an observed-but-wrong CR ID bypasses freshness. Added `fresh_failed_change_request_lookup_waits_for_the_landing_ttl_before_retrying` to lock that behavior down.

The terminal-phase 30s TTL is intentional because teardown now consumes the same fresh stored evidence and cannot probe on demand. I added that rationale to `convoy_needs_terminal_evidence`.

Issue #1368, opened 2026-08-03
Resource store representation: storage, event history, wire, erased access — one deliberate revisit


rjwittams:
## Direction (Robert, 2026-08-03/04)

The store is serde-JSON at every level today: sqlite rows are `serde_json::Value` in TEXT columns, the event history peers replay is JSON, the peer wire is `serde_json::to_string` envelopes, and the typed↔JSON round-trip is monomorphized into every ControllerLoop copy (#928). Five pressures now converge on this one representation layer:

1. The leaf evaluator's kind-erased field access (#1322 ruling 3, seam landed in #1365)
2. The erased-store dyn-erasure seam + per-kind codec (#928 threads 7/10)
3. Observation churn: coalesced writes, thin/no event history for the Observations class (#1322/#1366 riders)
4. Wire/storage density (binary codec candidates via descriptor-driven format crates)
5. **Direct SQL queryability of table-ish resources** (Robert): descriptor-driven projection of fields into real sqlite columns would make high-cardinality-ish resources directly queryable — a feature, not just an efficiency fix — and is *another* facet motivator on the pile.

## Scope when taken up

One deliberate revisit, not incremental drift: storage schema (JSON blob vs projected columns per class), event-history retention/compaction per replication class, wire codec, and the erased access layer — all downstream of the facet verdict (prototypes/1322-facet-leaf-evaluator.md when it lands). Cheap-making context: no-BC phase (rebuild-and-restart, no migrations) and ADR 0025 store authority (per-host mechanical swap).

Not before: the facet research verdict + slice-1/2 leaf-engine seams in use (so the design has real consumers to shape against).

Relates: #928, #1322, #1366, ADR 0025, the #1340/#1322 ruling trail.

rjwittams:
Facet research verdict landed (see prototypes/1322-facet-leaf-evaluator.md and the summary on #1322): sequence this revisit on a **shared declared-table module** — leaf vocabulary, ADR 0024 ownership rows, and SQL projection columns speaking one static table format — not on facet adoption. Facet's value-free SHAPE schema remains the end-state candidate for the column-projection half (serde structurally can't provide it); the named adopt-triggers and a proposed one-kind Convoy spike are in the note.

rjwittams:
Path correction for the comment above: the facet note landed at `docs/research/2026-08-03-facet-leaf-evaluator.md` (moved to the docs/research convention in PR #1369), not prototypes/.

rjwittams:
[...]

Issue #929, opened 2026-07-23
Measure the build: cargo --timings, monomorphization and dependency cost profile


rjwittams:
## Parent

#928 (convoy cycle time) — grounds threads 7 (compiler gotchas) and 10 (dependency sweep) in names instead of hunches. **Deliverable is a REPORT, not a diff** — commit it as `docs/reports/build-profile-2026-07.md` on a branch and PR that file alone.

## What to measure

On this workspace, clean-ish conditions (note sccache state in the report):

1. `cargo build --timings` (dev profile, full workspace): top crates by compile time, critical path, parallelism utilisation. Same for `cargo test --workspace --no-run` (the test-build cost crews pay every round).
2. **Monomorphization/LLVM-lines profile**: `cargo llvm-lines` (install if needed) on the fattest crates from (1) — name the top generic offenders (suspects worth checking: bon builders, serde derives on big protocol types, async-trait expansion).
3. **Link + disk**: where the 31-49G target/ goes (per-profile breakdown); incremental-link time for a one-line change in flotilla-core vs flotilla-daemon vs flotilla-tui (the crew inner-loop cost).
4. **Dependency sweep**: `cargo tree -d` duplicates; top 10 dependencies by compile time from (1); anything egregious.
5. Quick literature check (no experiments): Cranelift backend viability on aarch64-apple-darwin for dev/PR builds; macOS link-time options. One paragraph each, cited.

## Acceptance criteria

- [ ] Report names the top 10 compile-time crates (workspace + deps) with numbers
- [ ] Top monomorphization offenders named with llvm-lines counts
- [ ] Disk breakdown of a representative target/ dir
- [ ] Incremental rebuild+link timings for a leaf change in core/daemon/tui
- [ ] Recommendations section ranked by expected win, each tied to a measurement
- [ ] No workspace code changes in the PR — report only

PR #1372, opened 2026-08-03
feat: spike facet-backed convoy leaf reflection

### LINKED ISSUE #1371: Facet spike: derive the Convoy type tree behind the LeafSubject seam and measure the three OPEN adoption unknowns
## Why now (ruled 2026-08-04, superseding the note's "queued candidate" sequencing)

Robert's structural critique of the seam-now-facet-later verdict: deferring evaluation until after the declared-table family accretes (slice 2 rows, #1368 projection columns, consolidated module) is a fake rejection — the sunk cost decides instead of the arguments. The honest decision point is before the accretion, and the spike is the instrument. Outcome either way is an explicit ruling: blockers → facet rejected for stated reasons and the shared declared-table module becomes the *designed* architecture; clean → facet becomes the substrate for the table family and #1368 sequences on it.

## What to build

A spike branch (draft PR, not for merge as-is) that:

1. Adds current facet (verify the crate's present-day API against docs.rs — the research note's facet claims are marked OPEN/pre-2026) with `#[derive(Facet)]` on the **Convoy type tree only** (ResourceObject<Convoy> and everything it transitively contains, including chrono timestamps, BTreeMaps, Options, and enums like ConvoyPhase).
2. Implements the existing `LeafSubject` seam (landed in PR #1365) for the Convoy kind via facet field-walking instead of the hand table, behind the unchanged trait — both impls compile side by side, tests pass against either.
3. Measures and reports, in the PR description, the three OPEN unknowns from docs/research/2026-08-03-facet-leaf-evaluator.md:
   - **Attribute parity**: can facet represent our serde shapes — `rename`/`rename_all` sites in the Convoy tree, and (checked as standalone probes even though outside the Convoy tree) `#[serde(flatten)]` on `PlacementStatus` and `#[serde(untagged)]` on `InputValue`? Do facet's serialized names match serde's exactly (leaf paths address serialized names)?
   - **Derive cost**: clean-build and incremental compile-time delta for the crate(s) gaining the derive (measure both, state the method).
   - **Schema walk**: can the value-free SHAPE be walked to validate a dotted path's existence and comparable type without an instance (the admission-validation and #1368 column-projection capability)?

## Acceptance criteria

- [ ] Draft PR with the side-by-side impl and a findings report answering all three unknowns with evidence (compile timings, code snippets for parity cases, a working path-validation probe or a demonstrated impossibility)
- [ ] A recommendation line: adopt / reject-with-reasons, mapped to the named triggers in the research note
- [ ] No changes to the seam's public contract; hand-table impl remains the active one

Relates: #1322 (ruling 3), #1368, #928 threads 7/10, docs/research/2026-08-03-facet-leaf-evaluator.md (PR #1369).

rjwittams:
Closes #1371

## Outcome

This spike derives current stable `facet` over `ResourceObject<Convoy>` and its transitive local/protocol type graph, adds a dormant `FacetConvoyLeafSubject` behind the unchanged `LeafSubject` trait, and leaves the handwritten `ConvoyLeafSubject` active in `flotilla-core`.

**Recommendation: reject-with-reasons.** Keep the shared declared-table module as the designed architecture and sequence #1368 on it. None of the research note's adoption triggers has fired: operator vocabulary remains declared and small; leaving serde is not a live plan; and facet remains a fast-moving 0.x admission-control dependency. The spike confirms the unique value-free-schema capability, but today it requires a duplicated naming/schema description and adds a material incremental check cost.

## Current API checked

Checked against current docs.rs/crates.io on 2026-08-10. The latest non-prerelease is `facet 0.46.5` (published 2026-05-26); `0.50.0-rc.5` is prerelease. `0.46.5` has Rust 1.90 MSRV; this measurement host used `rustc 1.94.0 (4a4ef493e 2026-03-02)`.

Current APIs used by the spike:

- `Facet::SHAPE` / `Shape::{ty,def}` for value-free traversal
- `Type::User(UserType::Struct|Enum)` and `Def::Option`
- `Peek::{new,into_struct,into_option,into_enum}` for value traversal
- `Field::{name,rename,is_flattened}` and `Shape::is_untagged`

## OPEN 1: attribute parity

**Representable, but not automatically shared with serde.** Facet does not consume `#[serde(...)]`; each relevant site needs a mirrored facet attribute. For example:

```rust
#[serde(default, rename = "ref", skip_serializing_if = "Option::is_none")]
#[facet(rename = "ref")]
pub r#ref: Option<String>,

#[serde(rename_all = "snake_case")]
#[facet(rename_all = "snake_case")]
pub enum IssueState { Open, Closed }
```

Evidence in `facet_attributes_match_serde_names_and_shapes` compares the facet effective names and serde JSON for every rename family in the Convoy graph:

- `ConvoySpec.ref` -> `ref`
- `ObjectMeta.owner_references` -> `ownerReferences`
- `OwnerReference.api_version` -> `apiVersion`
- `PlacementTargetHost.reference` -> `ref`
- `IssueState::Open` -> `open` (`snake_case`)
- `Stance::WorkspaceWrite` -> `workspace-write` (`kebab-case`)

Standalone shape probes also pass:

```rust
#[serde(flatten)]
#[facet(flatten, opaque)]
pub fields: BTreeMap<String, serde_json::Value>,

#[serde(untagged)]
#[facet(untagged)]
pub enum InputValue { String(String) }
```

`PlacementStatus`'s facet field reports `is_flattened()`, while serde produces `{ "host": "host-a" }`; `InputValue::SHAPE.is_untagged()` is true, while serde produces the plain string `"value"`.

Constraints found:

- Facet requires explicit enum layout, so reflected enums gained `#[repr(u8)]` or `#[repr(C)]`.
- `serde_json::Value` has no usable foreign `Facet` implementation under Rust's orphan rules. Those leaves are marked field-level `#[facet(opaque)]`; facet can retain flatten metadata but cannot statically enumerate arbitrary dynamic keys.
- Serde-only controls such as `skip_serializing_if` are not needed for path identity, but a future facet serializer would require those to be duplicated too.

So serialized path names match exactly in this spike, but only because tests enforce a second, manually mirrored attribute set. That is the primary correctness/stability blocker.

## OPEN 2: derive cost

Method: one baseline clone at the PR base and this checkout were measured on the same host with separate empty target directories. Clean command:

```text
CARGO_TARGET_DIR=<empty> CARGO_INCREMENTAL=0 cargo check -p flotilla-resources --locked
```

Incremental command: touch `crates/flotilla-resources/src/convoy.rs`, then repeat the same package check against the already-populated target (without forcing `CARGO_INCREMENTAL=0`). `/usr/bin/time` wall clock is reported. This package check includes both crates gaining derives (`flotilla-protocol` and `flotilla-resources`). One paired sample:

| check | baseline | facet | delta |
|---|---:|---:|---:|
| clean | 16.44 s | 17.90 s | +1.46 s (+8.9%) |
| touched `convoy.rs` | 0.66 s | 1.08 s | +0.42 s (+63.6%) |

Peak RSS on the clean checks was 323,564 KiB baseline vs 348,428 KiB facet (+24,864 KiB, +7.7%). These are local paired measurements, not a statistically sampled benchmark; the direction and incremental magnitude are enough to reject “derive cost is negligible” for this graph.

## OPEN 3: value-free schema walk

**Yes.** `facet_schema_at_path` starts from `<ResourceObject<Convoy> as Facet>::SHAPE`, unwraps `Def::Option`, selects struct fields by `field.rename.unwrap_or(field.name)`, and classifies a unit enum leaf as comparable text without constructing a Convoy:

```rust
assert_eq!(
    facet_schema_at_path(shape, ".status.phase"),
    Some(FacetComparableType::Text),
);
assert_eq!(facet_schema_at_path(shape, ".status.nope"), None);
```

The runtime sibling uses the same serialized-name rule over `Peek`; `FacetConvoyLeafSubject` and the active handwritten subject both return `Landing` for `.status.phase`. This proves the admission-validation and declared projection capability. Extending the classifier to scalar/timestamp types is straightforward; dynamic map keys remain runtime data rather than value-free schema.

## Trigger mapping

- **Arbitrary operator-authored paths / vocabulary outgrows declared tables:** not fired; current admitted paths remain closed and small.
- **#929 makes leaving serde live:** not fired; facet would duplicate serde rather than replace it.
- **Facet reaches admission-control stability posture:** not fired; stable release remains 0.46.x while the next API line is 0.50 RC, and the derive imposed layout/attribute duplication.

The schema walk is technically clean, but it does not outweigh those blockers or the measured compile cost.

## Verification

The full suite passed. One first-run unrelated ETXTBSY race in `zellij_pipe_sink_retries_patch_after_child_stdin_closes` passed alone and the complete suite then passed on rerun.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -508,6 +508,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "const-fnv1a-hash"
+version = "1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "32b13ea120a812beba79e34316b3942a857c86ec1593cb34f27bb28272ce2cca"
+
+[[package]]
 name = "const-oid"
 version = "0.9.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -971,6 +977,98 @@ dependencies = [
 ]
 
 [[package]]
+name = "facet"
+version = "0.46.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "df488bcda61dde160301ee6bce43d42397db599ffc689381a1873b5f035135f9"
+dependencies = [
+ "autocfg",
+ "facet-core",
+ "facet-macros",
+ "facet-reflect",
+]
+
+[[package]]
+name = "facet-core"
+version = "0.46.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c33cf9d80307e9a29f93e5a65ddb65b227b5fa082d649c65d4083345f5d398d0"
+dependencies = [
+ "autocfg",
+ "chrono",
+ "const-fnv1a-hash",
+ "iddqd",
+ "impls",
+]
+
+[[package]]
+name = "facet-macro-parse"
+version = "0.46.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c5ab349bd1823fbfa129e595a826ef7fb4d1b9c7215428b28de18044d009dbc1"
+dependencies = [
+ "facet-macro-types",
+ "proc-macro2",
+ "quote",
+]
+
+[[package]]
+name = "facet-macro-types"
+version = "0.46.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "79622a74665b47d1744998dd0d15b73eb0eecdfeb16630b436c5f03365b02e55"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "unsynn",
+]
+
+[[package]]
+name = "facet-macros"
+version = "0.46.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a6656fc7457975f5e1c86791b7a810c40ba53993823ae655f7c331e808c4b5ff"
+dependencies = [
+ "facet-macros-impl",
+]
+
+[[package]]
+name = "facet-macros-impl"
+version = "0.46.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f462bad2b5a574b2c94e9ad54162704ceacf3200741b4f0fb88020936b9003ec"
+dependencies = [
+ "facet-macro-parse",
+ "facet-macro-types",
+ "proc-macro2",
+ "quote",
+ "strsim",
+ "unsynn",
+]
+
+[[package]]
+name = "facet-path"
+version = "0.46.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "20670a58be24eb604b76ed3a9ca719e5cc5e11b0885dc3847ebb899aa95bf1bc"
+dependencies = [
+ "facet-core",
+]
+
+[[package]]
+name = "facet-reflect"
+version = "0.46.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "399876f2bded4a572385dc414336aad0002ce4b4154cd0d1cfe9d1881243c400"
+dependencies = [
+ "facet-core",
+ "facet-path",
+ "hashbrown 0.17.1",
+ "regex",
+ "smallvec 2.0.0-alpha.12",
+]
+
+[[package]]
 name = "fallible-iterator"
 version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1237,6 +1335,7 @@ dependencies = [
  "base64",
  "bon",
  "chrono",
+ "facet",
  "gethostname",
  "indexmap",
  "percent-encoding",
@@ -1258,6 +1357,7 @@ dependencies = [
  "bon",
  "bytes",
  "chrono",
+ "facet",
  "flotilla-controllers",
  "flotilla-core",
  "flotilla-protocol",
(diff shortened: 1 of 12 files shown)


rjwittams:
Addressed the two code nits in ad432ac6: `FacetConvoyLeafSubject` is now private to the existing leaf module (with no crate-root re-export), and the parity test initializes `status` directly without a redundant mutation. Pinned fmt, workspace clippy, and the full sandbox-safe workspace suite all pass after the change.

I am intentionally leaving the PR ready for review. The convoy crew delivery contract for this assignment explicitly requires a ready-for-review PR and says never to deliver a draft; that delivery instruction supersedes the issue snapshot's draft wording. The PR description and type docs continue to state that this is a non-adopted spike and that the handwritten implementation remains active, so the ruling is visible to a human merger.

rjwittams:
Thanks for confirming the code fixes and test coverage. No further code action is needed from this review. The PR remains ready for review because the crew delivery brief explicitly requires a ready (never draft) PR; the owner can make the deliberate accept/reject decision described in the findings before merge.

rjwittams:
Ruling withheld (Robert, 2026-08-04) — this PR stays open, unmerged and unrejected, pending a full read of the research. Two counterpoints to the reject-with-reasons recommendation are recorded for that read: (1) **the mirrored-attribute objection is circular** — it treats a transition cost as steady-state and gates re-opening on 'serde replaced', but adopting facet is how a serde exit would begin; the objection only holds against permanent-coexistence, one of three possible end-states. (2) **Reflection demand is trajectory, not snapshot** — leaves, ownership rows, projection columns, admission schemas, and eventually TUI/CLI/presentation rendering all want shape access; each costs another hand-curated table in the no-facet world, and declared tables carry their own type-mirror drift surface (test-guarded, same structural liability as the attribute mirror). Residual either-way facts from the spike stand: +8.9% clean / +63.6% incremental check cost (measured), 0.x dependency posture, SHAPE capability confirmed working. Interim operative state: the hand table remains active because it is what is merged — that is a fact, not an architecture commitment. #1368 sequencing likewise awaits the same decision.

rjwittams:
Spike complete — recommendation ratified and ruled on #1371 (facet rejected for now; declared-table module is the designed architecture). Closing unmerged as preserved evidence per the spike contract.

PR #1374, opened 2026-08-04
feat: wake landing reconciler from leaves

### LINKED ISSUE #1364: Leaf engine slice 3: ReconcilerWake — Landing/Anchored convoys subscribe their armed leaves
## Parent

#1322 — rulings 5 and 6 (watcher kinds; per-host engine), plus ADR 0028's Landing semantics and ADR 0021's condition-written `Landed`.

## What to build

The **ReconcilerWake** watcher: a convoy entering `Landing` (later `Anchored`) subscribes its armed leaves — for today's hardwired exit, `cr.state ∈ {merged, closed}` for each expected change request — and its lifecycle evaluation re-runs when a leaf fires, instead of fetching integration state on every resync pass. Rows are **derived, never persisted**: recomputed idempotently from convoy phase at daemon boot. `Landing → Landed` becomes event-driven end-to-end: merge observed by the refresher → record write → leaf fires → reconciler evaluates → `Landed` with the reconciler as sole writer, unchanged per ADR 0021.

## Acceptance criteria

- [ ] A Landing convoy with an open CR holds; on observed merge the reconciler writes Landed without a resync interval elapsing (event-driven latency, tested with the in-memory backend + injected refresher)
- [ ] Boot re-derivation: restart with a Landing convoy present → subscriptions recreated, transition still fires
- [ ] Cross-host: authority host's engine drives a convoy whose checkout/CR evidence arrives via replication (depends on #1355)
- [ ] The per-resync integration fetch path is retired (not conditionally kept)

## Blocked by

Slice 2 (cr records/leaves) and #1355 (checkout replication) for the cross-host criterion.

rjwittams:
## Summary

- derive ephemeral reconciler-wake leaf rows for Landing and Anchored convoys from expected change requests
- wake the authoritative convoy reconciler when replicated change-request evidence becomes terminal
- retire per-resync integration fetching from the landing lifecycle path
- cover boot-time rederivation, event-driven landing, and cross-host replicated evidence

## Testing

Closes #1364


rjwittams:
Addressed all four findings in 010419eb:

1. `ReconcilerWake` now subscribes before deriving/spawning rows, so an already-terminal boot observation cannot fire into an empty broadcast channel.
2. Restored the Landing liveness battery against replicated `ChangeRequest` resources, including bounded convergence, quiescence, staleness refresh, degradation holding, and the deliberately re-broken freshness sentinel. `ConvoyReconciler` now takes an injectable clock so the freshness contract is deterministic.
3. Removed the orphaned `convoy_change_requests_settled(_for_checkouts)` API and its dedicated tests.
4. Extracted the authority/local/replica child selection into shared `select_convoy_children`, consumed by both reconciliation and wake-row derivation.

Verified with the repository's exact format, clippy, and sandbox-safe full workspace test commands.

rjwittams:
Fixed in 9ec7df9c. The no-CR fallback now requires `landed=True` evidence with a parseable observation timestamp younger than the checkout integration `LANDING_EVIDENCE_TTL`; the daemon passes that 30-second policy explicitly into `ConvoyReconciler`, whose injected clock makes the decision deterministic. Added `landing_holds_on_stale_vacuous_landed_evidence` at the replacement gate and updated the fresh evidence cases to carry timestamps.

Exact format, clippy, and the sandbox-safe full workspace test suite all pass after the change.

rjwittams:
Acknowledged. The restored enrollment intentionally omits `assert_actuation_drop_recovery` because Landing now consumes stored evidence and produces a status patch; it has no value-shaped Landing actuation that the harness can drop and redeliver. The property remains inapplicable rather than uncovered, so I’m leaving the passing implementation unchanged.

rjwittams:
Fixed in 9636c279. `fetch_dependencies` now lists ChangeRequest records only while the convoy is in `ConvoyPhase::Landing`.

Added `terminal_bound_change_request_settles_checkout_without_own_landed_evidence`: it uses a real expected checkout with fresh `landed=False` and no checkout-level change-request observation, plus a merged bound ChangeRequest record supplied through `with_change_requests`. Settlement therefore depends on the `bound_repository == repo_ref` match; deleting or inverting that check makes the regression fail.

The exact local gates all pass: nightly fmt check, workspace/all-target clippy with warnings denied, and full locked workspace tests.

rjwittams:
Acknowledged. The two-leaf grouping is currently fixed by the terminal-state vocabulary and is not a live defect; structurally encoding subject groups would be a broader API hardening change. Per this final review round, I’m leaving the verified implementation unchanged.

PR #1379, opened 2026-08-04
feat: compose vessel gitconfig fragments

### LINKED ISSUE #1377: vessel_config module: Fragment/composer contract, gitconfig + Forgejo migration
## Parent

Map #1046 · enacts the ratified #1351 contract (its ratification comment is the rationale of record; the prior-art note in docs/research/ has the full semantics).

## What to build

A `vessel_config` module (pure data-and-logic: `Fragment`, `TargetId`, target-typed keys, composers; no store/runtime deps): fragments carry `{target, key, value, order: u16 (sugar FIRST/EARLY/DEFAULT/LATE), priority: u16 (DEFAULT/NORMAL/FORCE), merge: Set|Append|ErrorOnDuplicate, provenance}`. Composer per target: stable-sort `(order, provenance)`, lowest priority wins per key, equal-priority different-valued `Set` is a **loud provisioning error naming both contributors**, `Append` accumulates, output is one staged file per target with provenance comments. Env delivery stays single-valued pointers (no `GIT_CONFIG_COUNT`-style slot schemes, per the PR #1349 ruling).

Migrate the two existing gitconfig writers onto it: `CredentialStore`'s `render_gitconfig` (gh adapter) becomes fragment emission + module composition; the Forgejo adapter's remaining `GIT_CONFIG_*` env entries become fragments in the same composed file — after this PR no adapter writes `GIT_CONFIG_*` env vars at all.

## Acceptance criteria

- [ ] Gh + Forgejo granted together → one staged gitconfig with both helper stanzas, both hosts resolving; regression test extends the existing adapter-seam pair test
- [ ] Equal-priority conflicting `Set` fragments fail provisioning loudly, error naming both contributors and the key
- [ ] `Append` ordering follows `(order, provenance)`; sugar constants covered
- [ ] No `GIT_CONFIG_COUNT/KEY_n/VALUE_n` env vars remain anywhere in delivery (test asserts absence)
- [ ] Preflights unchanged and green under the composed file

## Blocked by

None — can start immediately.

rjwittams:
Closes #1377

## Summary

- add the pure `vessel_config` fragment/composer contract with typed gitconfig keys, ordering and priority sugar, explicit merge policies, and provenance rendering
- fail composition on equal-priority conflicting `Set` values while preserving deterministic `Append` ordering
- migrate GitHub and Forgejo credential helpers to one composed staged gitconfig behind `GIT_CONFIG_GLOBAL`, with no `GIT_CONFIG_COUNT/KEY_n/VALUE_n` delivery
- retain both adapters' existing authentication and Git credential preflights

## Validation

### DIFF
--- a/crates/flotilla-daemon/src/credential.rs
+++ b/crates/flotilla-daemon/src/credential.rs
@@ -281,11 +281,7 @@ impl CredentialStore {
             composed_fragments.extend(new_git_config_fragments);
             let gitconfig = match compose(TargetId::GitConfig, composed_fragments.values().cloned()) {
                 Ok(gitconfig) => gitconfig,
-                Err(error) => {
-                    let (name, adapter, cache_key) = git_config_owner.expect("Git config fragments have an owner");
-                    self.materials.lock().await.remove(&cache_key);
-                    return Err(bounded_adapter_error(&name, &adapter, &format!("compose shared Git config: {error}")));
-                }
+                Err(error) => return Err(format!("compose shared Git config: {error}")),
             };
             if let Err(error) = runner.write_file(Path::new(GIT_CONFIG_PATH), &gitconfig.contents).await {
                 let (name, adapter, cache_key) = git_config_owner.expect("Git config fragments have an owner");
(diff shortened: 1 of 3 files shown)


rjwittams:
Addressed the findings from review comment 5178275647 in 91dfa825:

1. Added coverage for mixed winning merge policies, including the key and both contributors in the error.
2. Made the generated `Fragment` builder the sole construction API and removed the parallel manual constructor/`with_*` methods.
3. Composition failures now surface the composer's contributor-aware error directly and no longer blame or evict an arbitrary credential cache entry.
4. Provenance containing a line break is rejected before rendering, with regression coverage.

Re-ran the exact workspace format, clippy, and locked test gates successfully before pushing.

rjwittams:
Addressed all four findings in 91dfa825; the detailed resolution and validation summary is in the follow-up review-cycle comment.

rjwittams:
Fixed in 63945e28. The gitconfig composer now rejects line breaks in every key component and fragment value before rendering, alongside the existing provenance guard, with separate regression tests for key and value injection attempts. The exact workspace format, clippy, and locked test gates pass locally.

rjwittams:
Confirmed. `TargetKeyMismatch` is intentionally part of the target-dispatch contract for the next target-typed key variant; with only `GitConfig` today it is unreachable, and adding a second `TargetKey` variant will make the exhaustive composer match require the corresponding validation path. No code change needed for this non-blocking observation.

PR #1382, opened 2026-08-04
feat: compose Codex home material fragments

### LINKED ISSUE #1378: Migrate codex-home staging in agent_material onto vessel_config fragments
## Parent

Map #1046 · step 3 of the ratified #1351 migration order.

## What to build

`agent_material`'s codex-home staging becomes fragment emission against the `vessel_config` module's agent-config target(s): the staged files and env pointers it produces today are rendered by the shared composers instead of bespoke code, with provenance comments and the same loud-conflict semantics. No behavior change for a single contributor; the win is that credential adapters and agent material can now safely co-contribute to shared targets.

## Acceptance criteria

- [ ] Codex-home material renders byte-equivalent (or better) via the module; existing agent_material tests pass through the new path
- [ ] A credential adapter and agent_material contributing to one target compose with conflict detection (new test)
- [ ] No bespoke file-staging code remains in agent_material for targets the module owns

## Blocked by

#1377 (the module).

rjwittams:
Closes #1378

## Summary

- add an agent-environment target to `vessel_config`, with typed environment keys, provenance rendering, single-valued delivery, and the shared priority/merge resolver
- migrate Codex login-pool material to emit a `CODEX_HOME` fragment while preserving its writable mount and login-status preflight
- make Codex credential adapters emit the same target so credential and leased material conflicts fail before container creation and name both contributors
- stage the composed, provenance-commented agent environment inside the vessel while passing its resolved values through the existing environment delivery path

## Validation

### DIFF
--- a/crates/flotilla-daemon/src/agent_material.rs
+++ b/crates/flotilla-daemon/src/agent_material.rs
@@ -34,7 +34,6 @@ pub(crate) struct AgentMaterialPreflight {
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub(crate) struct AgentMaterialDelivery {
     pub(crate) mount: ProvisionedMount,
-    pub(crate) fragment: Box<Fragment>,
     pub(crate) preflight: AgentMaterialPreflight,
 }
 
@@ -49,6 +48,7 @@ pub(crate) enum AgentMaterialOutcome {
 trait AgentMaterialAdapter: Send + Sync {
     fn id(&self) -> &'static str;
     fn pool_ref(&self) -> &'static str;
+    fn fragment(&self, environment: &BTreeMap<String, String>) -> Option<Fragment>;
 
     async fn prepare(&self, holder_ref: &ResourceRef, environment: &BTreeMap<String, String>) -> Result<AgentMaterialOutcome, String>;
 }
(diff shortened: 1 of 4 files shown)


rjwittams:
Addressed the three observations in cc1bb881:

- Agent-material adapters now emit their fragments before preparation, so the runtime composes credential and material claims before acquiring a pool lease. A conflict therefore fails without transient lease churn.
- Added direct tests for invalid agent-environment names, NUL-containing values, and forbidden `Append` semantics.
- The provenance-commented staged file now renders `export KEY='value'`, making it safely sourceable while preserving the separately delivered single-valued environment entries.

Focused composition/runtime tests and daemon clippy pass after the changes.

rjwittams:
Fixed in the latest commit. `CredentialStore::vessel_config_fragments` now receives the vessel environment and suppresses its Codex-home fragment when `CODEX_HOME` is explicitly configured, matching agent-material override semantics. The credential regression test now covers both normal fragment emission and explicit-home suppression, so the staged file cannot diverge from the container environment.

rjwittams:
Confirmed. No further code change for the forward-looking note: Codex is currently the only credential consumer in this method, and the explicit-home early return is therefore scoped to the method's entire present output. A future second agent-environment contributor should move that override check into its consumer-specific branch as part of adding the new fragment emission. The current behavior and review findings are fully covered on 62edf910.

PR #1383, opened 2026-08-04
feat: materialize project ops entries

### LINKED ISSUE #1376: Materialize ops-member entries into resources: alias-keyed frontmatter, provenance, refresh
## Parent

Map #1046 · enacts #1142's classes 1/2 and keying rulings (resolution comment is the rationale of record).

## What to build

Materialization of authored operational entries from a project's `ops`-role member(s) into resource definitions: workflow templates first (the existing WorkflowTemplate resources become materialized-with-provenance instead of hand-applied), with verification-command entries (#1086's want) as the second entry kind. Entries carry `repos:` frontmatter naming member aliases (omitted = all `code`-role members); materialization resolves aliases to RepositoryKeys, stamps provenance (ops repo + commit + entry path), and runs only on `project register/refresh` — never a continuous watch. One-way: drifted materialized resources converge back on refresh.

## Acceptance criteria

- [ ] A workflow template authored in an ops member materializes with provenance and is admission-pinnable exactly like a hand-applied one
- [ ] `repos:` keying: an entry scoped to one alias of a two-member project materializes only for that repo; omitted scopes to all code members
- [ ] Refresh converges drift and reports what changed
- [ ] Directory-layout sugar is NOT read by materialization (frontmatter is authoritative) — test asserts a mis-placed file with correct frontmatter still materializes correctly

## Blocked by

#1375 (declaration schema + register/refresh verbs).

rjwittams:
Closes #1376

## Summary

- scan committed ops-member contents for frontmatter-declared workflow and verification-command entries during project registration/refresh
- resolve `repos` aliases to stable repository keys, default omitted scopes to all code members, and stamp source repository/commit/path provenance
- converge workflow and repository drift, remove stale materialized workflows, and report changed definitions from refresh
- document the entry format and cover misleading directory placement, scoped/default targeting, provenance, verification commands, and drift convergence

## Verification

### DIFF
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -87,7 +87,7 @@ use crate::{
     model::{provider_names_from_registry, repo_name, RepoModel},
     ops_entry::{
         parse_operational_entry, OperationalEntryDefinition, MATERIALIZED_PROJECT_ANNOTATION, SOURCE_COMMIT_ANNOTATION,
-        SOURCE_ENTRY_PATH_ANNOTATION, SOURCE_REPOSITORY_ANNOTATION, VERIFICATION_PROVENANCE_ANNOTATION,
+        SOURCE_ENTRY_PATH_ANNOTATION, SOURCE_REPOSITORY_ANNOTATION, VERIFICATION_PROJECT_ANNOTATION, VERIFICATION_PROVENANCE_ANNOTATION,
     },
     path_context::{DaemonHostPath, ExecutionEnvironmentPath},
     placement_policy::reconcile_registered_policy,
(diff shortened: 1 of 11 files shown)


rjwittams:
Addressed all four points in 2b713818.

- Existing hand-applied WorkflowTemplates now cause an ownership collision instead of being adopted, and a regression test verifies the original definition remains untouched.
- Repository verification-command sets now carry a project-owner annotation. A refresh only converges a set owned by that project, rejects manual/cross-project collisions, and leaves unrelated sets alone.
- Git inspection now performs one tree-wide `git grep` for content-declared entry candidates and only opens those candidate blobs, avoiding a subprocess per tracked file without making paths or extensions semantic.
- The integration coverage now exercises stale materialized-workflow deletion as well as the hand-applied collision error path.

Targeted parser and project CLI tests, formatting, and workspace clippy all pass locally.

rjwittams:
Fixed both re-review findings in 1d8b26cb.

- Operational-entry discovery now uses `CommandRunner::run_output`: no-match exit status is accepted only with empty stderr, while spawn/fatal Git failures are surfaced and cannot be mistaken for an empty desired set.
- Refresh now scans Repository resources owned by the project and releases verification commands plus ownership/provenance when a repository is no longer a current code member. The integration test removes a member's code role and verifies that ownership and commands are cleared.

The focused integration test, formatting, and workspace clippy pass locally.

rjwittams:
Fixed in 55ab78c3. Verification-command targets are now validated during alias resolution against the project current code-role members. An explicit ops-only or knowledge-only alias returns an error naming the entry path, command, and offending alias instead of being silently discarded. Added a regression test covering an ops-only target; the new test, neighboring materialization integration test, formatting, and workspace Clippy all pass locally.

rjwittams:
Confirmed: no new blocking findings remain at HEAD. The unavailable-source status signal and additional low-priority error-path coverage are useful follow-up ideas, but neither changes issue #1376 acceptance behavior; this PR deliberately preserves prior materialized state when a source is unavailable and covers the core materialization, drift, deletion, ownership, and role-validation paths.

PR #1387, opened 2026-08-04
feat: propose and observe dispatchable issues

### LINKED ISSUE #1385: Dispatch queue v1: propose and observe, never admit
## What to build (re-scoped 2026-08-04 — Robert: not ready to turn on auto-admission)

The original auto-admission scope is withdrawn: manual dispatch currently performs two judgment functions the machinery cannot yet — **(A) picking workflow and placement per issue** (neither decidable from nor encoded on issues today), and **(B) backpressure from real resources** (a vessel-demand queue with livelock hazards — depended-on vessels unschedulable — needing real design). Both are Quartermaster-design questions; see the Quartermaster brainstorm.

v1 is therefore **propose-and-observe**:

- The reconciler computes the **dispatchable set** — ready-labeled, unblocked (body-convention `## Blocked by` parse, unparseable = blocked), undispatched — per opted-in project, continuously.
- It **surfaces** the set: a `flotilla dispatch queue` read verb (+ fleet-view section), and an attention flag when the set is non-empty past a staleness threshold. It **never admits convoys**.
- It **observes**: when a dispatch happens (governor/human, by any path), and the issue was in the set, record the association — issue, chosen workflow, placement policy, stance, time-from-ready-to-dispatch — as durable observations. This is the "observe real behaviour, then automate" input the Quartermaster design consumes: the recorded distribution of A-decisions is what makes them automatable honestly.
- Kill switch and provenance as originally specified, applied to the proposing/observing machinery.

## Acceptance criteria

- [ ] Ready+unblocked+undispatched issues appear in the queue within a pass; blocked/dispatched drop out
- [ ] Zero convoys are ever created by this machinery (test asserts absence)
- [ ] A manual dispatch of a queued issue records the observation with the A-decision fields
- [ ] Queue staleness raises the attention flag; off switch immediate
- [ ] Behavior tests through injected issue sources + in-memory backend

Relates: the Quartermaster brainstorm (A/B design), #1375/#1380, ADR 0029.

rjwittams:
## What changed

- continuously compute the opted-in Project dispatchable set from ready, unblocked, undispatched issues without admitting convoys
- persist the dispatch queue and raise Project attention once its oldest entry crosses the configured staleness threshold
- add `flotilla dispatch queue` with human and JSON output, and include the queue in the fleet health view
- record immutable `DispatchObservation` resources when a human/governor convoy admits an issue that was queued, capturing workflow, placement policy, stance, dispatch time, and time from first ready observation
- preserve conservative `## Blocked by` parsing, blocker re-observation, kill-switch behavior, and durable provenance

## Safety invariant

The dispatch reconciler has no admission collaborator or admission call. Behavior coverage explicitly asserts that a ready, unblocked issue enters the queue while the Convoy store remains empty.

## Validation

- sandbox-safe locked workspace test suite with `flotilla-daemon/skip-no-sandbox-tests`

Closes #1385


rjwittams:
Fixed both actionable findings in b887c1a0:

- stance preference is now an admission floor, so a missing contained placement is surfaced as Project dispatch attention instead of silently falling back to trusted execution;
- capped passes sort eligible issues oldest-first to prevent newer work from starving older ready issues.

Both paths have regression coverage. The single-attention record and shared pass timestamp remain intentional v1 behavior.

rjwittams:
Acknowledged. The mixed numeric/opaque provider ordering is non-blocking for this change: IssueRef has no cross-provider creation timestamp, so the reconciler cannot promise a true global FIFO order across those identifier domains. This implementation keeps deterministic ordering and provides FIFO behavior for the common single-provider numeric queue covered by the regression test. A cross-provider age key would require a separate resource/provider contract change.

rjwittams:
Fixed both actionable findings in 653e21b6:

- Per-project reconciliation failures are now logged and counted while the pass continues; a regression test covers a broken project alongside a healthy project whose queue still updates.
- The dispatch command now has palette help text, with the affected snapshots updated.

The disable-window behavior is intentional under the rewritten contract: `enabled = false` is the immediate kill switch for both proposing and observing, so a convoy admitted in that race window may deliberately produce no observation.

rjwittams:
Fixed in 7445b1e2. A missing workflow now logs and skips only that convoy/issue observation, allowing queue recomputation and staleness attention to continue. The regression covers a dispatched queued issue with a missing workflow alongside a still-waiting issue, and verifies the waiting queue entry and stale attention both update.

rjwittams:
Confirmed intentional and documented in 74bf6986. In this contract, “dispatched” means the admission decision occurred, regardless of the convoy's later phase. Failed/cancelled/abandoned convoys remain durable evidence of that decision; automatic re-proposal from terminal phase would blur the observation history and risk retry loops, so retry requires explicit human/provider re-triage.

rjwittams:
Fixed in eef55cec. A transient blocker lookup failure now preserves an already-verified queue entry (including its original `ready_observed_at`) while still conservatively excluding a newly seen issue whose blockers cannot be verified. The regression advances past the staleness threshold and verifies both behaviors plus attention.

PR #1388, opened 2026-08-04
feat: explain convoy settlement evidence

### LINKED ISSUE #1386: flotilla convoy <name> explain: print the settlement evaluation as evidence, from any host
## What to build

One diagnostic verb that answers "why is this convoy holding" from data the machinery already has — replacing the log-grep/code-read archaeology that cost hours across the 2026-08-02/04 incidents:

`flotilla convoy <name> explain` prints, for the convoy's current phase:
- Expected checkouts (from work placements + adoptions) vs **observed** federated records, each with its integration conditions (`landed/clean/pushed`), `observed_at`, and freshness verdict against `LANDING_EVIDENCE_TTL`
- Bound/observed ChangeRequest records with their fields and staleness
- Armed subscription rows (leaves, bound literals, watcher kind) and the last leaf firings
- The settlement condition's actual evaluation result and, when holding, **which specific expectation is unmet** (missing record, stale evidence, False condition — named, not inferred)
- Last delivery rung per crew session where recorded

**Must work from any mesh host**: reads resolve via authority + replicas (this absorbs the known CLI replica-visibility gap for the read path — `resource get/list` on a non-authority host returning "not found" for records the daemon demonstrably replicates; fix it at least for the reads this verb needs, filing the general fix separately if it is bigger than this scope).

## Acceptance criteria

- [ ] A held-Landing convoy explains itself naming the unmet expectation (regression scenarios: expected-but-unobserved, stale evidence, False landed, artifact-less claim-exit)
- [ ] Works from a non-authority host against a remote convoy
- [ ] Human-readable default, `--json` for tooling
- [ ] Behavior tests via in-memory + federated two-backend setups

Relates: #1367, #1355, ADR 0029, Bosun runs' resource-read friction (#1321).

rjwittams:
## Summary

- add `flotilla convoy <name> explain` with human-readable and JSON output
- explain the exact settlement evaluation shared with reconciliation, including missing, stale, false, and unknown evidence
- surface federated checkout/change-request provenance, armed subscriptions and last leaf firings, and recorded crew delivery state
- make resource gets and convoy explanations resolve replicated resources on non-authority hosts

## Verification

Closes #1386

### DIFF
--- a/crates/flotilla-commands/src/commands/convoy.rs
+++ b/crates/flotilla-commands/src/commands/convoy.rs
@@ -21,6 +21,8 @@ pub struct ConvoyNoun {
 
 #[derive(Debug, Clone, PartialEq, Eq, Subcommand)]
 pub enum ConvoyVerb {
+    /// Explain why this convoy is holding in its current phase
+    Explain,
     /// Manage the work aboard a convoy's vessels
     Work(ConvoyWorkNoun),
     /// Delete a convoy and tear down its managed resources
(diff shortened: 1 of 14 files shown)


rjwittams:
Addressed all four points in 18e7a05a:

- added a bound-change-request explanation test covering missing, stale, and fresh-open observations, including the `bound` marker and freshness
- introduced `bound_change_request_record_name` so explanation derives the bound record directly instead of depending on leaf ordering
- moved `explain_convoy_internal` to an inherent `InProcessDaemon` method and removed the hidden `DaemonHandle` surface
- removed the unused `EvidenceFreshness::NotApplicable` variant

The pinned formatter, workspace clippy with warnings denied, and full sandbox-safe workspace test suite all pass after these changes.

rjwittams:
Confirmed the collection scan is a real consequence of the missing replica-aware point lookup and is outside this CLI feature's scope. Filed #1390 to add that primitive and restore an indexed/local-first get path before this becomes a hot path: https://github.com/flotilla-org/flotilla/issues/1390

PR #1389, opened 2026-08-04
chore: add Forgejo Darwin CI

rjwittams:
## Summary

- add the canonical Forgejo Actions workflow for Darwin builds and tests
- run it for pull requests, pushes to `main`, and manual dispatches while leaving GitHub Actions unchanged
- use a fully qualified, commit-pinned Forgejo checkout action without persisted credentials
- verify that the ephemeral worker has no code-signing identity, runner registration, or administrative credentials
- isolate daemon socket integration tests from the checkout's live remote by using a temporary local Git origin

## Live lab proof

The revised workflow passed on an ephemeral Tart macOS guest at commit `593b0922`:

https://forgejo.lab.flotilla.work/lab/flotilla/actions/runs/18

The one-job runner unregistered after completion and the guest was destroyed. The temporary lab proof branch was also removed.

## Validation

- live Forgejo workflow: Darwin build and full locked workspace test suite

### DIFF
--- a/.forgejo/workflows/ci.yml
+++ b/.forgejo/workflows/ci.yml
@@ -33,7 +33,7 @@ jobs:
           test ! -e "$HOME/.config/forgejo-runner/.runner"
           test -z "${FLEET_CODESIGN_IDENTITY:-}"
           test -z "${FORGEJO_ADMIN_TOKEN:-}"
-          ! git config --local --get-regexp 'http\..*\.extraheader'
+          ! git --no-pager config --local --get-regexp 'http\..*\.extraheader'
 
       - name: Build binaries
         run: cargo build --locked --bins
--- a/.forgejo/workflows/ci.yml
+++ b/.forgejo/workflows/ci.yml
@@ -12,6 +12,7 @@ permissions:
 
 env:
   CARGO_INCREMENTAL: "0"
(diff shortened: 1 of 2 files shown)


rjwittams:
Addressed in `593b0922`: the workflow now documents that fmt/Clippy remain platform-independent GitHub gates and that one-job guests intentionally do not persist build caches; the test uses the repository's grouped `Path`/`PathBuf` import style; and the codesigning assertion accepts whitespace variation. The revised credential assertion and full suite passed on a fresh ephemeral Darwin guest in Forgejo run 18, after which the runner unregistered and the guest was destroyed.

The lack of a cross-job cache remains deliberate for this first version: preserving the guest-destruction isolation boundary is more important than optimizing a roughly five-minute run. We can introduce a separately scoped external cache later if measured latency warrants it.

rjwittams:
Confirmed. A `git` binary is already a workspace build/test prerequisite rather than a new environmental requirement introduced by this fixture: `crates/flotilla-client/build.rs` invokes Git, and existing tests invoke it across `flotilla-core`, `flotilla-daemon`, and test-support code. The new fixture makes that existing dependency explicit in one more hermetic integration test.

The revised workflow passed on the fresh ephemeral Darwin guest in Forgejo run 18. During validation, run 17 also exposed an unrelated timing-sensitive log-capture assertion; it passed in isolation and on the subsequent unchanged full run, so I tracked that separately as #1393 rather than expanding this PR.

PR #1399, opened 2026-08-04
fix: reclaim squash-merged and force-deleted convoys safely

### LINKED ISSUE #1397: Pushed condition is permanently False for squash-merged branches — terminal reclaim refused forever, checkouts accrete
## What happened

2026-08-04: 25 checkout dirs / **713G** accumulated on feta. Every terminal convoy with a *merged* PR was being reclaim-refused in a loop: `automatic convoy reclaim refused ... Pushed=False (N unpushed commits)` (feta+kiwi daemon logs, one line per convoy per pass). Root cause: the repo squash-merges — the branch's commits never appear upstream as themselves — and the remote branch is deleted on merge, so the Pushed probe (`rev-list` against upstream/origin-HEAD) evaluates **permanently False** for every completed crew checkout. The ADR 0017 gate `(Clean ∧ Pushed ∧ Landed)` then refuses forever, in a warn-loop with no attention surfaced.

Interim: ~20 merged-terminal convoys force-reaped by the governor (2026-08-04).

## Fix

The Pushed condition's intent is "no unpreserved work." Make it CR-aware: when the checkout's bound change request is observed terminal-merged and the checkout's HEAD is reachable from (equal to or an ancestor of) the CR's observed `head_sha` — the observed CR record carries `head_sha` since #1366 — then Pushed is satisfied: everything at HEAD is preserved via the merge. Commits made *after* the merged head remain genuinely unpushed and still refuse. The no-CR path keeps today's upstream-based probe.

Also: repeated automatic reclaim refusal (attempts ≥ N on the same convoy) must raise an **attention flag** rather than warn-looping silently — this accreted for days across two incidents (#1333's ~230G, now 713G) precisely because refusal is invisible.

## Acceptance criteria

- [ ] Squash-merged checkout (HEAD == observed merged head_sha, remote branch deleted) passes the gate and reclaims automatically; a commit made after the merged head still refuses
- [ ] No-CR checkouts keep current semantics
- [ ] Persistent refusal raises an attention condition naming the checkout and reason
- [ ] Regression scenario reproducing the squash-merge shape via replay/injected runners

Relates: ADR 0017/0021, #1333 (prior accretion), #1367 (observed head_sha the fix leans on).

rjwittams:
## Summary

- satisfy `Pushed` from a terminal merged change request when checkout HEAD is preserved by its observed `head_sha`
- raise durable attention after repeated automatic reclaim refusals and flag checkout records whose convoy is missing
- cascade owned presentations, vessels, sessions, and checkouts before forced or ordinary convoy deletion

## Validation

Closes #1397


rjwittams:
Addressed the review findings in 533b89f8:

- Reclaim refusal attempts are now keyed by convoy rather than exact rendered error text, while retaining the latest reason for the attention annotation. The regression test deliberately changes the checkout-specific refusal text between attempts and verifies attention still raises at the threshold.
- The forced-delete cascade test now creates a convoy-targeted HumanGate Demand and asserts that demands are also empty afterward.

The per-checkout ChangeRequest scan is unchanged in this correctness-focused PR; it is a non-blocking optimization opportunity for indexed/cached CR lookup.

rjwittams:
Addressed both follow-up findings in 5b6801af:

- `rebuild_local_projection` now publishes awareness from the original salience-change result after row updates, eliminating the dead second rebuild. The reclaim-refusal test now registers an Awareness subscriber and verifies it receives the demand transition.
- Orphan membership now includes soft-deleting convoys, so attention only fires when the convoy record is actually absent. The orphan regression now verifies the checkout remains non-attention while its convoy has a deletion timestamp.

Focused aggregator tests, exact formatting, and daemon Clippy all pass locally.

PR #1401, opened 2026-08-04
fix: redrive failed clones on new demand

### LINKED ISSUE #1400: fix: Failed Clone resource latches permanently, killing every later convoy for that repo at birth
## What happened

Dogfood evidence (2026-08-04, feta): the shared Clone resource for andamento (`clone-l3emp…`, path `~/dev/flotilla-repos/andamento`) failed once with git's `destination path … already exists and is not an empty directory` — a transient collision with a leftover directory (itself debris from the `--force` cascade-skip defect, see #1397). The clone latched `Failed` and **never re-reconciled**. Every subsequent convoy on that project then failed at birth, re-reporting the stale message — even after the obstructing directory was gone. Observed twice in a row: dir verified absent between attempts, second `convoy start` still failed instantly with the same message, because it merely observed the tombstone. Manual `flotilla resource delete clone <name>` unblocked it; the next dispatch cloned cleanly and went Active.

## Defects

1. **Failed is a tombstone.** A `Failed` Clone never retries — not on new demand (a new convoy referencing it), not when the obstruction disappears. A shared, long-lived resource must not latch a transient environment error terminally; new demand or a path re-probe should re-drive reconcile (backoff is fine, permanence is not).
2. **The original error should be impossible.** `CloneControllerRuntime::clone_and_inspect` stages (`clone_staging_path`) and adopts existing valid clones (`recover_existing_clone`), so git's "destination already exists" at the *target* path should be unreachable. Work out which path actually issued that clone (the message names the bare target, not a staging path) and close the gap — likely a route that calls `clone_repo` straight at the target, or a reconcile double-fire racing itself (the `CloneActuator` fresh-clone path has no in-flight dedup, unlike `CloneFlights`).
3. **Convoy status surfaces the stale message with no hint it is historical.** The convoy reported the clone's old failure text as if it just happened, which sent diagnosis down the wrong path. If a convoy fails because a dependency resource is already Failed, the status should say so ("clone <name> is Failed since <time>") rather than replaying its message verbatim.

## Acceptance

- A Clone that failed transiently recovers without manual record deletion once the obstruction is gone (new-demand retry or re-probe; add a behavior test: Failed clone + path now absent + new checkout referencing it ⇒ clone retries and succeeds).
- The direct-to-target clone route is identified and either staged+adopting or deduped; regression test for the concurrent/dirty-target case.
- Convoy failure caused by an already-Failed dependency names the dependency and the failure's age.

rjwittams:
## Summary

- redrive a shared Clone once when a newly-created worktree Checkout observes an older failure
- record the Clone failure timestamp and surface dependency failures as `clone <name> is Failed since <time>: <detail>`
- remove the unused direct-to-target Clone actuator, leaving the staged, adopting, single-flight daemon runtime as the only shared-clone route
- cover failed-clone recovery, current-attempt failure context, dirty targets, and concurrent clone flights

## Root cause

`ClonePhase::Failed` was terminal, and Vessel reconciliation copied the Clone's old message directly into the child failure. New convoy demand therefore never changed the Clone state and could not distinguish the historical failure from a new attempt. Separately, an unused public actuator still exposed the original direct-to-target clone behavior even though the live daemon route had moved to staged publication and single-flight deduplication.

The Checkout now compares the Clone failure time with its own creation time. An older failure requests one timestamp-guarded retry; a failure produced after that Checkout began becomes a named, timestamped dependency failure.

## Validation

Closes #1400

### DIFF
--- a/crates/flotilla-controllers/src/actuators/mod.rs
+++ b/crates/flotilla-controllers/src/actuators/mod.rs
@@ -9,27 +9,11 @@ use flotilla_core::{
             CONTAINED_DAEMON_REQUIRED_ENV,
         },
         terminal::TerminalPool,
-        vcs::{CloneInspection, CloneProvisioner},
     },
 };
 use flotilla_resources::{
     DockerEnvironmentSpec, EnvironmentMount, EnvironmentMountMode, FreshCloneCheckoutSpec, TerminalSessionSource, TerminalSessionSpec,
 };
-
-pub struct CloneActuator {
-    provisioner: Arc<dyn CloneProvisioner>,
-}
-
-impl CloneActuator {
-    pub fn new(provisioner: Arc<dyn CloneProvisioner>) -> Self {
-        Self { provisioner }
-    }
-
-    pub async fn clone_and_inspect(&self, repo_url: &str, target_path: &ExecutionEnvironmentPath) -> Result<CloneInspection, String> {
-        self.provisioner.clone_repo(repo_url, target_path).await?;
-        self.provisioner.inspect_clone(target_path).await
-    }
-}
 
 pub struct DockerEnvironmentActuator {
     daemon_socket_path: DaemonHostPath,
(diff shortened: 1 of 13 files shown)


rjwittams:
Thanks for the verification. The exhaustive `CheckoutDeps::RetryClone` arm in the Ready-phase match is intentionally retained so the enum is handled explicitly; no code change is needed for this non-blocking observation.

PR #1405, opened 2026-08-05
fix: keep daemon root and socket identity aligned

### LINKED ISSUE #1404: fix: client-spawned daemon inherits FLOTILLA_ROOT from client env while binding the requested default socket — wrong-root daemon squats the fleet socket
## What happened (2026-08-05, kiwi daemon restart)

The moment the kiwi daemon was stopped for the fleet bump, a client inside the grouping-live-session interactive stack auto-respawned one via connect_or_spawn. The spawned daemon inherited `FLOTILLA_ROOT=~/dev/flotilla.grouping-live` from the client's environment (an intentionally isolated live-dev root) **but bound the default fleet socket** (`~/.config/flotilla/run/flotilla.sock`) it had been asked to spawn on. Result: the fleet socket served an empty store from the wrong root — indistinguishable from store loss until the process env was inspected. Recurred immediately on the next restart (respawn race); resolved only by winning the race with a clean-env daemon.

## Defect

Root and socket are two halves of one identity and were allowed to disagree. connect_or_spawn passes the socket explicitly but lets the store root come from ambient env.

## Expected

- A spawned daemon derives socket AND root from the same source: if the client wants the default socket, the daemon must use the default root (spawn with sanitized env), and a FLOTILLA_ROOT-scoped client must spawn on that root's socket, never the default.
- A daemon binding a socket that does not belong to its root refuses to start, loudly.
- Behavior test: client with FLOTILLA_ROOT set spawning on the default socket path ⇒ refused/redirected, never a wrong-root daemon on the fleet socket.

rjwittams:
## What changed

- define the daemon socket owned by a config root as `<config-dir>/run/flotilla.sock` and reject mismatched root/socket pairs before connecting or spawning
- pass the resolved config directory, state directory, and socket explicitly to detached `flotillad` processes so ambient `FLOTILLA_ROOT` cannot change half of their identity
- apply the same identity validation to direct daemon startup
- cover a root-scoped client attempting to spawn on the default fleet socket, including proving no socket or spawn lock is created

## Validation

- direct `flotillad` smoke test with `FLOTILLA_ROOT` scoped and a mismatched default socket (refused before bind)

Closes #1404

### DIFF
--- a/crates/flotilla-client/src/lib.rs
+++ b/crates/flotilla-client/src/lib.rs
@@ -360,15 +360,12 @@ fn resolve_flotillad_binary() -> Result<PathBuf, String> {
         .ok_or_else(|| format!("failed to locate flotillad next to {}", current.display()))
 }
 
-fn spawn_daemon(config_dir: &Path, config_dir_override: Option<&Path>, socket_override: Option<&Path>) -> Result<(), String> {
+fn spawn_daemon(socket_path: &Path, config_dir: &Path, state_dir: &Path) -> Result<(), String> {
     let daemon_binary = resolve_flotillad_binary()?;
     let mut cmd = std::process::Command::new(&daemon_binary);
-    if let Some(dir) = config_dir_override {
-        cmd.arg("--config-dir").arg(dir);
-    }
-    if let Some(socket) = socket_override {
-        cmd.arg("--socket").arg(socket);
-    }
+    cmd.arg("--config-dir").arg(config_dir);
+    cmd.arg("--state-dir").arg(state_dir);
+    cmd.arg("--socket").arg(socket_path);
     // Detach: own session so Ctrl-C doesn't kill daemon with TUI
     use std::os::unix::process::CommandExt;
     unsafe {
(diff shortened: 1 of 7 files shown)


rjwittams:
Fixed the managed-terminal case. A canonical scoped socket shaped as `<root>/config/run/flotilla.sock` now supplies `<root>/config` and `<root>/state` when the pane does not carry `FLOTILLA_ROOT`, preserving safe self-healing. The XDG fleet socket (`~/.config/flotilla/run/flotilla.sock`) does not match that scoped shape, so a root-scoped client still refuses the incident pairing. I added focused coverage for both outcomes and replaced the duplicate server-side socket formula with `daemon_socket_path`.

I left lexical path comparison and the test harness shape unchanged: canonical production paths are constructed from the same path components before validation, while canonicalizing a not-yet-created socket would not be reliable; the nested test path remains directly beneath the harness's short `/tmp/flt-*` root and within SUN_LEN.

rjwittams:
Fixed both wiring issues in `74f3bc38`.

Scoped-root derivation now happens in the top-level CLI path resolver, before the TUI constructs its `ConfigStore`, event log, daemon-log paths, or reconnect closure. The corrected state directory is also passed into `pm connect`, so local UI/config state and daemon commands share one root.

Derivation is provenance-sensitive: only an ambient `FLOTILLA_DAEMON_SOCKET` may supply a missing root, and only when `--config-dir` was not explicitly provided. `connect_or_spawn`, `flotillad`, and `flotilla daemon` are strict again, so explicitly conflicting root A/root B flags fail loudly. Tests cover ambient scoped derivation, the XDG fleet socket not encoding a root, explicit config preservation, and the original no-socket/no-lock refusal. A direct binary smoke test also confirms conflicting explicit flags exit with the identity-mismatch diagnostic.

rjwittams:
Fixed in `41621ed6`.

`flotilla daemon` now uses a dedicated strict path resolver that considers only explicit `--config-dir` / `--socket` values and `PathPolicy` defaults; it never reads `FLOTILLA_DAEMON_SOCKET`. Ambient socket-based root derivation remains confined to client paths.

I also resolved the optional recomputation note: multi-path consumers now resolve a single `CliPaths` bundle and thread its config, state, and socket paths together. Added focused coverage for strict daemon resolution, and the pinned format, full clippy, and full workspace test gates pass locally.

rjwittams:
The underlying config-only override concern is valid, but config and state are not generally sibling paths in the supported path model: the default `PathPolicy` deliberately resolves them through separate XDG categories (`~/.config/flotilla` and `~/.local/state/flotilla`). Extending the socket/config check to require a sibling state directory, or deriving state from arbitrary `--config-dir` syntax, would reject or misresolve normal XDG identities.

The incident path fixed here remains covered: `FLOTILLA_ROOT` resolves config and state together, scoped ambient sockets recover both together, and spawned daemons receive both resolved paths explicitly. The newly fixed `flotilla daemon` path ignores `FLOTILLA_DAEMON_SOCKET` and uses one strict `PathPolicy` bundle.

I filed #1408 for the distinct pre-existing CLI gap: design an explicit root/state selection so config-only overrides cannot make separate custom daemons share `resources.sqlite` or lifecycle locks, with behavior coverage for that collision.

https://github.com/flotilla-org/flotilla/issues/1408

PR #1406, opened 2026-08-05
fix: fingerprint protocol build identity

### LINKED ISSUE #1403: fix: version-skewed CLI against newer daemon returns silently empty result sets instead of failing loudly
## What happened (2026-08-05 fleet bump)

Feta's `~/.cargo/bin/flotilla` (built 2026-08-04) was missed in the bump's rename-dance and ran against the freshly restarted daemon. `flotilla resource list convoy|clone|checkout` all returned **well-formed, empty result sets** (valid envelope, `records: []`, cursor at resource_version 0) while the store held everything — the fresh `target/debug/flotilla` listed all records. This masqueraded as total store loss and burned significant diagnosis time; combined with a same-day incident where an empty list was real (wrong-root daemon squat), silent-empty is actively dangerous.

We are no-compat by policy — skewed pairs don't need to *work*, they need to *fail loudly*.

## Expected

A protocol identity/version stamp in the handshake (or per-request envelope); mismatch produces a hard error naming both versions ("client built <x> speaks proto <a>; daemon speaks <b> — rebuild/reinstall the CLI"), never a syntactically valid empty answer. A behavior test: mismatched version stamp ⇒ every query/command errors with the mismatch message.

rjwittams:
## Summary

- derive the client/daemon wire generation from both Git revision and a deterministic fingerprint of the compiled workspace inputs
- share the build-identity implementation so both socket endpoints advertise the same generation
- report protocol and build identity mismatches with both sides named and an explicit rebuild/reinstall action
- strengthen mismatch tests around the operator-facing error

## Why

The existing mandatory Hello guard used Git `HEAD` as its wire generation. Two binaries built from divergent dirty or incrementally updated source trees could therefore advertise the same generation and accept one another, allowing version-skewed responses to deserialize into convincing empty result sets.

The source fingerprint makes the advertised generation reflect what was actually compiled, while retaining `FLOTILLA_BUILD_ID` as the explicit identity for packaged builds.

## Validation

- post-commit client/daemon wire-generation equality test
- focused protocol and build mismatch handshake tests

Closes #1403


rjwittams:
Addressed the useful scope note in 338f2176: the helper now explicitly documents that whole-workspace over-invalidation is deliberate under the no-compat policy, because omitting an indirect wire input is the dangerous failure mode. I also removed the dead `tests` directory exclusion.

The two scans remain intentionally dependency-free and small, and `assets/` is a required compiled input (`include_bytes!`), so a missing asset should fail the build rather than silently produce a misleading identity.

rjwittams:
Acknowledged. I’m deliberately leaving the double scan in place: each build script must emit its own `rustc-env`, and moving the helper into a build-dependency crate would share code compilation but would not eliminate the two function calls/scans. Avoiding the second small source-tree read would require shared generated state and invalidation complexity; that is not justified without a measured build-time problem.

No further code changes are needed from this review round.

PR #1407, opened 2026-08-05
fix: reclaim remote convoy resources safely

### LINKED ISSUE #1402: fix: convoy teardown and reclaim are replica-blind after daemon restart; Landed convoys wedge on stale integration evidence
## Evidence (2026-08-05 fleet bump, kiwi+feta+udder restarts)

Five Landed convoys (kiwi-authority) with feta-authority checkouts could not be reaped after the fleet restart, in two distinct ways:

1. **CLI `convoy delete` refused with `missing checkout integration evidence for checkout-<name>`** — the teardown gate's checkout list simply did not contain the feta-authority checkout records. `flotilla resource list checkout` on kiwi showed only kiwi's own checkout, 30+ minutes after peers reconnected (peer connection re-established per log at 12:44:02).
2. **The automatic reclaim loop on the same kiwi daemon COULD see those checkouts** — it warn-looped every minute with `checkout-<name>: Clean, Pushed, Landed evidence is missing or stale`. So the records exist somewhere the daemon's reclaim path reads but the list/delete-gate path does not (or one of the two goes remote). And the staleness never resolves: integration observation runs when the owning convoy is Landing — a convoy already **Landed** gets no re-observation ever, so post-restart evidence ages out and automatic reclaim refuses forever, silently (a warn-loop is not attention).

Workaround used: `convoy delete --force` on kiwi + `flotilla resource delete checkout <name>` per record on feta (authority-side finalizer cascades to disk correctly — DIR-GONE verified). Force-cascade from kiwi alone did NOT remove feta's records or dirs — the #1399 cascade is also blind to records it cannot see.

## Expected

- One answer to "which checkout records exist" — the teardown gate, the reclaim loop, and `resource list` must read the same (replica-inclusive) view; replication invariant per ADR 0029 says every leaf-addressable kind replicates.
- Reclaim of a Landed convoy must be able to re-demand integration observation (or accept durable evidence recorded at settlement) rather than wedge on freshness that nothing will ever refresh.
- Persistent reclaim refusal must raise attention, not warn-loop silently (this was #1397's acceptance; this failure mode escaped it).

## Related cleanup found during diagnosis

Stale `terminal-checkout-cascade-work-coder` TerminalSession on feta warn-looping `provider registry unavailable for environment env-checkout-cascade…` every minute — debris of a torn-down environment; TerminalSession records whose environment is gone should be GC'd, not reconciled forever.

rjwittams:
## Summary

- make teardown gates and resource listing replica-aware by default
- let checkout authorities reclaim managed records for landed, abandoned, or deleting convoys while the convoy finalizer waits for replicated cleanup to converge
- scope managed checkout identities by placement and dispatch generation so redispatches cannot collide with stale same-named records
- garbage-collect terminal sessions whose owning environment has disappeared

## Root cause

Cleanup decisions mixed local-only and federated resource views. That made remote checkout authorities invisible to manual teardown, while name-only managed checkout identities let stale records from an earlier placement collide with a later redispatch. Abandon also had no authority-side reclamation path.

## Validation

- `cargo test -p flotilla-daemon --locked --features skip-no-sandbox-tests`
- `git diff --check`

Closes #1402

### DIFF
--- a/crates/flotilla-controllers/src/reconcilers/checkout.rs
+++ b/crates/flotilla-controllers/src/reconcilers/checkout.rs
@@ -8,8 +8,8 @@ use flotilla_core::checkout_integration::{
 use flotilla_resources::{
     controller::{Actuation, ReconcileOutcome, Reconciler, ReplicaConvoyCheckoutWatch, SecondaryWatch},
     Checkout, CheckoutBranchProvenance, CheckoutIntegrationStatus, CheckoutPhase, CheckoutSpec, CheckoutStatus, CheckoutStatusPatch, Clock,
-    Clone, ClonePhase, Convoy, ConvoyPhase, IntegrationCondition, ReplicaReadResolver, ResourceBackend, ResourceError, ResourceObject,
-    ResourceProvenance, SystemClock, TypedResolver, ACTUATOR_SOURCE_ROOT_ANNOTATION, CONVOY_LABEL,
+    Clone, ClonePhase, Convoy, ConvoyPhase, IntegrationCondition, LifecycleAuthority, ReplicaReadResolver, Resource, ResourceBackend,
+    ResourceError, ResourceObject, ResourceProvenance, SystemClock, TypedResolver, ACTUATOR_SOURCE_ROOT_ANNOTATION, CONVOY_LABEL,
 };
 use tracing::warn;
 
@@ -152,6 +152,7 @@ fn convoy_claims_checkout(convoy: &ResourceObject<Convoy>, checkout_name: &str) 
 
 pub enum CheckoutDeps {
     None,
+    OwnerTerminal,
     Ready { prepared: PreparedCheckout },
     Integration { status: Box<CheckoutIntegrationStatus> },
     RetryClone { clone_name: String, failed_at: DateTime<Utc> },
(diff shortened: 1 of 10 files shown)


rjwittams:
Fixed in `63f803d5`. `Failed` and `Cancelled` convoys now retain the 30-second terminal-evidence refresh cadence, while `Landed` and `Abandoned` continue through the durable authority-side reclaim path. The replicated-authority regression test now covers `Landing`, `Failed`, and `Cancelled`, and the full controller suite plus strict workspace Clippy pass locally.

rjwittams:
CI follow-up: `wedge-regressions` exposed that making replica-inclusive listing the default had removed the explicitly spelled `--include-replicas` form used by integration callers. Fixed in `89b48e18`: default listing remains federated, `--local-only` selects the authority-local view, and `--include-replicas` remains accepted for compatibility. Focused CLI parsing, strict workspace Clippy, pinned formatting, and diff checks pass locally.

rjwittams:
Fixed in `be1d94c3`. `Landed` now retains the 30-second evidence cadence for checkouts that reach the observation path; managed convoy-owned checkouts still return earlier through `OwnerTerminal`, while adopted checkouts continue refreshing the landing evidence consumed by teardown. The replicated-authority regression now uses adopted checkouts and covers `Landing`, `Landed`, `Failed`, and `Cancelled`. Focused tests and strict workspace Clippy pass locally.

rjwittams:
Evaluated: this confirms the prior findings are resolved and raises no new action item. No further code change needed.

rjwittams:
Evaluated: this third-pass review reports no new correctness, security, or coverage concerns. No further code change needed.

Issue #1410, opened 2026-08-06
task: verify contained claude crew end-to-end on udder (governor pre-flight)


rjwittams:
## Re-scope (2026-08-13)

Original body predates #1419 and is stale: the crew image already bakes the claude CLI (pinned `CLAUDE_CODE_VERSION`, image check runs `claude --version`) and #1419 delivered the subscription OAuth credential consumer. What remains is **verification on the target host**, not construction.

## What to build

Run a contained claude-code crew end-to-end **on udder** (the governor placement target, #1440): dispatch a trivial convoy with `--agent claude-code` and a contained stance to udder's docker placement, confirm the crew authenticates via the #1419 OAuth consumer path (no interactive login, no credential material in the image or repo), does one round-trip of real work (e.g. a docs-typo contract), and lands it.

If anything is genuinely missing (docker placement policy for udder, credential grant delivery to udder, image pull), fix or file it by name — that's the residual scope this issue absorbs.

## Acceptance criteria

- [ ] Contained claude crew on udder completes a trivial convoy end-to-end
- [ ] Auth flows through the credential consumer; no material readable inside the container beyond the granted scope
- [ ] Gaps found are fixed here or filed by name
- [ ] Result recorded here; #1440 unblocked or blocked-by updated

rjwittams:
Scope correction after verification (2026-08-06): the provisioning is mostly already done.

- **feta**: claude 2.1.223 at `~/.local/bin/claude` (installed today) + `~/.claude/.credentials.json` present.
- **udder**: claude 2.1.223 + credentials present.
- **Contained crew image** (`flotilla-crew:2026-07-26.1`): already ships `/usr/local/bin/claude` alongside codex.

Remaining actual work:

1. **Daemon discovery check**: the feta/udder daemons were last restarted from non-login ssh shells; if the daemon's PATH lacks `~/.local/bin`, `find_binary("claude")` won't assert the adapter and placement will refuse claude dispatches despite the binary existing. Verify the EnvironmentBag sees claude on both (and fix the restart ritual to export PATH consistently — udder's start line already does, feta's didn't).
2. **Contained credential delivery**: the image has the binary but no auth. Claude subscription material into containers should ride the existing material-pool/CredentialSpec machinery (same shape as the github-app grant), informed by the #1140 config/credential seeding research. Host-local secret class per #1142 — never baked into the image.
[...]

Issue #1394, opened 2026-08-04
Quartermaster: dispatch judgment as a fleet-wide independent — A-decisions, backpressure, and the escape hatch


rjwittams:
## Question

The dispatch-queue re-scope (#1385, 2026-08-04) named the two judgment functions manual dispatch performs that automation cannot yet, and ruled "observe real behaviour, build tools to reduce drudgery, keep the escape hatch." This brainstorm holds the design questions until the observations exist:

- **A — per-dispatch decisions**: workflow and placement/stance selection. If an *Issue* is the dispatched thing, A must become either decidable from the issue (content classification? labels-as-encoding? per-issue ops entries?) or encoded onto it at triage. What does the observed distribution of A-decisions (captured by #1385's observation records) actually look like — near-constant per project, or genuinely per-issue?
- **B — backpressure**: manual dispatch implicitly gates on fleet capacity. Automated admission needs a real **vessel-demand queue** with careful liveness analysis: livelock/deadlock when depended-on vessels are unschedulable (a queued convoy's dependency needs capacity held by other queued work), priority/aging, per-host and per-stance capacity accounting (interacts with #1135's admission free-space floor).
- **The vehicle**: Quartermaster as the first **fleet-wide independent** (PersistentAgent) — consuming #1385's observations, proposing then gradually taking A-decisions and queue admission, always with the human escape hatch. This pushes the independents machinery generally (the Governor/Bosun-first ordering from the meta-agent direction — the Bosun's mechanical half is now substrate, per #1384, which reorders Quartermaster earlier than expected).

Sequencing corrected (Robert, 2026-08-04): grill **now** — the observation machinery needs the grill to define what it observes; data cannot precede the thing that gathers it. #1385's fields are the floor, not the ceiling.

rjwittams:
## Vehicle resolved (2026-08-04): the independent is a standing convoy

[...]

PR #1411, opened 2026-08-06
feat: dispatch-time agent selection for convoy crews

### LINKED ISSUE #1409: feat: dispatch-time agent selection — claude-code crews (opus/sonnet) as first-class convoy option
## Why now

Codex weekly subscription usage hit 8% remaining (resets Saturday); the fleet currently cannot shift coding load to Claude even though the `claude-code` adapter already exists and is proven end-to-end for `review`/`code-review` capabilities (autonomy args, per-invocation `--settings` hook injection, `--model` plumbing all present in `agent_adapter.rs`). Longer term the *chooser* is Quartermaster-shaped (#1394); this issue is the mechanism it will actuate.

## What to build

Dispatch-time agent requirement override, keeping capability as the template vocabulary:

1. Convoy spec gains an optional `agent_requirements: {<capability>: {adapter, model?}}` override map. Workflow templates keep their capability-only `Selector` — templates never name harnesses (existing test states this invariant).
2. All `CapabilityTable::seeded().resolve(...)` call sites (runtime.rs ×3, in_process.rs, vessel.rs) go through one resolver that consults the convoy's override map first, then the seeded table. No behavior change when no override is present.
3. CLI: `flotilla convoy start --agent <adapter>[:<model>]` applies the override to the `code`/`coding` capability (the common case); `--agent <capability>=<adapter>[:<model>]` (repeatable) for the general form. `--agent claude-code:opus` and `--agent claude-code:sonnet` are the target invocations.
4. Placement admission must evaluate adapter availability against the *effective* requirement (override included), so a claude dispatch refuses feta/udder today with the existing clear per-placement reason rather than failing at launch.
5. Model rides the existing `AgentRequirement.model` → `--model` arg path.

## Acceptance

- `flotilla convoy start --project flotilla --issue N --agent claude-code:opus` places on a host with the claude binary (kiwi today), launches claude with `--dangerously-skip-permissions --model opus`, hooks fire phase/attention, crew completes normally.
- Omitting `--agent` behaves exactly as today (codex).
- A claude dispatch targeting a placement without the adapter is refused at admission with the adapter named.
- Behavior tests for resolver precedence (override > seeded) and placement refusal on effective requirement.

## Explicitly out of scope (tracked separately)

- Installing/authenticating claude on feta/udder and in the contained image (host provisioning, needs human-held login material — follow-up).
- Automatic adapter/model choice from usage budgets (#1394 Quartermaster input).

rjwittams:
Closes #1409. Hand-built by the governor (the assigned crew stalled at zero progress on the exhausted codex weekly limit — this PR is the mechanism that lets future dispatches route around exactly that).

## Shape

- **`Selector` gains optional `adapter`/`model`** (`workflow_template.rs`). Templates never set them — `apply_agent_overrides` writes them into the convoy's workflow snapshot at admission, so placement validation, the vessel reconciler, and all three terminal-launch sites read the effective requirement from the selector itself with no convoy-state threading. Loud on a duplicate capability or one the workflow doesn't carry (the error names the workflow's actual agent capabilities).
- **`CapabilityTable::resolve_selector`**: an adapter override wins outright and deliberately does not inherit the seeded table's model (a model name only means something against the harness it was chosen for); a model-only override reskins the table's adapter; no override = today's behavior, bit-for-bit. With an explicit adapter the capability is template vocabulary only; without one, unknown capabilities stay loud.
- **CLI**: `flotilla convoy start --agent claude-code:opus` (bare form applies to `code`), `--agent review=codex` general form, repeatable; malformed forms rejected at parse. Rendered back canonically by the Display impl for round-trip fidelity.
- **Placement admission evaluates the effective adapter** — covered in both directions: a claude-only host admits an overridden workflow, and an override toward an absent adapter refuses naming the overridden adapter, not the template default.
- **Mechanical ripple**: the selector growth tipped `TerminalSessionSource::Agent` over clippy's large-variant threshold; `context` is now `Box<TerminalCrewContext>`.

## Verification

- `cargo +nightly-2026-03-12 fmt --check` clean; `cargo clippy --workspace --all-targets --locked -- -D warnings` clean; `cargo test --workspace --locked`: 4028 passed, 0 failed.
- New tests: resolver precedence matrix (`agent_adapter.rs`), override rewrite + loud-miss + duplicate (`in_process/tests.rs`), placement admission both directions (`in_process/tests.rs`), CLI parse forms + malformed rejection + round-trip (`convoy.rs`).

## Follow-ups (tracked)

- #1410: contained-image claude credential delivery + daemon-PATH discovery verification on feta/udder; e2e `--agent claude-code:sonnet` dispatch once this merges.
- #1394: Quartermaster consumes this as its actuation surface.


rjwittams:
Review round processed: took the cosmetic nit despite the "not worth it" — an empty capability list after a colon reads as a rendering bug, and the loud-error convention here favors saying "this workflow has no agent crew to override" outright (`fd4fa874`). Everything else in the review confirmed the intended design (admission ordering ahead of auto-placement scoring was indeed the load-bearing property). Checks re-running on the new head.

rjwittams:
Second-round finding confirmed and fixed in `latest`: `Arg::Literal` is indeed raw shell at the current depth, and `--agent` made the model field an untrusted entry point into that trust boundary. Three layers now:

1. **Sink**: `Arg::Quoted(model)` at launch — plus a sink-level test asserting a metacharacter model renders inside single quotes even if upstream validation regressed.
2. **Protocol boundary**: `apply_agent_overrides` validates adapter and model against `[A-Za-z0-9._-]+` — this is the load-bearing check, since arbitrary protocol clients (not just the charset-validated CLI) can send intents.
3. **CLI parse**: same charset, so a typo'd shell fragment fails locally with a clear message; malformed-case test extended with `; rm -rf ~`, `$(whoami)`, and embedded-space forms.

Also verified the rejected-override path leaves the workflow spec untouched (asserted in the new admission test).

rjwittams:
Third review round: nothing to action. The pass independently re-verified the three-layer injection fix (`Arg::Quoted` at the sink, charset validation at the protocol boundary in `apply_agent_overrides`, charset validation at CLI parse) and the selector wiring, and found no new issues. Its observation that `adapter` needs no shell-safety treatment — it is only ever a `registry.agent_adapters` map key, never spliced into a command string — matches the code; the charset check on it stays as consistency, not as an injection guard.

## The red `Test` check is a pre-existing bug, not this PR

`runtime::tests::passing_teardown_gate_removes_the_managed_worktree_path` fails here and failed again on a clean `gh run rerun --failed`. I chased it rather than re-rolling the dice, and it is not caused by this branch. Filed as #1413.

Reproduction on this machine, full-suite runs (`cargo test -p flotilla-daemon --lib --locked`) — it passes reliably when run alone, and needs the parallel suite to lose the race:

| Commit | Result |
|---|---|
| `88c44bc1` (#1406, predates #1407) | 2 of 4 runs failed |
| `854d1b9d` (`main`) | 2 of 3 runs failed |
| `6253fa99` (this branch) | 3 of 3 runs failed |

Instrumenting the gate condition shows the same state on every poll until the deadline:

```
TEARDOWNGATE landed=true ws_empty=false ws=["convoy-a-implement"] path_gone=true
```

The convoy lands and the worktree is removed, but the `Vessel` is never deleted. Raising the deadline from 5s to 60s does not help, so this is a stall rather than a slow runner. Root cause, verified against the code: `CheckoutReconciler` deletes the `Checkout` as soon as the convoy is `Landed` (`checkout.rs:185-188`, short-circuiting to `OwnerTerminal` at `checkout.rs:202-207`), while `ConvoyReconciler` computes `reclaim_eligible` from the *pre-patch* status (`convoy/reconcile.rs:498`) and so can only reclaim on a later pass — by which point the gate rejects the now-absent checkout as "missing checkout integration evidence" (`in_process.rs:6910-6920`). With `requeue_after: None` (`reconcile.rs:527`), it never recovers. That is a real production leak of `Vessel`/`Presentation` resources after a convoy lands, not only a test artifact.

Fixing it belongs in #1413, not in an agent-selection PR.

Rest of the gates on this head, run locally: `cargo +nightly-2026-03-12 fmt --check` clean, `cargo clippy --workspace --all-targets --locked -- -D warnings` clean, and `cargo test --workspace --locked --no-fail-fast` fails exactly one target — `flotilla-daemon --lib`, on that one test. CI never got past it, so nothing after `flotilla-daemon` had been exercised on the runner; locally everything downstream passes.

No code changes pushed this round; the head is still `6253fa99`.

rjwittams:
A third `Test` attempt on the same head (`6253fa99`) went green, so all six checks now pass and the branch is unblocked.

To be precise about what that means: the race was won this time, not fixed. `runtime::tests::passing_teardown_gate_removes_the_managed_worktree_path` failed on the first two CI attempts and still reproduces locally on `main` and on commits predating both this PR and #1407, so #1413 remains a live bug — a convoy that lands can permanently leak its `Vessel` and `Presentation` resources. A green run here is a coin flip landing the right way, and this test should be expected to redden other PRs until #1413 is fixed.

Nothing was pushed for this; the head is unchanged.

PR #1418, opened 2026-08-06
fix: convoy reclaim accepts sanctioned checkout absence and retries refusals

### LINKED ISSUE #1413: Convoy reclaim deadlocks when the checkout controller deletes the Checkout first, leaking Vessels
Discovered while shepherding #1411. **Not caused by that PR** — reproduced on `main` and on ancestors (see Evidence).

## Symptom

`runtime::tests::passing_teardown_gate_removes_the_managed_worktree_path` (`crates/flotilla-daemon/src/runtime.rs`) fails intermittently with `timed out waiting for condition` at the teardown-gate wait. Instrumenting the gate condition shows the same state on every poll until the deadline:

```
TEARDOWNGATE landed=true ws_empty=false ws=["convoy-a-implement"] path_gone=true
```

The convoy reaches `Landed`, the managed worktree **is** removed from disk, but the `Vessel` `convoy-a-implement` is never deleted. Raising the deadline from 5s to 60s does not help — the resource lingers indefinitely, so this is a stall, not a slow machine.

## Root cause

Two controllers can delete the `Checkout`, and only one consults the reclaim gate.

1. `CheckoutReconciler` is ungated. `convoy_authorizes_checkout_reclaim` (`crates/flotilla-controllers/src/reconcilers/checkout.rs:185-188`) returns true as soon as the convoy is `Landed`, and `fetch_dependencies` short-circuits to `CheckoutDeps::OwnerTerminal` (`checkout.rs:202-207`), emitting `Actuation::DeleteCheckout` (`checkout.rs:329-331`). Its finalizer removes the worktree. It is woken by the same convoy status write that flips the phase to `Landed`.

2. `ConvoyReconciler` cannot reclaim on that same pass. `reclaim_eligible` is computed from the **pre-patch** object (`crates/flotilla-resources/src/convoy/reconcile.rs:498`, requires `obj.status.phase.is_terminal()`), while `cleanup_plan` evaluates the **predicted** post-patch status (`reconcile.rs:1201-1204`). Reclaim therefore always needs a *subsequent* pass.

3. On that subsequent pass the `Checkout` may already be gone, and the gate treats absence as refusal: `expected_checkout_refs` is derived from convoy work placement, not live objects, so any expected checkout missing from `checkout_list` produces a hard error, `"missing checkout integration evidence for …"` (`crates/flotilla-core/src/in_process.rs:6910-6920`).

`reclaim_eligible` is then false on every later resync forever, so `Actuation::DeleteVessel` (`reconcile.rs:1228-1236`) is never emitted. `ConvoyReconciler::reconcile` returns `requeue_after: None` (`reconcile.rs:527`), so nothing else retries it.

A second leg of the same class: once the convoy is `Landed`, the `OwnerTerminal` short-circuit at `checkout.rs:202-207` returns before the integration-refresh block (`checkout.rs:209-225`), so integration evidence is never refreshed again. If reclaim does not succeed within `LANDING_EVIDENCE_TTL` (30s, `crates/flotilla-core/src/checkout_integration.rs:18`), the freshness check (`in_process.rs:6934-6941`) refuses permanently too.

This is a production defect, not only a test artifact: a convoy that lands can permanently leak its `Vessel` (and `Presentation`) resources.

## What needs to change

Both of these look necessary:

- Treat an absent expected checkout as **already reclaimed** rather than as missing evidence, or hold the checkout until the convoy's reclaim gate has passed. The current gate makes the safe outcome unreachable once the checkout is collected.
- Close the ordering race so the checkout is not deleted out from under the gate — the checkout controller's `OwnerTerminal` path should not destroy the only evidence the gate accepts.

Within a single convoy pass the ordering is already correct (Vessel before Checkout, `reconcile.rs:1249-1253`); the defect is cross-controller.

## Evidence

Full-suite runs (`cargo test -p flotilla-daemon --lib --locked`) on this machine. The test passes reliably when run alone; it needs the parallel suite to lose the race.

| Commit | Result |
|---|---|
| `88c44bc1` (#1406, predates #1407) | 2 of 4 runs failed |
| `854d1b9d` (`main`, #1407) | 2 of 3 runs failed |
| `6253fa99` (#1411 branch) | 3 of 3 runs failed |

Also failed twice on CI for #1411, including on a clean `gh run rerun --failed`.

Since it reproduces before #1411 and before #1407, and #1411 touches only selector resolution and `Box<TerminalCrewContext>`, it is not a regression from either.

rjwittams:
## Summary

Fixes the cross-controller reclaim deadlock where `CheckoutReconciler`'s authorized `OwnerTerminal` cascade deletes the `Checkout` before the convoy's own reclaim pass, and the teardown gate then refuses forever with `missing checkout integration evidence`, permanently leaking the convoy's `Vessel` and `Presentation` resources.

### Design cut: one sanction, defined once

Rather than gating the checkout authority on the convoy's evidence gate (which would reintroduce #1402's cross-host wedges) or sequencing the two controllers (fragile distributed ordering), the sanction for collecting a managed checkout is now a single shared predicate in `flotilla-resources` — `convoy_sanctions_checkout_reclaim`: the convoy is being deleted, `Landed` (the world terminal fired, per ADR 0028), or `Abandoned` (explicit override).

- The checkout reconciler's `OwnerTerminal` path (from #1407) now derives from that predicate instead of its own private copy.
- The convoy teardown gate consumes the same predicate from the other side: an expected checkout that is **absent** — or already carrying a deletion timestamp — while the sanction holds is evidence of *completed reclaim*, never missing evidence. Outside the sanction (e.g. `Failed`, `Cancelled`, or a live convoy), absence stays a hard refusal.

This also fixes the operator-facing leg of #1402: `convoy delete` on a Landed convoy no longer refuses after the checkout authority already collected its record.

### Refused reclaims retry

`ConvoyReconciler::reconcile` returned `requeue_after: None`, so a transient refusal (integration evidence mid-refresh, a cascade in flight) stalled until the next full resync — or forever. A refused reclaim on a terminal convoy now requeues at the landing-evidence staleness horizon (`landing_evidence_stale_after`, the freshness the gate is waiting on). A successful or fully reclaimed pass does not requeue, so retained terminal convoys don't hot-loop.

### Regression tests

- `landed_convoy_reclaim_accepts_sanctioned_checkout_absence` — gate level: absence passes for `Landed`, a deleting checkout passes, absence still refuses for `Failed`.
- `convoy_reclaims_vessels_after_checkout_authority_collected_the_checkout_first` — pins the race itself: checkout gone before the convoy reclaim pass ⇒ `DeleteVessel` still emitted (fails on `main` without the fix).
- `refused_reclaim_requeues_at_the_evidence_staleness_horizon` — pins the retry leg.

### Validation

Reproduction confirmed without the fix on this machine (full parallel `cargo test -p flotilla-daemon --lib --locked`: `passing_teardown_gate_removes_the_managed_worktree_path` FAILED). With the fix, ten consecutive clean full-suite runs (five pre-rebase, five on the final tree rebased onto #1411):

| Run (final tree) | Result |
|---|---|
| 1 | ok, 416 passed |
| 2 | ok, 416 passed |
| 3 | ok, 416 passed |
| 4 | ok, 416 passed |
| 5 | ok, 416 passed |

Gates: `cargo +nightly-2026-03-12 fmt --check`, `cargo clippy --workspace --all-targets --locked -- -D warnings`, `cargo test --workspace --locked` — all clean.

Fixes #1413. Unblocks the red Test check that was hitting #1411 (this flake reproduced on `main` and its ancestors; #1411 has since merged and this branch is rebased on it).

### DIFF
--- a/crates/flotilla-controllers/src/reconcilers/checkout.rs
+++ b/crates/flotilla-controllers/src/reconcilers/checkout.rs
@@ -7,9 +7,10 @@ use flotilla_core::checkout_integration::{
 };
 use flotilla_resources::{
     controller::{Actuation, ReconcileOutcome, Reconciler, ReplicaConvoyCheckoutWatch, SecondaryWatch},
-    Checkout, CheckoutBranchProvenance, CheckoutIntegrationStatus, CheckoutPhase, CheckoutSpec, CheckoutStatus, CheckoutStatusPatch, Clock,
-    Clone, ClonePhase, Convoy, ConvoyPhase, IntegrationCondition, LifecycleAuthority, ReplicaReadResolver, Resource, ResourceBackend,
-    ResourceError, ResourceObject, ResourceProvenance, SystemClock, TypedResolver, ACTUATOR_SOURCE_ROOT_ANNOTATION, CONVOY_LABEL,
+    convoy_sanctions_checkout_reclaim, Checkout, CheckoutBranchProvenance, CheckoutIntegrationStatus, CheckoutPhase, CheckoutSpec,
+    CheckoutStatus, CheckoutStatusPatch, Clock, Clone, ClonePhase, Convoy, ConvoyPhase, IntegrationCondition, LifecycleAuthority,
+    ReplicaReadResolver, Resource, ResourceBackend, ResourceError, ResourceObject, ResourceProvenance, SystemClock, TypedResolver,
+    ACTUATOR_SOURCE_ROOT_ANNOTATION, CONVOY_LABEL,
 };
 use tracing::warn;
 
@@ -182,11 +183,6 @@ fn convoy_needs_terminal_evidence(convoy: Option<&ResourceObject<Convoy>>) -> bo
     })
 }
 
-fn convoy_authorizes_checkout_reclaim(convoy: &ResourceObject<Convoy>) -> bool {
-    convoy.metadata.deletion_timestamp.is_some()
-        || convoy.status.as_ref().is_some_and(|status| matches!(status.phase, ConvoyPhase::Landed | ConvoyPhase::Abandoned))
-}
-
 impl<R> Reconciler for CheckoutReconciler<R>
 where
     R: CheckoutRuntime + 'static,
@@ -201,7 +197,7 @@ where
         let convoy = self.owning_convoy(obj).await?;
         if lifecycle_authority == Some(LifecycleAuthority::Managed)
             && has_convoy_owner
-            && convoy.as_ref().is_some_and(convoy_authorizes_checkout_reclaim)
+            && convoy.as_ref().is_some_and(convoy_sanctions_checkout_reclaim)
         {
             return Ok(CheckoutDeps::OwnerTerminal);
         }
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -6907,12 +6907,20 @@ impl InProcessDaemon {
         if expected.is_empty() {
             return Ok(());
         }
+        // Once the convoy sanctions checkout reclaim (Landed, or the convoy is
+        // being deleted), the checkout authority's `OwnerTerminal` cascade may
+        // legitimately collect an expected checkout before this gate runs. An
+        // absent or already-deleting expected checkout under that sanction is
+        // evidence of completed reclaim, not missing evidence — refusing here
+        // would wedge vessel reclaim forever on a deletion the substrate itself
+        // authorized. Outside the sanction, absence stays a hard refusal.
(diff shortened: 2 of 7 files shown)


rjwittams:
Acknowledging the review — no changes requested, nothing to address. One clarification worth recording on the behavioral note: the indefinite requeue for unsanctioned terminal refusals (`Failed`/`Cancelled` with a real unreclaimed checkout) is intentional, precisely so the existing `RECLAIM_REFUSAL_ATTENTION_AFTER` counter reaches its threshold and raises a persistent `Demand` instead of the refusal hiding between resyncs — the #1397/#1402 acceptance that persistent refusal must raise attention, not warn-loop silently.

All six checks are green, including `wedge-regressions` and `hub-spoke`.

PR #1419, opened 2026-08-06
feat: claude subscription OAuth credential consumer

### LINKED ISSUE #1416: feat: claude subscription OAuth credential consumer (CLAUDE_CODE_OAUTH_TOKEN) for crew launches
Follow-up to the first claude-crew e2e (#1414/#1415) which fell through to ambient `~/.claude` login and hit "Login expired". Crews must not depend on host login freshness — login material is meant to be delivered non-interactively per launch, per the codex precedent.

## Codex precedent (existing, `credential.rs` `CredentialConsumer::Codex`)

Material = API key → per-credential `CODEX_HOME` slot at `/run/flotilla/credentials/<name>/codex`, transformed via `codex login --with-api-key`, verified (`codex login status` + API preflight), delivered as a `CODEX_HOME` env fragment.

## What to build

A claude **subscription OAuth** path (the existing `CredentialConsumer::Claude` takes an API key → `ANTHROPIC_API_KEY`, which bills metered API, not the Max subscription — keep it, but it's the wrong material for crews):

- New consumer variant (e.g. `ClaudeOauth`): material is a long-lived OAuth token minted once by the human via `claude setup-token` (interactive, out of band; host-local secret per the #1142 material classes).
- Delivery: `CLAUDE_CODE_OAUTH_TOKEN=<material>` env fragment — no transformation step needed. Optionally a per-credential `CLAUDE_CONFIG_DIR` slot for settings isolation, mirroring the codex slot pattern (decide during implementation whether the isolation is needed for v1 given `--settings` already isolates hooks).
- Non-interactive preflight so a dead token fails at prep, not as a silent "Login expired" pane (whatever claude exposes headlessly — e.g. a trivial `claude -p` invocation under the token — and record the account email the token belongs to).
- CredentialSpec + grant wiring identical in shape to `github-crew-pr`/`github-contained-pr`; works for trusted host-direct first, contained image second (binary already present in `flotilla-crew` image).

## Direction to keep in view (#1394)

Credentials should end up **account-identity indexed** — the usage-observer (#1412) keys `usage/<account>`, and the chooser will want "this launch burns account X" as a first-class link. v1 can carry the account email in the CredentialSpec; don't bake in anything that prevents multiple claude accounts side by side (the codex slot pattern already generalizes).

## Acceptance

- A crew dispatched `--agent claude-code:sonnet` on a host whose ambient `~/.claude` login is expired/absent launches and works, using the delivered token.
- Dead token ⇒ loud prep failure naming the credential, before any session exists.
- Two claude credentials (different accounts) can coexist and be granted independently.

rjwittams:
Fixes #1416

Adds `CredentialConsumer::ClaudeOauth { account_email }` — a claude subscription OAuth credential consumer whose material is a long-lived token minted once, out of band, via `claude setup-token`. The existing `Claude` API-key consumer (`ANTHROPIC_API_KEY`, metered API billing) is untouched; this is a sibling for subscription-backed crews.

## Delivery

- `CLAUDE_CODE_OAUTH_TOKEN=<material>` per-process env — no login transformation exists or is needed (ADR 0022 amendment: non-rotating material, one mint serves N crews).
- A per-credential `CLAUDE_CONFIG_DIR` slot at `/run/flotilla/credentials/<name>/claude`, mirroring the codex `CODEX_HOME` slot: delivered in the prepared env and as a vessel-config fragment for contained provisioning (suppressed when the environment pins `CLAUDE_CONFIG_DIR` explicitly, same as the codex suppression).

## Design decision: config-dir slot included in v1

The crew adapter's `--settings` overlay isolates hooks, but it does not isolate:

- the **per-config-dir singleton supervisor**, which captures its environment from the first shell that uses it — a shared dir would serve crew A's dispatches with crew B's credentials;
- the mutable **`.claude.json`** blob (lost writes observed in practice — docs/research/2026-07-28-multi-crew-agent-config-seeding.md);
- an ambient **`apiKeyHelper`**, whose documented precedence outranks `CLAUDE_CODE_OAUTH_TOKEN` and would silently override the delivered identity (and mask a dead token during preflight).

The slot is keyed by credential name, so two claude accounts coexist exactly like two codex homes — nothing precludes side-by-side accounts (#1394 direction; the spec carries `account_email` for the future account-identity index).

## Preflight choice: trivial `claude -p` under the token

There is no documented headless status command for claude and no documented HTTP endpoint that accepts subscription OAuth bearers (the `/v1/models` x-api-key probe used by the API-key adapter takes API keys only), so the cheapest *reliable* probe is the CLI itself, on exactly the path the crew will use. The probe runs with `ANTHROPIC_API_KEY`/`ANTHROPIC_AUTH_TOKEN` unset and `CLAUDE_CONFIG_DIR` pointed at the empty per-credential slot, so nothing that outranks the token can mask a dead one (the stored ambient login ranks *below* `CLAUDE_CODE_OAUTH_TOKEN` and cannot). Material travels via stdin, never argv. A dead token fails preparation with a bounded error naming the credential and adapter, material redacted, before any session exists. Cost: one minimal model request per environment preparation, cached by the existing prepared-set for static lifecycles.

## Loud failure on same-slot mixes

`Claude` and `ClaudeOauth` share the `claude` delivery slot: `ANTHROPIC_API_KEY` outranks `CLAUDE_CODE_OAUTH_TOKEN`, so granting both to one environment would silently drop the OAuth identity — it now fails preparation loudly instead, as does granting two OAuth credentials to one environment.

## Tests

Behavior tests through the injected `CommandRunner` (no real subprocesses):

- delivered env pair, fragment emission/suppression, mkdir + stdin preflight, material never in argv;
- two accounts prepare side by side in separate environments with distinct config dirs; same-environment mixes (oauth+oauth, oauth+api-key) fail loudly;
- dead token: bounded failure naming `credential`/`adapter`, `[redacted]` material, failure not cached as prepared (preflight re-runs);
- resources: serde shape (`{"adapter":"claude-oauth","account_email":…}`) and shared delivery slot.

Gates: `cargo +nightly-2026-03-12 fmt --check`, `cargo clippy --workspace --all-targets --locked -- -D warnings`, `cargo test --workspace --locked` — all green locally.

### DIFF
--- a/crates/flotilla-daemon/src/credential.rs
+++ b/crates/flotilla-daemon/src/credential.rs
@@ -162,14 +162,15 @@ impl CredentialStore {
         credential_refs: &BTreeSet<String>,
         environment: &BTreeMap<String, String>,
     ) -> Result<Vec<Fragment>, String> {
-        if environment.contains_key("CODEX_HOME") {
-            return Ok(Vec::new());
-        }
         let mut fragments = Vec::new();
         for name in credential_refs {
             let spec = self.spec(name).await?;
-            if matches!(spec.consumer, CredentialConsumer::Codex) {
-                fragments.push(codex_home_fragment(name));
+            match spec.consumer {
+                CredentialConsumer::Codex if !environment.contains_key("CODEX_HOME") => fragments.push(codex_home_fragment(name)),
+                CredentialConsumer::ClaudeOauth { .. } if !environment.contains_key("CLAUDE_CONFIG_DIR") => {
+                    fragments.push(claude_config_dir_fragment(name))
+                }
+                _ => {}
             }
         }
         Ok(fragments)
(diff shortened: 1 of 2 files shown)


rjwittams:
Thanks for the thorough pass.

On nit 1 (comment verbosity): keeping the full rationale in the source is deliberate — the issue (#1416) asked for the config-dir-slot decision and the preflight choice to be recorded "in code comment + PR body" explicitly, and PR bodies don't travel with the code. The two blocks are the durable record of *why* the slot exists and why the probe is shaped the way it is (unset ordering, precedence masking); trimming them to a doc pointer would push the load-bearing argument out of the file where the next editor needs it. If the reasoning changes, the comment is exactly the thing that should change with it.

On nit 2: agreed it's real and pre-existing — the crew-launch environment scrub for env vars that outrank the delivered credential applies equally to the existing `Claude`/`Codex` adapters and belongs to the uncontained-vessel env-allowlist work (ADR 0022 "backstop"), not this diff.

rjwittams:
Ruling on the preflight's credit accounting (raised by Robert): since June 15 2026, `claude -p`/headless invocations meter against a separate Agent SDK credit pool, not the subscription's interactive limits — apparently deliberate on Anthropic's side to stop pure wrapper apps riding consumer subscriptions. (Third-party sources; official docs at code.claude.com/docs/en/headless.md have not caught up, but they do confirm `setup-token`/`CLAUDE_CODE_OAUTH_TOKEN` remains the sanctioned CI path and that auth precedence is identical between modes.)

Consequences for this PR:

- The preflight stays valid as a credential check — same token, same auth path as the interactive crews.
- Its cost lands on a pool crews never touch (crews are interactive), so it will not eat crew headroom. Acceptable as-is; no zero-cost probe exists (no headless auth-status subcommand, no introspection endpoint).
- One caveat: `--bare` does not read `CLAUDE_CODE_OAUTH_TOKEN` — the preflight correctly avoids it; anyone touching this later should keep avoiding it.
- One false-negative: an exhausted Agent SDK pool fails the preflight while interactive crews would work, and the failure would masquerade as a dead token. Filed as follow-up #1421 (error classification), not a merge blocker.

Verdict: merge as-is.

PR #1424, opened 2026-08-08
fix: scope daemon identity strictness to spawning

### LINKED ISSUE #1417: fix: config-identity strictness belongs at daemon spawn, not client connect; serialize auto-start
Robert's ruling on the #1404/#1405 follow-through (2026-08-06): the enacted client-side check is **too restrictive** — refusing a client whose `--config-dir` differs from the daemon's root makes it hard to try out different configs, which is a legitimate thing to do. The reframe:

> the proper fix should be around the daemon auto startup (ie every client shouldn't be racing on that)… this is about client vs daemon config, as the issue is the daemon being replaced with a weirdly configured one

## What to change

1. **Connect freely.** A client connecting to an *existing* socket may have any config identity — inspecting or driving a daemon of another root is allowed (at most an informational notice). Drop/soften `ensure_daemon_socket_belongs_to_config` on the pure-connect path.
2. **Spawn strictly, and once.** The hazard is exclusively at daemon (re)creation: a client auto-spawning a daemon whose store root comes from the client's ambient env while binding a socket that belongs to a different root (the 2026-08-05 squatter incident — twice). Spawn must (a) derive root and socket from one identity, refusing mismatches loudly (keep #1405's strictness *here*), and (b) not race: all clients wanting the same socket must single-flight the spawn (lockfile or equivalent), with losers connecting to the winner. The kiwi incident was precisely a respawn race after daemon death — first weird-env client wins the socket.
3. Worth considering while there: whether client-triggered auto-spawn is the right lifecycle at all versus explicit daemon start + connect-or-fail for fleet hosts (the bump ritual already treats daemons as explicitly managed); auto-spawn could remain a dev-convenience opt-in.

## Acceptance

- A client with `--config-dir A` can connect to and query a daemon serving root B without error.
- Concurrent `connect_or_spawn` from N clients with mixed environments produces exactly one daemon whose root matches its socket identity; wrong-root candidates lose the race and connect (or error naming the identity mismatch) — never bind.
- Regression test for the squatter shape: ambient FLOTILLA_ROOT client racing a clean client for the default socket ⇒ default-root daemon wins deterministically.

rjwittams:
Closes #1417.

## Summary

- allow clients with a different config identity to connect to an existing daemon
- retain strict socket/config identity validation before daemon auto-spawn
- keep spawn lockfiles persistent so queued and new contenders always flock the same inode
- add regression coverage for mixed-identity races and three-client lock handoff

## Verification

### DIFF
--- a/crates/flotilla-client/src/lib.rs
+++ b/crates/flotilla-client/src/lib.rs
@@ -33,30 +33,18 @@ type SeqMap = std::sync::RwLock<HashMap<StreamKey, u64>>;
 /// connection's subscription.
 type QuerySet = std::sync::RwLock<HashSet<QueryId>>;
 
-/// RAII guard that removes a lock file when dropped.
+/// RAII guard that holds the daemon spawn flock.
 ///
-/// Holds the open file handle (which keeps the OS flock) and removes the
-/// lock file on drop.  The flock is released *before* the path is unlinked
-/// so that concurrent clients racing on the same path always contend on the
-/// same inode — unlinking first would let them create a new file and flock
-/// a different inode, breaking mutual exclusion.
+/// The lock file remains on disk after release so every contender opens the
+/// same inode. Unlinking it during a handoff would let a new client create and
+/// lock a replacement inode while an already-queued client owns the old one.
 struct SpawnLockGuard {
-    file: Option<std::fs::File>,
-    path: PathBuf,
+    _file: std::fs::File,
 }
 
 impl SpawnLockGuard {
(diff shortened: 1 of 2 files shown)


rjwittams:
Verified. I’m keeping the cross-identity connect path silent: #1417 explicitly makes this a legitimate operation and permits “at most” an informational notice, so warning would make an allowed workflow look suspect. The spawn-path mismatch remains loud and actionable. No code change needed.

Issue #1396, opened 2026-08-04
Ensured convoys: declare standing convoys as ops entries; the ensure loop


rjwittams:
## Parent

Enacts the #1394 vehicle ruling (standing convoy = convoy with no exit table, automatically ensured) and the #1384 line's applicative surface. Depends on #1391's amended no-exit semantics.

## What to build

- An **ensure entry** kind for project ops material (class-1, materialized per #1383's machinery): declares a standing convoy — name, workflow template (one with no exit table), placement/stance preferences, and the member-alias scoping the entry applies to where relevant.
- The **ensure loop**: a reconciler that starts the declared convoy when absent and restarts it when its convoy dies (Failed/reaped-without-removal-of-the-declaration), with provenance ("ensured from <ops entry> @ <commit>") and restart backoff. Removing the declaration + refresh reaps the convoy through the normal explicit path.
- **Fleet-level independents by convention, not mechanism**: one project is "the fleet project" (for this fleet: project-map) and its ensure entries are the fleet-scoped independents. No scope concept in the machinery. A presentation hint on the entry (e.g. `presents-as: fleet`) is permitted as an annotation for future UI distinction — surfacing it is #969-territory, not this issue.
- **Known interaction, handle minimally**: sleep inhibition keys on non-terminal convoys, so a standing convoy holds its host awake permanently. Acceptable default for v1; record the quiescent-keyed-inhibition refinement as a linked follow-up rather than solving it here.

## Acceptance criteria

- [ ] An ensure entry materializes; the convoy starts, restarts after failure with backoff, and reaps on declaration removal + refresh
- [ ] The standing convoy never reaches Landed; explain names it as standing (no exit table) rather than stuck
- [ ] Provenance on the convoy names the entry and commit
- [ ] Behavior tests via in-memory backend + injected reconciler harness

## Blocked by

#1391 (no-exit semantics).

PR #1426, opened 2026-08-08
feat: add observed usage resource

### LINKED ISSUE #1412: feat: usage-observer independent — poll CodexBarCLI into usage/<account> observed records
First cut of subscription/token observation (ruled direction on #1394; enacts the probe validated 2026-08-06).

## Shape (Robert's sketch)

An **independent** (standing convoy, no exit table — needs #1396 ensured convoys) placed on the host where CodexBar lives (kiwi — the CLI helper at `/Applications/CodexBar.app/Contents/Helpers/CodexBarCLI` reads that host's auth material), running a simple script every N minutes:

1. `CodexBarCLI usage --format json --provider <codex|claude>` per provider.
2. Translate into flotilla's terms and write `usage/<account>` observed records.

## Translation notes (from the live probe)

- **Windows are a set, not a scalar**: per-window `usedPercent` + `resetsAt` + `windowMinutes`, including scoped pools (`extraRateWindows` like model-specific weeklies, Spark). The 2026-08-06 incident: the harness's in-session banner and the gating window disagreed (8% vs 100%) and the gating window won — never collapse to one number.
- Identity: account email, plan (`loginMethod`), org. Key records by account, not provider (#1394: multi-account is first-class).
- Extras worth carrying: pace projections, provider cost balance, unredeemed reset credits (`codexResetCredits`).
- Records must be leaf-addressable and in the peer replication set (ADR 0029 invariant), so standing rules and dispatch admission anywhere in the fleet can react (e.g. `usage/<account>.weekly.usedPercent > 90` → route `code` to claude-code via the #1411 override surface).

## Slices

1. `Usage` observed resource kind (subject-keyed `usage/<account>`, windows as a list, `observed-at`; demand-scoped GC probably NOT wanted here — this is a standing observation, closer to Host status than a bound CR record; decide at implementation).
2. The poller script + its independent declaration (gated on #1396; interim vehicle can be a plain cron/launchd on kiwi feeding the same records, so the record shape can land first).

## Blocked by

- #1396 (ensured convoys) for the independent form; slice 1 and the interim cron form can start immediately once codex budget resets (or via a claude crew once #1411 merges and deploys).

rjwittams:
## Summary

- add the subject-keyed `Usage` observed resource and Kubernetes CRD
- preserve quota windows as a named list with pace, provider cost, reset-credit, identity, and observation metadata
- replicate usage observations fleet-wide and expose numeric/timestamp leaf paths through `usage/<account>`
- keep usage records standing (no demand-scoped garbage collection) with thin observation event history

Slice 2, the CodexBar poller/independent, remains out of scope pending #1396.

## Test plan

Closes #1412

### DIFF
--- a/crates/flotilla-core/src/leaf_engine.rs
+++ b/crates/flotilla-core/src/leaf_engine.rs
@@ -5,7 +5,7 @@ use flotilla_protocol::{DaemonEvent, Leaf, LeafAddress, LeafFire, WaitSubscripti
 use flotilla_resources::{
     admit_leaf, controller::SecondaryWatch, evaluate_leaf, instantiate_exit, select_convoy_children, ChangeRequest,
     ChangeRequestLeafSubject, Checkout, Convoy, ConvoyLeafSubject, ConvoyPhase, InstantiatedExit, ResourceBackend, ResourceError,
-    ResourceObject, ThreeValue, Vessel, VesselLeafSubject, WatchEvent, WatchStart, WorkLeafSubject,
+    ResourceObject, ThreeValue, Usage, UsageLeafSubject, Vessel, VesselLeafSubject, WatchEvent, WatchStart, WorkLeafSubject,
 };
 use futures::StreamExt;
 use tokio::{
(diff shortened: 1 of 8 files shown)


rjwittams:
Verified, and keeping the current behavior intentionally. The normalized hash defines record lookup identity, while the first observation preserves the account spelling as the immutable stored subject; subsequent observations replace status rather than updating the spec. Slice 2 should therefore get-or-create by `usage_record_name(account)` and update status on an existing record, regardless of later casing or surrounding whitespace. If the poller needs canonical display identity, that belongs in its translation policy rather than a case-only spec mutation.

PR #1427, opened 2026-08-08
fix: launch crew adapters by absolute path

### LINKED ISSUE #1415: fix: crew launch uses bare adapter binary name, inheriting a stale cleat-daemon PATH; dead session leaves Working/Running status forever
Found during the first claude-crew e2e (#1414 dispatch, post-#1411).

## What happened

`--agent claude-code:sonnet` dispatched cleanly, placement admitted feta (daemon discovery sees `~/.local/bin/claude`), the TerminalSession recorded a correct launch command — and the crew never existed. The launched command was bare `claude …`, resolved not against the flotilla daemon's PATH but against the **cleat daemon's** ambient PATH — a process running since Jul 27 whose PATH predates today's claude install and lacks `~/.local/bin`. `claude: command not found` → the session's command died at spawn → cleat reaped the dead session (`capture`/`inspect` return "not found" within seconds).

Two defects:

1. **Bare binary names cross an environment boundary they shouldn't.** Discovery asserts a concrete binary path (`find_binary`), and placement admission passes because *the flotilla daemon* can see it — but the launch command is executed inside a cleat-daemon environment with a different PATH, so admission's promise doesn't hold at spawn. The adapter should launch the **absolute discovered path** (or the pool should carry an explicit PATH into the session env). Codex only ever worked by coincidence of PATH overlap.
2. **A session that dies at spawn is invisible.** TerminalSession status stayed `phase=Running, inner_command_status=Running` with a recorded launch command, and crew hook state stayed `Working`, indefinitely — while `cleat` had already reaped the session. Session-gone must be observed and surfaced: fail the work / relaunch / raise attention, and never let `Running` bookkeeping outlive a reapable session by more than a poll interval.

## Interim mitigation (in place)

`~/.cargo/bin/claude` symlinks on feta and udder put claude on the old cleat daemons' PATH; verified `claude --version` now runs inside cleat sessions. This evaporates on the next claude version bump (symlink targets the versioned binary) — the absolute-path fix is the real one.

## Acceptance

- Adapter launch commands use the discovered absolute binary path (behavior test: launch plan for an adapter discovered at `/x/y/claude` contains `/x/y/claude`, not bare `claude`).
- A TerminalSession whose cleat session no longer exists transitions out of Running within one reconcile/poll cycle and surfaces the failure (test: reap the session out from under a Running record ⇒ status flips, work fails or relaunches, attention raised).

rjwittams:
Closes #1415

## What changed

- resolve PATH-discovered Claude Code and Codex executables to absolute paths inside the target execution environment
- retain those resolved paths in discovery assertions so adapter launch plans cross the cleat-daemon boundary without relying on its ambient PATH
- reject non-absolute resolution results instead of recording a bare command that cannot uphold placement admission
- add a launch-plan regression test and update host/provisioned-environment discovery fixtures

## Root cause

The version probes established that an adapter binary existed in the Flotilla runner environment, but recorded the bare command name. Cleat later evaluated that command in its daemon environment, whose PATH can differ or be stale.

The existing terminal and vessel reconciliation tests also verify the second reported failure mode: a reaped running session is marked stopped on the next observation, interrupts active work, and requests a restart.

## Validation

### DIFF
--- a/crates/flotilla-core/src/agent_adapter.rs
+++ b/crates/flotilla-core/src/agent_adapter.rs
@@ -1132,6 +1132,23 @@ mod tests {
         assert!(plan.command.contains("--model 'opus; touch /tmp/pwned'"));
     }
 
+    #[test]
+    fn launch_plan_uses_the_discovered_absolute_binary_path() {
+        let env = EnvironmentBag::new().with(EnvironmentAssertion::binary("claude", "/x/y/claude"));
+        let registry = AgentAdapterRegistry::discover(&env, Arc::new(MockRunner::new(Vec::new())));
+        let brief =
+            flotilla_resources::TerminalBrief { path: ".flotilla/briefs/coder.md".into(), content: String::new(), copies: Vec::new() };
+
+        let plan = registry
+            .get("claude-code")
+            .expect("claude adapter")
+            .launch(&AgentLaunchRequest { role: "coder".into(), model: None, brief })
+            .expect("launch plan");
+
+        assert!(plan.command.starts_with("/x/y/claude "));
+        assert!(!plan.command.starts_with("claude "));
+    }
+
     #[tokio::test]
     async fn claude_prepare_writes_a_settings_overlay_the_launch_command_loads() {
         let temp = tempfile::tempdir().expect("tempdir");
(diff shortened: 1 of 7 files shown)


rjwittams:
Fixed in 5e078da8. The resolver now uses non-login `sh -c`, so `command -v` inherits the same runner PATH as the direct version probe instead of allowing login-shell startup files to change it. The discovery-focused tests, provisioned-environment regression, formatting, and workspace clippy all pass with the revised invocation.

rjwittams:
Confirmed the re-review finds the PATH fix sound. I filed the non-blocking diagnostic improvement as #1430 so resolution failures after a successful version probe can be surfaced without expanding this correctness fix.

Issue #1425, opened 2026-08-08
Sleep inhibition: key standing convoys on quiescence


rjwittams:
## Context

Follow-up to #1396. Standing convoys intentionally have no exit table and remain non-terminal, so the v1 sleep inhibitor keeps their host awake permanently. #1396 accepts that behavior for the first slice.

## What to build

Refine sleep inhibition so a standing convoy inhibits sleep only while non-quiescent, without weakening the existing safety guarantee for ordinary active convoys. Define the authoritative quiescence signal and cover wake/re-inhibit transitions when new work or delivery arrives.

## Acceptance criteria

- A quiescent standing convoy does not permanently inhibit host sleep.
- New standing-convoy activity re-establishes inhibition before actuation depends on the host.
- Ordinary non-terminal convoys retain their current inhibition behavior.
- Behavior tests cover quiescent, wake, and race-boundary transitions.

Linked: #1396, #1394.

PR #1428, opened 2026-08-08
feat: ensure standing convoys from ops entries

### LINKED ISSUE #1396: Ensured convoys: declare standing convoys as ops entries; the ensure loop

rjwittams:
Closes #1396

## Summary

- add a definitions-class `ConvoyEnsure` resource materialized from project ops `ensure` entries, including repository alias selection, workflow, placement/stance preferences, presentation hint, and source provenance
- run a level-triggered standing-convoy ensure loop through normal admission and teardown paths, with persisted exponential restart backoff for failed or reaped convoys
- reap ensured convoys when their declarations disappear on project refresh and describe no-exit convoys as standing rather than stuck
- document the entry format and record the deferred quiescent sleep-inhibition refinement in #1425

## Verification


rjwittams:
Addressed the review suggestions in 9e4fe5a8:

- restart pressure now resets after 10 minutes of healthy running, tracked via `running_since`
- automatic recovery force-reaps only terminal ensured convoys; declaration removal continues through the normal safety gate
- docs now call out both lifecycle choices explicitly
- ensure materialization eagerly rejects workflows that declare an exit workflow
- added CRD parsing coverage and removed the duplicate `presents-as` annotation copy path

The exact formatting, clippy, and full workspace test commands all pass locally.

rjwittams:
Addressed both follow-up observations in 306c1a50:

- protocol now owns the machine-safe `standing` settlement-mode constant, used by core, TUI formatting, and tests; the human output remains `STANDING (no exit table)`
- partial ensure-reconcile failures now include any successful changes in the returned error, so the daemon warning preserves observability of mutations completed in the same pass

Nightly format, clippy with warnings denied, and the exact full workspace test command all pass locally.

rjwittams:
Acknowledged. I’m keeping `bon::Builder` on `ConvoyEnsureSpec`: the resource has several optional fields and the derive follows the repository convention for public resource-construction APIs, even though the two current internal call sites use literals where all fields are intentionally visible. I’m leaving those literals explicit rather than adding a style-only commit that would restart the long-running integration suite.

PR #1431, opened 2026-08-08
feat: poll CodexBar usage observations

### LINKED ISSUE #1412: feat: usage-observer independent — poll CodexBarCLI into usage/<account> observed records

rjwittams:
## Summary

- poll the host-local CodexBarCLI for Codex and Claude every five minutes
- translate every quota lane, including named extra windows, plus identity, pace, provider cost, and reset credits into account-keyed Usage observations
- publish Usage status through the daemon so observations remain leaf-addressable and replicate with the existing observations class
- declare a no-exit standing tool workflow ensured on kiwi via `host-direct-d49f4c59-811f-44aa-a4ca-d4cac66cf2a3`

## Test plan

Closes #1412


rjwittams:
Thanks for the detailed review. I evaluated both minor notes and am leaving them unchanged for this slice:

- The checked-in record/replay test targets the requested CodexBar contract and preservation of the full observation shape. Unsupported, mismatched, unavailable, and empty payloads already fail closed with surfaced provider errors; direct negative-path coverage would be useful follow-up hardening but is not required for this delivery.
- The human output is intentional as a low-frequency health trace for the standing vessel (two providers on a five-minute cadence), while failures remain on stderr.

No blocking feedback remains from this review.

rjwittams:
Implemented the merge ruling in `f208029e`.

Usage publication now skips the status write unless a window moves by at least 1.0 percentage point, the window set/reset time/plan/provider cost/reset credits changes, or the stored observation reaches the one-hour heartbeat. The decision is re-evaluated inside the optimistic-concurrency retry loop.

Regression coverage proves an unchanged five-minute observation preserves the existing `resourceVersion`, and covers each material-change category plus the heartbeat boundary.

Local gates passed:

rjwittams:
Agreed that monotonicity becomes necessary once Usage has concurrent or multi-source writers. I filed #1433 to define and enforce that policy at the Usage resource/federation boundary, so every writer gets the same ordering guarantee rather than adding a CodexBar-only guard here.

This PR's deployed writer remains single-host and sequential, so the observation is non-blocking for the current delivery as reviewed.

rjwittams:
Implemented the architecture rework ruling in `0720ade2`.

- Replaced the bespoke `UsageObserve` command with generic `ResourceStatusPatch { namespace, kind, name, status }`.
- Added registry-selected typed status decoding and bounded conflict retry for every registered resource kind.
- Routed the verb through the daemon resource dispatcher and exposed `flotilla resource patch-status <kind> <name> -f <file>` (including explicit host routing).
- Added in-process coverage for generic Usage creation + status patch, including platform-side rejection of malformed Usage status.
- Removed the compiled CodexBar module, hidden binary subcommand, poll interval, and replay fixture.
- Rewrote `scripts/usage-observer` as the standalone Python 3 polling workload with the exact CodexBar CLI path, Usage name hashing, resource creation, generic status publication, 1-point/structural coalescing, and hourly heartbeat.
- Kept both ops declarations unchanged.

Exact local gates passed:

rjwittams:
Confirmed against the architecture ruling: this workload must use the documented `usage_record_name` scheme exactly (`sha256("usage-account-v1\0" + normalized account)`), so changing this script alone to key on `(provider, account)` would violate the platform resource identity contract and diverge from `usage.rs`. Any provider-disambiguation change needs to be made at the Usage resource/domain boundary, including its immutable spec identity and all writers/readers, rather than as a CodexBar-specific key variant. I am therefore leaving the ruled account-keyed behavior unchanged in this PR.

Likewise, the coalescing predicate intentionally matches the ruling’s exhaustive structural list: window set/`resets_at`, `plan`, `provider_cost`, and `reset_credits`, plus >=1pp movement and the hourly heartbeat. `organization`/`pace` remain bounded by that heartbeat rather than expanding this PR’s write-trigger policy.

rjwittams:
Tracked this domain-level gap in #1443. The issue preserves the ruled account-first identity while requiring the Usage resource/status model to represent multiple provider observations without silent replacement. This remains out of scope for the CodexBar workload and needs resolution at the shared Usage seam.

rjwittams:
The provider-scoped keying amendment arrived after this PR had already merged. The fix is rebased onto the merge commit and continues in #1446; local fmt, clippy, and workspace tests are green.

PR #1434, opened 2026-08-10
fix: classify Claude headless credit exhaustion

### LINKED ISSUE #1421: ClaudeOauth preflight: distinguish exhausted headless credit pool from dead token
## Context

The ClaudeOauth credential preflight (PR #1419) verifies a delivered `CLAUDE_CODE_OAUTH_TOKEN` by running a minimal `claude -p ok` under the token. Since Anthropic's June 15 2026 billing change, `-p`/headless invocations meter against a separate Agent SDK credit pool (per-plan monthly allowance), while interactive `claude` sessions — which is what crews actually run — meter against the normal subscription limits. Sources are third-party writeups; official docs have not caught up. Same auth path either way, so the preflight remains a valid credential check.

## Problem

This creates one false-negative scenario: if the account's Agent SDK credit pool is exhausted, `claude -p ok` fails even though the token is valid and an interactive crew session would work fine. Environment prep would then refuse with a message implying the credential is dead, which sends the operator to re-mint a token that is perfectly healthy.

## What to build

Make the preflight failure handling distinguish the two cases:

- Token invalid/expired → fail prep loudly naming the credential (current behavior, correct).
- Headless credit pool exhausted (identifiable from the CLI's error output mentioning credits/quota/billing) → either proceed with a warning (the crew's interactive path is unaffected) or fail with a message that names the actual cause and does NOT suggest re-minting.

Capture the `claude -p` stderr/stdout on failure and classify before deciding. Add injected-runner tests for both failure shapes.

## Acceptance criteria

- [ ] Exhausted-pool error output does not produce a "credential invalid" style failure
- [ ] Dead-token failure message unchanged (loud, names credential)
- [ ] Tests cover both classifications with realistic CLI error output

## Blocked by

PR #1419 merging.

rjwittams:
Closes #1421

## Summary

- capture combined stdout and stderr from the Claude OAuth print-mode preflight
- classify credit, quota, and billing failures as exhausted headless Agent SDK capacity
- continue preparation with a structured warning because interactive crew sessions use subscription limits
- preserve the existing loud, credential-named, redacted failure for invalid or expired tokens

## Tests

- injected runner covers realistic exhausted Agent SDK credit output and verifies preparation succeeds and caches
- injected runner preserves dead-token failure behavior and rerun semantics
- probe-shape test verifies stdout/stderr capture

## Verification

### DIFF
--- a/crates/flotilla-daemon/src/credential.rs
+++ b/crates/flotilla-daemon/src/credential.rs
@@ -722,7 +722,8 @@ impl CredentialStore {
                             &[
                                 "-c",
                                 "IFS= read -r token; unset ANTHROPIC_API_KEY ANTHROPIC_AUTH_TOKEN; \
-                                 CLAUDE_CODE_OAUTH_TOKEN=\"$token\" CLAUDE_CONFIG_DIR=\"$1\" claude -p ok",
+                                 output=$(CLAUDE_CODE_OAUTH_TOKEN=\"$token\" CLAUDE_CONFIG_DIR=\"$1\" claude -p ok 2>&1) || \
+                                 { status=$?; printf '%s\\n' \"$output\" >&2; exit \"$status\"; }; printf '%s' \"$output\"",
                                 "flotilla-claude-oauth-preflight",
                                 &config_dir,
                             ],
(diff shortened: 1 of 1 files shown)


rjwittams:
Verified against the issue contract. The fixture is intentionally representative rather than presented as a captured string: #1421 defines exhausted-pool output as mentioning credits, quota, or billing and explicitly records that no official/captured wording is available yet. The classifier mirrors that specified observable vocabulary case-insensitively, while the dead-token regression uses the documented real wording and confirms the existing credential-named failure remains unchanged.

I am leaving the classifier as scoped instead of adding speculative depletion/auth phrases that could create false negatives outside the vocabulary the issue gives us. If Anthropic's observed wording later differs or adds generic billing boilerplate to unrelated failures, that concrete output should drive a focused fixture/classifier update.

PR #1435, opened 2026-08-10
fix: log agent adapter path resolution failures

### LINKED ISSUE #1430: Log agent adapter binary path resolution failures
From PR #1427 review by claude-review bot: https://github.com/flotilla-org/flotilla/pull/1427#issuecomment-5226555152

When an agent binary version probe succeeds but absolute-path resolution fails or returns a relative path, discovery correctly omits the adapter rather than recording a bare command. Add a targeted diagnostic for that anomalous transition so future adapter-registration failures identify the rejected resolution at the discovery boundary.

Keep normal missing-binary detection quiet; this should cover only the case where the executable probe succeeded and the subsequent absolute-path lookup could not uphold the launch guarantee.

rjwittams:
Closes #1430.

## What changed

- warn when an agent adapter's version probe succeeds but absolute-path resolution fails
- warn when path resolution returns a relative path, including the rejected path
- keep ordinary missing-binary detection quiet
- add focused tracing tests for all three transitions

## Why

Agent adapters must launch through an absolute executable path. Discovery already rejected failed or relative resolutions, but did so silently, making an otherwise successful probe followed by adapter omission difficult to diagnose.

## Validation

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1177,6 +1177,7 @@ dependencies = [
  "toml",
  "toml_edit 0.22.27",
  "tracing",
+ "tracing-subscriber",
  "urlencoding",
  "uuid",
 ]
--- a/crates/flotilla-core/Cargo.toml
+++ b/crates/flotilla-core/Cargo.toml
@@ -49,6 +49,7 @@ flotilla-protocol = { path = "../flotilla-protocol", features = ["test-support"]
 flotilla-resources = { path = "../flotilla-resources", features = ["test-support"] }
 tempfile = "3"
 insta = "1"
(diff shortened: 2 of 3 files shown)


rjwittams:
Reviewed both non-blocking observations; no change needed for this fix. The log-capture helper has only two call sites across different crate test surfaces, so extracting shared test infrastructure now would expand scope without current leverage. A failed `sh` resolution after a successful adapter version probe is precisely the anomalous transition #1430 asks discovery to diagnose; ordinary missing-binary probes remain quiet and are covered by the regression test.

PR #1436, opened 2026-08-10
fix: stabilize tracked project log capture

### LINKED ISSUE #1393: Stabilize tracked-project reconciliation log assertion
Observed while validating PR #1389 on the lab Forgejo Darwin runner: https://forgejo.lab.flotilla.work/lab/flotilla/actions/runs/17

`tracked_repo_reconciles_generator_owned_project_fields_and_preserves_custom_fields` intermittently fails at `crates/flotilla-daemon/tests/project_cli.rs:469` because its captured logs do not contain the expected warning, even though the reconciled resource assertions immediately before it pass.

Evidence that this is timing-sensitive:

- the exact test passed immediately when rerun in isolation;
- the unchanged commit passed the complete workspace suite on a fresh ephemeral Darwin guest in run 18;
- the failure is unrelated to the files or behavior changed by PR #1389.

Investigate whether the warning is emitted outside the future-scoped tracing subscriber or whether the assertion should synchronize with the emitting task before inspecting the capture buffer.

rjwittams:
## Summary

- install the reconciliation warning subscriber before introducing Project drift
- keep the test on a current-thread Tokio runtime so the thread-local subscriber remains active across awaits
- capture warnings from either the explicit reconciliation or the daemon's background poll

## Root cause

The test introduced generator-owned Project drift before installing a subscriber on only the explicit reconciliation future. The daemon's background poll could reconcile that drift first and emit the warning outside the future-scoped subscriber. The resource assertions would then pass, while the later explicit reconciliation was already at steady state and produced no captured warning.

## Validation

- 1,000 isolated repetitions of the affected test
- 25 concurrent repetitions of the `project_cli` integration test suite

Closes #1393

### DIFF
--- a/crates/flotilla-daemon/tests/project_cli.rs
+++ b/crates/flotilla-daemon/tests/project_cli.rs
@@ -28,8 +28,6 @@ use flotilla_resources::{
     ProjectRepositoryRole, ProjectSpec, Repository, RepositoryKey, RepositorySpec, RepositoryStatus, ResourceBackend, Stance,
     WorkflowTemplate, WorkflowTemplateSpec, MANAGED_BY_LABEL,
 };
-use tracing::instrument::WithSubscriber;
-
 #[derive(Clone)]
 struct LogCaptureWriter(Arc<Mutex<Vec<u8>>>);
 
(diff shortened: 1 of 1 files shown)


rjwittams:
Confirmed. The deterministic reproduction reconciled the drift once before the future-scoped capture: resource assertions stayed green while the warning buffer was empty. Moving the thread-local subscriber ahead of drift creation on the explicit current-thread runtime closes that gap for both reconciliation paths.

PR #1437, opened 2026-08-10
fix: collect orphaned resource replicas

### LINKED ISSUE #1422: Orphaned replica records with no living authority are undeletable and survive reaping forever
## Observed (2026-08-06, post-bump fleet: kiwi/feta/udder all on 5a24d519)

Reaping dead convoys left two kinds of zombie records visible in every daemon's replicated overlay that **no daemon will delete**:

- Convoy `clone-single-flight` (phase Landed) — listed by kiwi (and udder), absent from feta's list entirely.
- 14 TerminalSession records (`terminal-checkout-cascade-work-coder`, `terminal-compose-ci-cost-work-coder`, `terminal-docker-crew-smoke-r4/5/6-work-coder`, `terminal-smoke-1278-interior-cleat-3/4-work-coder`, `terminal-ssh-stub-etxtbsy-work-coder`, `terminal-ssh-transport-seam-work-coder`, `terminal-landing-vacuous-settle-work-coder`, `terminal-seam-review-round-work-coder`, `terminal-sleep-placement-keyed-work-coder`, `terminal-compose-integration-red-work-coder`, `terminal-daemon-log-silence-work-coder`) — all belonging to convoys deleted long ago (pre-#1418 teardown-leak era).

On **all three daemons**, both `flotilla convoy delete <name>` and the gate-bypassing `flotilla resource delete <kind> <name>` return `error: resource not found` — yet `flotilla resource list` keeps showing the records. Deletion apparently only targets locally-authoritative records, and these records' authority is gone (env-scoped daemon, or a store reset on the authority host), so the replicas are pinned everywhere with no owner to cascade from and no tool that can touch them.

Contrast: same-session reaps of feta-authority convoys (`smoke-1278-interior-cleat*`, `agent-e2e-docs2`) worked exactly as designed when issued on the authority daemon — cascade removed checkouts, dirs, sessions. The gap is specifically authority-less replicas.

## Expected

Either (a) replica GC: a replicated record whose authority no longer exists (or whose authority host is durably gone) is collectable — deliberately, not silently; or (b) an operator escape hatch: `resource delete` (or a `--replica` flag) can purge a replica from the local store when no authority claims the record. The #1142 invariant (replicas are read-only projections) argues for (a) with an explicit tombstone/expiry story rather than ad-hoc local purges.

Related: #1402 (replica-blind reaping), #1142 (replication invariants). This may be partially an artifact of past no-compat store surgery, but the fleet has no way to recover from it, which makes it a standing operational bug.

## Acceptance criteria

- [ ] The 15 records above can be removed from the fleet by some sanctioned operation
- [ ] The chosen mechanism is covered by an InProcessDaemon-level test (authority record vanishes → replica is collectable)

rjwittams:
Closes #1422.

## Summary

- add an explicit `resource delete --replica <origin-root>` escape hatch for authority-less replica records
- persist the collection as a replica tombstone so peer relay converges and stale copies cannot immediately resurrect the record
- refuse replica collection while its authority root is connected, preserving the authoritative deletion boundary
- support `--host` on resource deletion for targeted fleet maintenance
- cover authority disappearance and replica collection at the InProcessDaemon level

## Verification

### DIFF
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -8435,9 +8435,20 @@ impl InProcessDaemon {
             return Ok(id);
         }
 
-        if let flotilla_protocol::CommandAction::ResourceDelete { namespace, kind, name } = &command.action {
+        if let flotilla_protocol::CommandAction::ResourceDelete { namespace, kind, name, replica_origin } = &command.action {
             let empty_identity = self.start_context_free_command(id, command.description().to_string());
-            let result = match flotilla_resources::delete_resource_kind(&self.resource_backend, namespace, kind, name).await {
+            let deleted = if let Some(origin_root) = replica_origin {
+                if self.peer_connection_status(origin_root).await == PeerConnectionState::Connected {
+                    Err(ResourceError::invalid(format!(
+                        "replica origin {origin_root} is connected; delete the authoritative resource instead"
+                    )))
+                } else {
+                    flotilla_resources::collect_resource_replica_kind(&self.resource_backend, namespace, kind, name, origin_root).await
+                }
+            } else {
+                flotilla_resources::delete_resource_kind(&self.resource_backend, namespace, kind, name).await
+            };
+            let result = match deleted {
                 Ok(deleted) => flotilla_protocol::CommandValue::ResourceDeleted(Box::new(ResourceJsonResponse {
                     kind: deleted.kind,
                     plural: deleted.plural,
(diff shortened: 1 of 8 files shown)


rjwittams:
Added focused tests for both `ResourceDeleted` human-output paths: replica collection includes the origin root and recreation note, while authoritative deletion retains its existing wording. The exact format, clippy, and workspace test gates all pass after the addition.

rjwittams:
Fixed. `ResourceJsonResponse` now carries an explicit optional `replicaOrigin` populated from the delete command, and the human formatter uses that signal instead of inspecting user-writable annotations. The authoritative-deletion test now deliberately includes a spoofed origin annotation and verifies it still reports a normal delete. Exact format, clippy, and workspace test gates pass.

rjwittams:
Confirmed. No further action is needed from this review round; the previously raised presentation concerns are covered by the explicit `replicaOrigin` signal and regression tests.

PR #1438, opened 2026-08-10
fix: keep federated convoys on admitting store

### LINKED ISSUE #1188: Contained placement on a remote host resolves its host-direct environment in the wrong store

rjwittams:
## Summary

- keep every newly admitted Convoy and its immutable workflow and placement snapshots in the admitting daemon's resource store, regardless of the selected placement host
- remove the retracted prepared-start peer protocol and host-direct command-routing path, leaving replicated placement intent to the existing placed-on-me actuator projector
- retain actuator-local resources while owner deletion intent has not replicated, then allow teardown once that intent is observed

## Root cause

Remote host-direct starts still used the earlier ownership-routing design: admission prepared an opaque resource bundle, forwarded it to the placement daemon, and made that daemon the Convoy authority. This contradicted the ruled federation model and meant host-direct and contained remote placements had different ownership semantics.

The admitting path also pinned only the workflow locally; placement snapshots were created only by the obsolete remote prepared-start path.

## Impact

Convoy spec, metadata, deletion intent, pinned inputs, and phase roll-up now have one authority: the admitting store. Placement hosts consume replicated intent and append actuator facts to their own stores. Mutable placement policies are snapshotted at admission and replicated with the Convoy's resources.

## Validation

- focused two-store host-direct and Docker worktree-on-host-and-mount provisioning tests
- focused owner-offline deletion-freeze and command-routing regressions

Closes #1188

### DIFF
--- a/crates/flotilla-controllers/tests/presentation_reconciler.rs
+++ b/crates/flotilla-controllers/tests/presentation_reconciler.rs
@@ -691,6 +691,7 @@ async fn create_ready_host(backend: &ResourceBackend, name: &str) {
         daemon_version: None,
         daemon_started_at: None,
         disk_free_bytes: None,
+        admission_free_space_floor_bytes: None,
     }
     .apply(&mut status);
     hosts.update_status(name, &created.metadata.resource_version, &status).await.expect("host status update should succeed");
--- a/crates/flotilla-controllers/tests/vessel_placement.rs
+++ b/crates/flotilla-controllers/tests/vessel_placement.rs
@@ -4,8 +4,8 @@ use chrono::Utc;
 use flotilla_controllers::reconcilers::VesselPlacementProjector;
 use flotilla_protocol::{NodeId, PlacementDecision, PlacementTargetHost};
 use flotilla_resources::{
-    Convoy, ConvoySpec, ConvoyStatus, InMemoryBackend, InputMeta, ResourceBackend, Vessel, VesselSpec, ACTUATOR_HOST_REF_ANNOTATION,
-    ACTUATOR_SOURCE_ROOT_ANNOTATION, CONVOY_LABEL,
+    Convoy, ConvoySpec, ConvoyStatus, InMemoryBackend, InputMeta, ResourceBackend, ResourceError, Vessel, VesselSpec,
+    ACTUATOR_HOST_REF_ANNOTATION, ACTUATOR_SOURCE_ROOT_ANNOTATION, CONVOY_LABEL,
 };
 
 const NAMESPACE: &str = "flotilla";
@@ -57,7 +57,7 @@ async fn placed_replica_is_projected_into_the_actuation_hosts_local_store() {
         .replace(&convoys.list().await.expect("list Convoys"), Utc::now())
         .await
         .expect("replicate Convoy");
-    feta.replica_writer::<Vessel>(kiwi_root, NAMESPACE)
+    feta.replica_writer::<Vessel>(kiwi_root.clone(), NAMESPACE)
         .replace(&kiwi.using::<Vessel>(NAMESPACE).list().await.expect("list Vessels"), Utc::now())
         .await
         .expect("replicate Vessel");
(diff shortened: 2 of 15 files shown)


rjwittams:
Fixed in ec91779d. Each daemon now advertises both measured free bytes and its configured admission floor in replicated Host status. The admitting store enforces those federated facts before persisting a remotely placed convoy, so target-host capacity protection is preserved without restoring the routed PreparedConvoyStart ownership transfer. The integration coverage now proves both successful local ownership for a healthy remote target and refusal with no Convoy on either store when the target is below its advertised floor.

I also filed #1439 for the separate obsolete pending prepared-snapshot recovery cleanup.

rjwittams:
Follow-up CI fix in 98e0f604: explicit remote host-direct placement now performs the established connectivity preflight before local admission hydration. This preserves the stopped-host error contract exercised by wedge-regressions without routing the ConvoyStart or changing admitting-store ownership. Exact local format, clippy, and full sandbox-safe workspace tests pass.

rjwittams:
Fixed in e14c2c88. Capacity enforcement now recognizes every host-targeted placement via the shared placement_host_ref helper, including docker_per_vessel, and fails closed when the target has not advertised capacity. Added remote_docker_admission_fails_closed_without_target_capacity to prove the missing-replica/capacity case refuses before Convoy persistence. Local Docker fixtures now advertise their modeled host capacity explicitly.

rjwittams:
Resolved at the new head e14c2c88: the gate now calls placement_host_ref(&policy), so both host_direct and docker_per_vessel are fail-closed, with a dedicated missing-capacity remote Docker regression test. This comment reviewed the preceding 98e0f604 head before that fix landed.

rjwittams:
Fixed in eb1d13e7. Legacy/template `ConvoyCreate` now uses the same two remote host-direct safeguards as `ConvoyStart`: the request-session router verifies the selected host is connected without forwarding authority, and the admitting store checks the target host's published free-space floor before persisting the convoy. The regression test exercises legacy remote Docker placement with missing capacity and verifies refusal before convoy creation; the existing provenance test now publishes target capacity explicitly.

rjwittams:
Fixed in 8a566561. `start_ensured_convoy` now runs both the local free-space check and the federated target-host capacity check after resolving admission and before creating snapshots or Convoy state. The standing-convoy reconciler test now adds a remote host-direct ensure whose host has no published capacity, verifies the reconcile pass reports the fail-closed refusal/backoff, and confirms no Convoy is created. Exact local format, Clippy, and full sandbox-safe workspace tests pass.

PR #1447, opened 2026-08-11
fix: tombstone missing authoritative resources

### LINKED ISSUE #1445: fix: replica collection deadlocks when a connected origin has lost the record
## Parent

#1422 (closed by #1437) — the shipped escape hatch cannot clear the exact 15-record inventory that motivated it.

## Observed (2026-08-11, post-bump fleet on current main)

All 15 zombie records from #1422 (convoy `clone-single-flight` + 14 `TerminalSession` records) carry origin root `9ad3ffb1931f9996979535186f45027f`, which is **feta's live store root**. Feta's store lost the records in past no-compat store surgery, so:

- On feta (authority): `flotilla resource delete convoys clone-single-flight` → `resource not found`.
- Anywhere: `flotilla resource delete convoys clone-single-flight --replica 9ad3ffb1931f9996979535186f45027f` → `replica origin ... is connected; delete the authoritative resource instead`.

The guard (`in_process.rs`, ResourceDelete handling: `peer_connection_status(origin_root) == Connected` → refuse) assumes a connected origin can cascade the deletion. When the origin lost the record — a case #1422 called out explicitly ("a store reset on the authority host") — neither path works and the replicas are pinned fleet-wide, exactly as before #1437.

## Expected

Preserve the authoritative-deletion boundary while covering the lost-record case. Preferred shape: **origin-side tombstone-by-name** — the authority host's `resource delete` of a name it does not hold succeeds by writing a replica tombstone that relays through the normal peer path (deletion authority stays with the origin; no replica-side judgment about the origin's contents). Alternative: the replica-collection guard verifies record presence at the connected origin instead of bare connectivity, refusing only when the origin actually holds the record.

## Acceptance criteria

- [ ] The 15 records above are removable from the fleet via the sanctioned operation on current main
- [ ] Connected-origin-holds-record still refuses replica collection (boundary preserved)
- [ ] InProcessDaemon-level test: origin connected but record absent → collectable/tombstoned, converges on peers

rjwittams:
## Summary

- add name-only deletion tombstones to authoritative resource watch streams
- persist and relay those tombstones through in-memory, SQLite, HTTP, and routed replica paths
- remove stale replicas when the authority no longer has the typed resource object
- preserve the connected-origin guard for direct replica collection

## Root cause

Replica collection refused whenever the origin was connected, assuming the origin could delete and cascade the authoritative record. After an authority store reset, the origin no longer had the object needed by the typed delete event, so its delete returned not found while replicas remained protected by the connectivity guard.

The authority can now advance its resource stream with a name-only tombstone when a replicated kind is already absent. Peers retain that tombstone in the existing replica tombstone store, preventing stale relay resurrection without transferring deletion judgment to replicas.

## Validation

- focused InProcessDaemon connected-origin convergence regression
- in-memory and SQLite resource backend tombstone contracts

Closes #1445

### DIFF
--- a/crates/flotilla-controllers/src/reconcilers/vessel_placement.rs
+++ b/crates/flotilla-controllers/src/reconcilers/vessel_placement.rs
@@ -181,8 +181,9 @@ impl VesselPlacementProjector {
 
 fn replica_changed<T: Resource>(event: Option<Result<ReadWatchEvent<T>, ResourceError>>, kind: &str) -> Result<bool, ResourceError> {
     let event = event.ok_or_else(|| ResourceError::invalid(format!("{kind} replica watch ended")))?;
-    let source = match event? {
-        ReadWatchEvent::Added(source) | ReadWatchEvent::Modified(source) | ReadWatchEvent::Deleted(source) => source,
+    let provenance = match event? {
+        ReadWatchEvent::Added(source) | ReadWatchEvent::Modified(source) | ReadWatchEvent::Deleted(source) => source.provenance,
+        ReadWatchEvent::DeletedByName { provenance, .. } => provenance,
     };
-    Ok(matches!(source.provenance, ResourceProvenance::Replica { .. }))
+    Ok(matches!(provenance, ResourceProvenance::Replica { .. }))
 }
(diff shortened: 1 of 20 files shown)


rjwittams:
Addressed both review notes in ffa91f83:

- Added a focused HTTP watch test covering metadata-only DELETED decoding and reconstruction of replica provenance annotations.
- Made tombstone discrimination explicit in the in-memory and SQLite authority/replica watch paths: only objects without spec are decoded as name tombstones, while malformed objects with a present spec now preserve their decode error.

Revalidated with the pinned formatter, workspace clippy with warnings denied, and the full locked workspace test suite.

PR #1449, opened 2026-08-11
fix: converge lost record tombstones

### LINKED ISSUE #1448: fix: tombstone delete reports success but lost-record replicas persist fleet-wide
## Parent

#1445 (closed by #1447). The tombstone path unblocked the refusal deadlock, but for records the authority store actually lost, deletion does not converge.

## Observed (2026-08-11, fleet on current main incl. #1447)

On feta (authority for origin root `9ad3ffb1…`), `flotilla resource delete terminalsessions terminal-checkout-cascade-work-coder` prints:

```
deleted flotilla.work/v1/TerminalSession/flotilla/terminal-checkout-cascade-work-coder
Controllers may recreate code-owned objects.
```

— yet the record remains listed on **all three hosts afterward, including feta's own overlay**, and a second identical delete "succeeds" again with the same non-effect. 10 of the 15 #1422 zombies behave this way (`terminal-checkout-cascade`, `compose-ci-cost`, `compose-integration-red`, `daemon-log-silence`, `docker-crew-smoke-r6`, `landing-vacuous-settle`, `seam-review-round`, `sleep-placement-keyed`, `ssh-stub-etxtbsy`, `ssh-transport-seam`, all `-work-coder`).

Contrast: `clone-single-flight` (Convoy) and 4 sessions (`docker-crew-smoke-r4/r5`, `smoke-1278-interior-cleat-3/4`) **did** converge to deleted fleet-wide via the same operation — plausibly the subset whose authoritative records still existed in feta's store, i.e. the ordinary delete path. The failing 10 are the genuinely-lost class #1445 targeted.

## Suspected shape

The tombstone written for a missing authoritative record either (a) is not relayed/applied on peers, or (b) is applied but the peers' stored replicas re-relay and resurrect the record in the overlay (the resurrection loop #1437's replica tombstone was designed to prevent, but on the replica-collection path rather than this authority-side path). Either way, "deleted" output with no fleet-wide effect is a silent failure — the operation should be verifiable.

## Acceptance criteria

- [ ] The 10 records above are removable: authority-side delete of a lost record converges on all peers and the record stays gone
- [ ] Repeat delete of an already-tombstoned name is idempotent and says so (not a fresh "deleted")
- [ ] InProcessDaemon-level test: authority lost record + two peers holding replicas → delete on authority → both peers converge to absent, no resurrection after relay cycles

rjwittams:
## Summary

- retain authority-side name tombstones so repeated deletes are idempotent and report `already deleted`
- advance recovered tombstones beyond the authority's self-origin replica cursor, preventing a reset store from reusing a stale replica resource version
- apply authority tombstones to the local replica partition and reject ahead-of-stream watch cursors so stale overlays relist or consume the newer delete
- cover durable SQLite restart behavior plus a real three-host authority → peer → peer relay topology with no resurrection

## Validation

Closes #1448

### DIFF
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -8347,26 +8347,42 @@ impl InProcessDaemon {
 
         if let flotilla_protocol::CommandAction::ResourceDelete { namespace, kind, name, replica_origin } = &command.action {
             let empty_identity = self.start_context_free_command(id, command.description().to_string());
-            let deleted = if let Some(origin_root) = replica_origin {
-                if self.peer_connection_status(origin_root).await == PeerConnectionState::Connected {
+            let result = if let Some(origin_root) = replica_origin {
+                let deleted = if self.peer_connection_status(origin_root).await == PeerConnectionState::Connected {
                     Err(ResourceError::invalid(format!(
                         "replica origin {origin_root} is connected; delete the authoritative resource instead"
                     )))
                 } else {
                     flotilla_resources::collect_resource_replica_kind(&self.resource_backend, namespace, kind, name, origin_root).await
+                };
+                match deleted {
+                    Ok(deleted) => flotilla_protocol::CommandValue::ResourceDeleted(Box::new(ResourceJsonResponse {
+                        kind: deleted.kind,
+                        plural: deleted.plural,
+                        namespace: deleted.namespace,
+                        value: deleted.value,
+                        replica_origin: replica_origin.clone(),
+                    })),
+                    Err(error) => flotilla_protocol::CommandValue::Error { message: error.to_string() },
                 }
             } else {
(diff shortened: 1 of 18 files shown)


rjwittams:
Thanks for the close read. I verified both observations against the mutation path.

For the HTTP diagnostic, `delete_resource_kind` is only invoked against the daemon's embedded authoritative store (`InMemory` or `Sqlite`); `HttpBackend` remains a pull-only replication transport, so the earlier `local_root` error is not user-reachable through this command path.

For the live-authority plus stale-self-replica coexistence case, this PR deliberately keeps the new replica-tombstone write scoped to the missing-authority recovery path from #1448. Existing authoritative objects retain their ordinary object-bearing delete/finalizer semantics; broadening that path would need to account for pending finalization rather than treating every delete request as absence. There is no concrete reproduction of that separate state here, so I am not changing or filing speculative work from this PR.

PR #1451, opened 2026-08-12
fix: force raw deletion past abandoned finalizers

### LINKED ISSUE #1450: fix: version-less legacy rows are undeletable — delete reports success, row persists and keeps relaying
## Parent

#1422 → #1445 (#1447) → #1448 (#1449). Third field failure of the same cleanup; this time with the decisive evidence.

## Observed (2026-08-12, fleet on current main incl. #1449)

The 10 remaining #1422 zombie TerminalSession records are **not** authority-less replicas. From feta's own overlay list, `terminal-checkout-cascade-work-coder` shows:

```
provenance: {"nodeId": "9ad3ffb1931f9996979535186f45027f", "source": "local"}   ← feta IS the authority
metadata.resource_version: None                                                  ← legacy/broken row shape
```

Against that row, on feta:
- `flotilla resource delete terminalsessions terminal-checkout-cascade-work-coder` prints `deleted flotilla.work/v1/TerminalSession/...` — **every time**, never #1449's `already deleted`, and the record persists on all three hosts including feta's own list.
- Pre-#1447 the same delete said `resource not found`.

So the delete path (lookup by name+version?) never matches the version-less row; each round's fix validated against synthesized store state (the #1449 three-host relay test passes) while the real rows — presumably artifacts of past no-compat store surgery — have `resource_version: None` and slip through delete, yet are still served by list and relayed to peers.

## What to build

- Reproduce against the real row shape first: extract the actual rows from feta's `~/.local/state/flotilla/resources.sqlite` (read-only inspection on the placement host) into a regression fixture/test — do not assume the synthesized shape.
- Make deletion actually remove (or tombstone) version-less rows; success output must reflect reality, and repeat deletes must report `already deleted`.
- Decide whether version-less rows should be servable at all: if they're invalid, listing/relaying them silently is its own bug — consider quarantining or normalizing on startup instead.

## Acceptance criteria

- [ ] The 10 named records are removed from all three hosts and stay gone through relay cycles
- [ ] A test covers delete/list/relay against a version-less row extracted from the real store shape
- [ ] Delete output is truthful: removal says deleted once, repeats say already deleted, non-removal is an error

rjwittams:
## What changed

- make the raw `resource delete` recovery path clear abandoned runtime finalizers and verify the object is physically absent before reporting success
- retain and relay an authoritative name tombstone after deleting replicated resources, making repeat deletes report `already deleted` and preventing stale resurrection
- add a SQLite regression fixture extracted read-only from feta and exercise local delete plus two-peer relay

## Root cause

Read-only inspection of feta corrected the initial version-less hypothesis: the physical row had resource version `9234` in both the scalar column and JSON metadata. The ten zombie terminal rows were already marked for deletion but retained `flotilla.work/terminal-teardown`. Ordinary store deletion correctly no-ops while finalization is pending, while the raw administrative command incorrectly rendered that no-op as `deleted` on every call.

The raw command is documented as the sharp recovery path that bypasses lifecycle and teardown gates. It now forces only that administrative path past abandoned finalizers; controller-facing deletion keeps normal finalizer semantics.

## Impact

A raw deletion either removes the object and reports `deleted` once, reports `already deleted` from its retained tombstone on repeats, or returns an error if the object remains. Replicas converge on the tombstone and reject older relayed writes.

## Validation

Closes #1450

### DIFF
--- a/crates/flotilla-resources/src/registry.rs
+++ b/crates/flotilla-resources/src/registry.rs
@@ -580,16 +580,30 @@ async fn delete_typed<T: Resource>(backend: &ResourceBackend, namespace: &str, n
     let (value, already_deleted) = match resolver.get(name).await {
         Ok(object) => {
             resolver.delete(name).await?;
+            if T::REPLICATION_CLASS == crate::ReplicationClass::Definitions {
+                return Ok(DynamicResourceDelete {
+                    object: DynamicResourceObject {
+                        kind: T::API_PATHS.kind.to_string(),
+                        plural: T::API_PATHS.plural.to_string(),
+                        namespace: namespace.to_string(),
+                        value: object_value(&object)?,
+                    },
+                    already_deleted: false,
+                });
+            }
+            force_pending_finalization(&resolver, name).await?;
+            match resolver.get(name).await {
+                Err(ResourceError::NotFound { .. }) => {}
+                Ok(_) => return Err(ResourceError::other(format!("delete reported success but {name} remains in the resource store"))),
+                Err(error) => return Err(error),
+            }
+            if T::REPLICATION_CLASS != crate::ReplicationClass::None {
+                retain_authoritative_name_tombstone::<T>(backend, namespace, name).await?;
+            }
             (object_value(&object)?, false)
         }
         Err(ResourceError::NotFound { .. }) if T::REPLICATION_CLASS != crate::ReplicationClass::None => {
-            let write = resolver.tombstone(name).await?;
-            if write.created {
-                backend
-                    .replica_writer::<T>(backend.local_root()?, namespace)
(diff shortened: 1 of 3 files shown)


rjwittams:
Addressed the first observation in e75c4ff1: a successful finalizer-clearing update now returns immediately.

I kept the explicit post-delete presence check from the second observation. The issue contract requires non-removal to be reported as an error, so if the same name is present at command completion the success output would be untruthful. For replicated kinds, the subsequent authoritative name-tombstone transaction also rejects a concurrent recreation and prevents stale relay resurrection; for non-replicated code-owned kinds, the existing CLI warning still explains that a later controller reconciliation may recreate the object after a successful command.

rjwittams:
Re-review confirmed the final state and raised no new action items. No further code changes needed.

PR #1453, opened 2026-08-13
fix: unify free-space measurement path

### LINKED ISSUE #1441: task: reconcile free-space measurement paths (state_dir vs repo_default_dir)
From PR #1438 review (final round, non-blocking): the two free-space checks measure different filesystem paths.

- `check_local_free_space_floor` probes `config.state_dir()` (crates/flotilla-core/src/in_process.rs, `check_free_space_floor` call).
- `HostStatus.disk_free_bytes` — which #1438 makes load-bearing for *hard federated admission refusal* via `check_remote_placement_free_space_floor` — probes `profile.repo_default_dir` (crates/flotilla-daemon/src/runtime.rs, `measure_available_space`).

These can be configured onto different mounts, in which case a host can pass one gate and fail the other for the same admission, or advertise capacity that doesn't reflect where checkouts actually land.

## What to build

Decide the semantic: convoy admission consumes disk primarily under the checkout root (`repo_default_dir`) but also writes daemon state (`state_dir`). Either measure both paths and gate on the minimum, or pick one canonical measurement and use it consistently for both the local gate and the published `disk_free_bytes`. Add a note or test pinning the choice.

## Acceptance criteria

- [ ] Local and federated free-space gates use a consistent, documented measurement basis
- [ ] Behavior when the two dirs sit on different mounts is deliberate, not incidental
- [ ] Regression test or doc comment pins the semantic

## Blocked by

#1438 merging.

rjwittams:
## Summary

- make the host-direct checkout root the canonical free-space measurement path for convoy admission
- use the same injected capacity probe for the local admission gate and published `HostStatus.disk_free_bytes`
- document the field semantic and cover separate state/checkouts mounts with a regression test

## Why

The local gate measured `state_dir` while federated admission consumed capacity measured at `repo_default_dir`. Hosts with those directories on different mounts could make contradictory admission decisions. Checkout storage is the primary disk consumer, so both paths now use `repo_default_dir`.

## Validation

Closes #1441

### DIFF
--- a/crates/flotilla-core/src/admission.rs
+++ b/crates/flotilla-core/src/admission.rs
@@ -43,10 +43,6 @@ pub(crate) fn check_free_space_floor(probe: &dyn AvailableSpaceProbe, host: &str
     check_measured_free_space(host, free_bytes, floor_bytes)
 }
 
-pub fn measure_available_space(path: &Path) -> Option<u64> {
-    SystemAvailableSpaceProbe.measure(path)
-}
-
 pub(crate) fn check_measured_free_space(host: &str, free_bytes: u64, floor_bytes: u64) -> Result<(), String> {
     if free_bytes >= floor_bytes {
         return Ok(());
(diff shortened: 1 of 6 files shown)


rjwittams:
Fixed the attribute ordering so the `disk_free_bytes` doc comment now precedes `#[serde(...)]`, matching the surrounding documented fields.

rjwittams:
Fixed. Removed the now-unused public `measure_available_space` helper and its crate-root re-export; the shared injected admission probe is now the only capacity measurement path.

rjwittams:
Confirmed. The final review found no remaining issues, so no further code changes are needed.

PR #1454, opened 2026-08-13
fix: select daemon config and state together

### LINKED ISSUE #1408: Make CLI daemon identity select config and state together
From PR #1405 review by claude-review bot: https://github.com/flotilla-org/flotilla/pull/1405#issuecomment-5193654259

The top-level `flotilla` CLI exposes `--config-dir` but not `--state-dir` or a first-class root selector. Two explicitly configured daemon sockets can therefore use different config directories while retaining the same ambient XDG state directory, sharing `resources.sqlite`, lifecycle locks, peer sockets, and runtime caches.

Config and state cannot generally be validated as sibling directories: the default supported `PathPolicy` deliberately maps them to separate XDG categories (`~/.config/flotilla` and `~/.local/state/flotilla`). A sound fix should make daemon identity explicit instead of inferring a state path from arbitrary config syntax. Options to evaluate:

- expose a root option that deterministically selects all path categories;
- expose `--state-dir` alongside `--config-dir` and require complete identity overrides for daemon-starting commands;
- replace category-specific top-level overrides with a typed daemon-path selection.

Add behavior coverage proving two custom daemon identities cannot silently share a resource store or lifecycle lock.

Context: PR #1405 fixes the incident path where `FLOTILLA_ROOT` and the requested socket disagreed. That environment-based root already resolves config and state together. This follow-up covers the older, distinct config-only CLI override semantics without imposing an invalid sibling-directory invariant on XDG installations.

rjwittams:
## Summary

- add a first-class `--root` selector that resolves daemon config, state, and socket paths together
- expose `--state-dir` and require it alongside `--config-dir` on daemon-starting paths
- preserve config-only socket selection for commands that never spawn a daemon
- cover independent roots using distinct resource databases and lifecycle locks

## Root cause

The top-level CLI applied `--config-dir` to config and socket resolution while retaining the ambient XDG state directory. Distinct custom config identities could therefore share `resources.sqlite` and daemon lifecycle state.

## Validation

Closes #1408

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -34,9 +34,17 @@ struct Cli {
     #[arg(long)]
     repo_root: Vec<PathBuf>,
 
-    /// Config directory
-    #[arg(long)]
+    /// Daemon identity root (config and state use ROOT/config and ROOT/state)
+    #[arg(long, conflicts_with_all = ["config_dir", "state_dir"])]
+    root: Option<PathBuf>,
+
+    /// Config directory (daemon-starting commands also require --state-dir)
+    #[arg(long, conflicts_with = "root")]
     config_dir: Option<PathBuf>,
+
+    /// State directory (daemon-starting commands also require --config-dir)
+    #[arg(long, conflicts_with = "root")]
+    state_dir: Option<PathBuf>,
 
     /// Socket path (default: ${config_dir}/run/flotilla.sock)
     #[arg(long)]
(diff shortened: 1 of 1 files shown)


rjwittams:
Addressed both clarity observations. `socket_path()` now documents why socket-only commands intentionally bypass paired state resolution, and the help/error wording now says the command may start a daemon. The pairing test also asserts the revised user-facing error. The exact formatter, Clippy, and workspace test gates pass locally.

rjwittams:
Verified this against the downstream connection path; no follow-up is needed. Every spawn-capable path reaches `connect_or_spawn_with_optional_surface_using`, which calls `ensure_daemon_socket_belongs_to_config(socket_path, config_dir)` before inspecting lifecycle state or spawning. Therefore `--root /other/root` plus a conflicting ambient socket fails closed rather than mixing the selected state with another daemon. Contained-host and socket-only paths intentionally connect to the ambient socket without opening or spawning against the selected state directory, so they cannot share a resource store or lifecycle lock through this combination.

PR #1455, opened 2026-08-13
Add replica-aware resource point lookup

### LINKED ISSUE #1390: Add replica-aware point lookup for resource reads
From PR #1388 review by claude-review bot: https://github.com/flotilla-org/flotilla/pull/1388#issuecomment-5179836107

Add a replica-aware point-lookup primitive, such as `ReplicaReadResolver::get`, so `QueryResourceGet` can look up the local object directly and fall back to replicas without listing and serializing the full local and federated collections.

Context: #1388 makes `flotilla get <kind> <name>` correctly see replicated resources on non-authority hosts. The available replica resolver currently exposes collection reads only, so the correct implementation performs two collection scans. This is acceptable for the present one-shot CLI path but should be deepened before the operation becomes a polling or refresh hot path.

rjwittams:
## What changed

- add `ReplicaReadResolver::get` with local-first reads and deterministic replica fallback
- implement point lookup for in-memory, SQLite, and HTTP resource backends while preserving replica provenance
- route `QueryResourceGet` through the point lookup instead of listing and serializing the complete replica overlay
- add contract, HTTP wire, daemon socket, and query regression coverage

## Why

`flotilla get <kind> <name>` must see remote replicas, but collection-only replica reads made each lookup scan both the local and federated collections. This adds the missing point-read primitive while retaining the local collection cursor needed for replay-safe query responses.

## Validation

Closes #1390

### DIFF
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -39,23 +39,23 @@ use flotilla_resources::{
     api_version, apply_resource_document, apply_status_patch as apply_resource_status_patch,
     apply_status_patch_checked as apply_resource_status_patch_checked, bound_change_request_record_name,
     controller::delete_lifecycle_owned_matching, ensure_repository, evaluate_landing_settlement, expected_change_request_leaves,
-    expected_checkout_refs, external_patches as convoy_external_patches, list_resource_kind, list_resource_kind_including_replicas,
-    normalize_project_spec, repository_display_labels, resolve_project_issue_sources, terminal_session_attach_target, watch_resource_kind,
-    watch_resource_kind_from, watch_resource_kind_including_replicas, watch_resource_kind_replica_sources, BoundChangeRequest,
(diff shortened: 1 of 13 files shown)


rjwittams:
Fixed in 2ad1561c. `DefinitionResolver::get` now calls the per-name `sources_for_name` path, merges only those sources, and applies the same deleted/conflicted-deletion visibility rule as collection reads. The shared in-memory/SQLite definition contract now exercises replica-aware point reads and missing-name behavior. Format, Clippy, and the full workspace suite pass locally.

rjwittams:
Addressed both notes in 34d694c6. SQLite now creates `replica_objects_by_resource_name` over `(group_name, version, kind, namespace, name, origin_root)`, and a regression uses `EXPLAIN QUERY PLAN` to verify the replica point query selects that index. The definition deleted/conflict visibility rule now lives in one shared helper used by both `get` and `list`. Format, Clippy, and the full workspace suite pass locally.

rjwittams:
Confirmed. No further changes needed; this review finds the implementation complete and the prior findings resolved.

PR #1460, opened 2026-08-13
Decouple providers from follower polling

### LINKED ISSUE #1432: Decouple external service providers from follower mode: on-demand reads work on any credentialed host
## Ruling (2026-08-13 grill, supersedes the original leader-forwarding proposal)

"Leader" conflates three orthogonal things, and this issue's original fix (forward issue resolution to the leader) would have baked the conflation in deeper:

1. **Continuous polling** (Plane-A observer) — the only genuine singleton concern (#885), and it is deleted entirely by the Plane-A cull (see the observer-pipeline deletion issue).
2. **On-demand external reads** — e.g. resolving `--issue N` at dispatch time. One API call per invocation, not polling; safe from any host holding credentials (crews on feta/udder already use `gh` freely).
3. **Admission** — already leaderless post-#1438: convoys live on the admitting host's store.

## What to build

Stop gating provider *construction* on `daemon_config.follower` (`DiscoveryRuntime::for_process(follower)`). External service providers are constructed on every host with the necessary credentials/binaries; only **continuous polling/refresh loops** remain gated by the flag (until the observer cull removes them, at which point the flag evaporates).

Concretely:
- `flotilla convoy start --project <p> --issue <N>` succeeds on a follower-mode host, resolving the issue on-demand through its own provider.
- `missing_external_provider_error`'s "dispatch from a leader host" guidance disappears for hosts that have credentials; a host genuinely lacking credentials/binaries reports that, by name.
- Host provider status no longer reports whole categories as `disabled("follower mode")` when only polling is disabled — the status should distinguish "polling disabled" from "unavailable".
- No polling loops start on follower hosts (preserve the #885 single-poller property).

## Acceptance criteria

- [ ] Issue-dispatch from feta/udder works end-to-end (InProcessDaemon-level test with an injected issue provider + follower flag)
- [ ] No observer polling starts on follower hosts (test)
- [ ] Error for a genuinely credential-less host names the missing capability, not "leader"

## Notes

Governor-independent placement (#1440) gates on this: the governor's core verb is dispatch, and it will live on an always-on host, not the current leader.

rjwittams:
Closes #1432

## Summary
- construct credentialed external providers on follower-mode hosts
- gate only continuous change-request and cloud-agent observer polling
- report constructed providers as polling disabled instead of unavailable in follower mode
- remove leader guidance from genuinely missing issue-provider errors

## Tests
- follower convoy dispatch resolves an injected issue provider end to end
- follower refresh never calls an injected external observer provider
- credential-less follower error names the missing issue-provider capability

### DIFF
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -1944,6 +1944,17 @@ const FLEET_REPLICA_FRESH_SECS: i64 = 90;
 const FLEET_REPLICA_REFRESH_TIMEOUT: Duration = Duration::from_secs(2);
 const ENSURE_BACKOFF_RESET_AFTER: ChronoDuration = ChronoDuration::minutes(10);
 
+fn mark_external_polling_disabled(providers: &mut [HostProviderStatus], follower: bool) {
+    if !follower {
+        return;
+    }
+    for provider in providers {
+        if matches!(provider.category.as_str(), "change_request" | "cloud_agent") {
+            provider.disabled_reason = Some("polling disabled".to_string());
+        }
+    }
+}
+
 fn ensure_retry_delay(restart_count: u32) -> ChronoDuration {
     let exponent = restart_count.min(5);
     ChronoDuration::seconds((30_i64.saturating_mul(1_i64 << exponent)).min(15 * 60))
@@ -2070,7 +2081,7 @@ impl InProcessDaemon {
                 }
             }
             let slug = repo_slug.clone();
-            let mut model = RepoModel::new(
+            let mut model = RepoModel::new_with_external_polling(
                 path.clone(),
                 registry,
                 repo_slug,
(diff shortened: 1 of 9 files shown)


rjwittams:
Applied the useful non-blocking cleanup from the automated review: removed the now-empty missing-provider wrapper, centralized the discovery-to-refresh polling policy conversion, and updated the provider status documentation to cover partial disablement. The suggested second counting mock was not added because change-request and cloud-agent polling share the same `ExternalPolling::Disabled` match in one function; the existing end-to-end counting test exercises that common gate.

rjwittams:
Addressed the useful cleanup points in a98d2a7e: removed the pass-through missing-provider wrapper, centralized polling-policy conversion, and broadened the provider-status field documentation. I kept the single counting provider test because both external categories cross the same `ExternalPolling::Disabled` gate in `refresh_slow_providers`; duplicating the adapter would not exercise a distinct policy path.

PR #1467, opened 2026-08-13
fix: prefer local authority over self-origin replicas

### LINKED ISSUE #1465: fix: self-targeted admission reads a stale self-origin Host replica instead of the live local record
## Observed (2026-08-13, fleet on current main incl. #1451, during the #1410 governor pre-flight smoke)

Dispatching on udder with a docker policy targeting udder itself:

```
error: placement refused on host `udder`: admission free-space floor is unavailable
```

— while udder's own overlay shows TWO Host records named `udder`:

- `source: local` — `disk_free_bytes: 36099731456`, `admission_free_space_floor_bytes: 21474836480` (healthy, above floor)
- `source: replica`, origin `1c8df992acde8c3e863343c568a95be7` — **udder's own live root**, with `disk_free_bytes: None`, `floor: None` (a stale pre-#1438 self-origin replica-partition copy)

So `check_remote_placement_free_space_floor` resolves the target through the replica partition and finds the stale copy, refusing — even though the admitting host IS the target and its authoritative record is current. `resource delete hosts udder --replica <root>` correctly refuses (origin connected), so the stale copy is also uncollectable by the operator path — but it should not need collecting: it should never shadow a live local record.

## What to fix

1. **Self-targeted admission consults the local authoritative record** when the admitting host is the placement target (or more generally: prefer the authoritative record when the daemon owns the target's identity).
2. **Stale self-origin replica copies must converge**: a live record's status updates should overwrite (or invalidate) the self-origin replica-partition copy rather than letting a pre-schema snapshot shadow current state. Related pathology: #1448/#1450/#1451 (self-origin residue for deleted records); this is the live-record variant.

## Acceptance criteria

- [ ] Self-targeted dispatch on a host with healthy published capacity admits (repro above passes)
- [ ] A stale self-origin replica copy cannot shadow a live local record in admission or list views
- [ ] InProcessDaemon-level test: admitting host == target host, replica partition holds a stale copy → admission uses the live record

rjwittams:
## What changed

- suppress self-origin replica rows when the same resource name has a live locally authoritative record
- suppress matching self-origin watch events while local authority exists
- make self-targeted capacity admission read the local Host status exclusively
- cover the convergence contract on both in-memory and SQLite stores and reproduce admission through `InProcessDaemon`

## Why

The overlay read layer emitted both a live local Host and a stale replica of that same Host from the daemon's own origin partition. Admission could then select stale pre-schema capacity data, and list/watch consumers saw duplicate records.

## Impact

Self-targeted placement uses current authoritative capacity. Stale self-origin replicas remain durable internally but cannot shadow or overwrite a live local record in list/watch views. Replicas from genuinely remote origins remain visible.

## Validation

Closes #1465

### DIFF
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -4961,9 +4961,6 @@ impl InProcessDaemon {
             return Ok(());
         };
         let target_host = &placement.target_host;
-        if self.local_host_id().as_ref().is_some_and(|host_id| host_id.as_str() == target_host.reference) {
-            return Ok(());
-        }
 
         let sources =
             self.resource_backend.including_replicas::<ResourceHost>(namespace).list().await.map_err(|error| error.to_string())?;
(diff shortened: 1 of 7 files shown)


rjwittams:
Thanks. I verified the fixture change against the full workspace suite: both in-process daemons inherit the same physical-machine fingerprint in this test process, so their derived Host/root identities collide despite separate temp config directories. Once self-origin shadowing was fixed, that made the purported remote replica correctly look self-origin and exposed the fixture's invalid topology. Giving the simulated peer an explicit `feta` origin and `feta-host-id` preserves the test's intended genuinely-remote setup; the targeted test and full workspace suite both pass with that correction.

PR #1473, opened 2026-08-13
fix: canonicalize placement host references

### LINKED ISSUE #1465: fix: self-targeted admission reads a stale self-origin Host replica instead of the live local record

rjwittams:
Closes #1465.

## What changed

- resolve a placement policy's host reference against Host resource identity as soon as the placement target is built
- preserve exact Host IDs, while accepting a unique display-name alias and converting it to the canonical Host ID
- reject unknown and ambiguous aliases instead of carrying hostname-like strings into admission
- add a regression matching the live topology: a UUID-keyed healthy local Host, a stale foreign replica, and a policy persisted with the local display name
- clarify that `resource delete --host` deletes the authoritative record on the selected peer, which is the cleanup path for incorrectly authored residue

## Live-store finding

On udder, the policy stored `host_ref = "udder"`, but the authoritative Host resource name is its UUID. The self-targeted admission check compared those strings directly, so the local-only capacity path never ran. The stale row visible on udder is foreign-origin and is not a failure of the self-origin suppression added in #1467.

Kiwi's local-only store was checked separately: it genuinely authors a Host row for udder under kiwi's node ID. That is residue, not list-envelope misclassification, and is deliberately not treated as admission authority by this change.

## Verification

- focused core, remote-placement, vessel-placement, and resource-backend regression suites

### DIFF
--- a/crates/flotilla-controllers/src/reconcilers/presentation.rs
+++ b/crates/flotilla-controllers/src/reconcilers/presentation.rs
@@ -31,6 +31,18 @@ use tracing::warn;
 
 type RegistryLookup = dyn Fn(&str) -> Result<Arc<ProviderRegistry>, String> + Send + Sync;
 
+fn canonical_host_ref(hosts: &[ResourceObject<Host>], host_ref: &str) -> Result<String, String> {
+    if hosts.iter().any(|host| host.metadata.name == host_ref) {
+        return Ok(host_ref.to_string());
+    }
+    let mut matching = hosts.iter().filter(|host| host.spec.display_name == host_ref);
+    let host = matching.next().ok_or_else(|| format!("host {host_ref} lookup failed: not found"))?;
+    if matching.any(|candidate| candidate.metadata.name != host.metadata.name) {
+        return Err(format!("host reference {host_ref} is ambiguous"));
+    }
+    Ok(host.metadata.name.clone())
+}
+
 #[async_trait]
 pub trait PresentationRuntime: Send + Sync {
     async fn apply(&self, plan: &PresentationPlan) -> Result<AppliedPresentation, ApplyPresentationError>;
@@ -232,6 +244,11 @@ impl<R> PresentationReconciler<R> {
         vec![Box::new(LabelJoinWatch::<TerminalSession, Presentation> { label_key: CONVOY_LABEL, _marker: PhantomData })]
     }
 
+    async fn canonical_host_ref(&self, host_ref: &str) -> Result<String, String> {
+        let hosts = self.hosts.list().await.map_err(|error| format!("host list failed: {error}"))?;
+        canonical_host_ref(&hosts.items, host_ref)
+    }
+
     async fn resolve_process(&self, session: &ResourceObject<TerminalSession>) -> Result<ResolvedProcess, String> {
         let environment = self
             .environments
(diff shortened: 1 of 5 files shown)


rjwittams:
Addressed the review finding in 64c1ca06. Host aliases now pass through the shared canonical resolver for placement admission, credential/readiness validation, host-direct adapter capabilities, remote host-direct routing, environment/session routing, and the vessel strategy derived from the placement decision. Added explicit unknown/ambiguous alias tests and a local-alias routing assertion. Focused routing/fleet/vessel regressions, full workspace tests, fmt, and Clippy pass locally.

rjwittams:
Folded both final review findings into b8b6cb08. Default placement now resolves every viable target before sorting and uses canonical Host IDs for the local tie-break and its explanation. Presentation resolves exact IDs or a unique display-name alias before local-vs-SSH routing. Added alias-specific regressions for both paths. Exact fmt, Clippy, and full workspace test gates pass locally.

rjwittams:
Fixed in the updated HEAD. Host resolution now returns `Result<Option<PlacementTargetHost>, String>`, so unknown hosts use a typed `None` branch and the environment-routing fallbacks no longer inspect error text. The viable-candidate pass converts unknown or ambiguous host resolution into a refusal for that policy while continuing to select clean candidates. Added a Tool-only/no-credential regression with one clean policy and one deleted-host reference, plus a direct test of the unregistered-host routing fallback. Exact fmt, Clippy, and full workspace gates pass locally.

rjwittams:
Fixed in the updated HEAD. `local_fleet_rows` now loads the replica-aware Host view once, reuses that snapshot for every session, and treats unknown or ambiguous per-row aliases as a raw-reference fallback instead of aborting the namespace projection. Added a regression with two Hosts sharing an alias plus an unrelated local session; both fleet rows remain visible and the affected row retains the raw alias. Exact fmt, Clippy, and full workspace gates pass locally.

rjwittams:
Addressed across the current HEAD: canonical Host resolution is shared by admission, readiness/credential checks, routing, and vessel/environment consumers, with unknown/ambiguous and live-topology regressions.

rjwittams:
Addressed across the current HEAD: prefer-local ordering and presentation attach routing both consume canonical Host identity, with explicit display-name alias regressions.

rjwittams:
Added the requested alias-specific vessel strategy regression in 2bafc02d. The policy carries display-name alias `udder`, the placement decision carries canonical reference `1c8df992`, and the test asserts `PlacementStrategy::canonicalize_host_ref` replaces the alias with the canonical reference. This is test-only; exact fmt, Clippy, and full workspace gates pass locally.

rjwittams:
Acknowledged: the sixth-pass review is clean and confirms the alias regression closes the final coverage gap. No further changes are needed.

PR #1479, opened 2026-08-13
fix: small batch — #1466, #1469, #1462

### LINKED ISSUE #1462: docs: document the structured log location (daemon.log is legacy)
## What to build

During the 2026-08-11 fleet bump we discovered `~/.local/state/flotilla/daemon.log` on feta had not been written since July — current daemons log structured JSONL to `~/.local/state/flotilla/log/flotillad.jsonl` (rotated: `.1`…`.4`), and the old file only receives whatever the launch command's stdout redirect happens to capture (startup errors, panics). Diagnosing "stale residue" errors in the old file cost real time.

Add a short "Logs" note to `docs/development.md`: where the structured logs live, the rotation scheme, that `daemon.log` is only the launch-redirect tail, and one example of filtering the JSONL (e.g. with `grep`/`jq` for level/target).

## Acceptance criteria

- [ ] `docs/development.md` documents both log locations and their roles
- [ ] One working JSONL filter example
- [ ] CI gates green

### LINKED ISSUE #1466: fix: resource apply/delete --host is accepted but silently applies locally
## Observed (2026-08-13)

`flotilla resource apply -f policy.json --host udder` (run on kiwi) reported success but created the record **locally on kiwi** (udder saw only a replica; its admission lookup then failed with `resource not found`). The `--host` flag parsed, did nothing, and gave no indication.

Consistent with the #1431 review note: only `ResourceStatusPatch` was added to the peer-forwarding allowlist; `ResourceApply`/`ResourceDelete` stay local by design — but then the CLI must not accept `--host` for them.

## What to fix

Either honor `--host` for apply/delete via the same forwarding path as `ResourceStatusPatch`, or reject the flag with a clear error for verbs that cannot route. Silent local fallback is the only wrong option.

## Acceptance criteria

- [ ] `resource apply --host <peer>` either creates the record on the peer's store or errors clearly
- [ ] Same for `resource delete --host`
- [ ] CLI test covering the routed/rejected behavior

### LINKED ISSUE #1469: flaky: tracked_repo_reconciles_generator_owned_project_fields_and_preserves_custom_fields fails on CI, passes locally
## Observed (2026-08-13)

`cargo test --workspace --locked` on CI (run 31730616503, PR #1468 — a **docs-only** change) failed:

```
thread 'tracked_repo_reconciles_generator_owned_project_fields_and_preserves_custom_fields' (8699)
panicked at crates/flotilla-daemon/tests/project_cli.rs:720:5
test result: FAILED. 31 passed; 1 failed
```

Locally on the same main: passes in isolation AND with the full `project_cli` binary (32/32, default parallelism). So the sensitivity is CI-environmental (timing, load, or shared-resource), not ordering within the file as reproduced locally.

Note the lineage: #1436 ("fix: stabilize tracked project log capture", merged 2026-08-13) already stabilized one environmental sensitivity in this same test file's domain. This is either an incompletely-fixed cousin or a new one at project_cli.rs:720 — the bare panic (assertion with no message reaching the log) makes CI triage harder than it should be, which is worth fixing along the way.

## What to build

Reproduce under constrained conditions (e.g. `--test-threads` sweep, cpu/io throttling, or repeated runs on a loaded machine), identify the shared state or timing assumption at project_cli.rs:720, and fix the test or the underlying race. Give the assertion a message so future CI failures are self-describing.

## Acceptance criteria

- [ ] Root cause identified (not just retried into passing)
- [ ] Test is deterministic under load/parallelism, or the product race it exposes is fixed
- [ ] The line-720 assertion carries a diagnostic message

rjwittams:
## Summary

- route `resource apply --host` and `resource delete --host` through the peer-forwarding path, with two-daemon coverage proving only the peer authority changes
- remove the `project_cli` logging race by initializing a real Git checkout before daemon startup, then scope log capture to the explicit reconciliation and add diagnostic assertion messages
- document structured daemon JSONL locations, rotation, filtering, and the legacy launch-redirect log

## Verification

- flaky test: 20 isolated runs plus the full `project_cli` binary with `--test-threads=16`

Closes #1466
Closes #1469
Closes #1462

### DIFF
--- a/crates/flotilla-daemon/src/server/remote_commands.rs
+++ b/crates/flotilla-daemon/src/server/remote_commands.rs
@@ -208,6 +208,8 @@ impl RemoteCommandRouter {
                         | CommandAction::CrewComplete { .. }
                         | CommandAction::CrewFail { .. }
                         | CommandAction::CrewHandoff { .. }
+                        | CommandAction::ResourceApply { .. }
+                        | CommandAction::ResourceDelete { .. }
                         | CommandAction::ResourceStatusPatch { .. }
                         | CommandAction::ResourceWatch { .. }
                 )
--- a/crates/flotilla-daemon/tests/project_cli.rs
+++ b/crates/flotilla-daemon/tests/project_cli.rs
@@ -18,7 +18,7 @@ use flotilla_core::{
         SOURCE_ENTRY_PATH_ANNOTATION, SOURCE_REPOSITORY_ANNOTATION, VERIFICATION_PROJECT_ANNOTATION,
     },
     project_declaration::{BOOTSTRAP_COMMIT_ANNOTATION, BOOTSTRAP_PATH_ANNOTATION, BOOTSTRAP_REPOSITORY_ANNOTATION},
-    providers::discovery::test_support::fake_discovery,
+    providers::discovery::test_support::{fake_discovery, git_process_discovery, init_git_repo_with_remote},
     repository_inspection::{LocalCheckoutInspection, ProjectDeclarationInspection, RepositoryInspection, RepositoryInspector},
 };
 use flotilla_daemon::runtime::{DaemonRuntime, RuntimeOptions};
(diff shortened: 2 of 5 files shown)


rjwittams:
Thanks. I checked the background path as part of the root-cause analysis: `InProcessDaemon::new_with_resource_backend_and_clock` inspects each startup checkout and seeds `repository_keys_by_path` when the checkout is a valid Git repository. Later, `poll_snapshots` only calls `reconcile_whole_repository_project` when the stored key differs or is absent; with the initialized fixture it instead takes the `ensure_repository` branch. That leaves the explicit `materialize_tracked_repo_projects` call as the sole project reconciliation and warning emitter in this test.

PR #1482, opened 2026-08-13
fix: keep failed convoys visible until reaped

### LINKED ISSUE #1476: Failed convoys vanish from the convoy view; their vessel/checkout rows remain orphaned
## Observed (2026-08-13)

convergence-harness went `phase: Failed` (correctly — it filed #1474 and could not meet acceptance). Afterward:

- `flotilla resource list convoy` shows the record, phase Failed.
- The TUI convoy view does NOT list it; `flotilla ls` shows only its vessel row (feta) with no convoy row — while every Active convoy shows both. Its checkout still appears in the checkouts list.

So the convoy projection/query appears to filter out Failed (state column renders as "active" for the surviving rows, suggesting a phase filter). The operator experience: the convoy disappears at exactly the moment it most needs attention, leaving orphan-looking vessel/checkout rows with no parent.

## Expected

Terminal-but-unreaped convoys (Failed especially, arguably Landed too) stay visible with distinct salience until torn down. Disappearance should only follow deletion/reaping. Related: #1470 (blocked-crew visibility) — same principle: states needing operator attention must not be the invisible ones.

## Acceptance criteria

- [ ] A Failed convoy appears in the TUI convoy view and `flotilla ls` with a distinct state rendering until reaped
- [ ] No resource rows (vessel/checkout/session) render parentless while their convoy record exists
- [ ] Test covering the phase-filter regression

rjwittams:
## What changed

- keep terminal Convoys in the store-backed Convoys result set instead of filtering them by phase
- retain soft-deleting Convoy rows until finalizer-driven reaping emits the actual delete event
- include every stored Convoy in projection parity monitoring
- make `flotilla ls` regression coverage assert a local Failed Convoy renders its failure state

## Why

The Convoys query filtered terminal phases and the Aggregator also dropped resources as soon as `deletion_timestamp` appeared. A Failed Convoy therefore vanished while its finalizers were still reaping vessels, checkouts, and sessions, leaving those child rows looking parentless at the point where operator attention mattered most.

Deletion timestamps begin teardown; they do not mean the resource has left the store. The projection now follows store lifetime and removes a Convoy only after the resource watch reports deletion.

## Impact

Failed and other terminal-but-unreaped Convoys remain visible in the TUI and `flotilla ls`. Failed rows retain their existing error-tone rendering and failure message. Child resource rows keep a visible parent throughout teardown.

## Validation

Closes #1476

### DIFF
--- a/crates/flotilla-core/src/aggregator_projection.rs
+++ b/crates/flotilla-core/src/aggregator_projection.rs
@@ -153,7 +153,7 @@ impl AggregatorProjectionState {
     }
 
     pub async fn result_set(&self) -> ResultSet {
-        hide_terminal_convoys(self.convoys.read().await.result_set())
+        self.convoys.read().await.result_set()
     }
 
     pub async fn convoy_result_set(&self, scope: &Option<QueryScope>) -> ResultSet {
@@ -168,7 +168,7 @@ impl AggregatorProjectionState {
     }
 
     pub async fn local_result_set(&self) -> ResultSet {
-        hide_terminal_convoys(self.convoys.read().await.local_result_set())
+        self.convoys.read().await.local_result_set()
     }
 
     pub async fn independents_result_set(&self, scope: &Option<QueryScope>) -> ResultSet {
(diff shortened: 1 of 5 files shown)


rjwittams:
Confirmed the steady-state behavior is intentional: terminal Convoys remain visible until explicit deletion, matching #1476's resource-lifetime contract. I also addressed the naming nit in ee618e92 by renaming the salience test so it no longer implies that it independently exercises the final `Deleted` event; `failed_convoy_remains_projected_until_reaping_finishes` retains that end-to-end lifecycle assertion.

rjwittams:
Reviewed the follow-up review. It confirms the requested behavior and the naming follow-up are correct, with no remaining action items.

PR #1484, opened 2026-08-13
feat: split portfolio DAG generated state

### LINKED ISSUE #1091: DAG board hardening: split generated state from authored judgment
## Learnings (2026-07-26)

The hand-maintained portfolio board (`~/dev/flotilla-tickets-dag.html`) rotted twice in one day in exactly the same way: **six tickets sat marked in-flight after their issues had closed**, because updates only touched tickets explicitly named in the edit. The rot is entirely in the *derivable* layer; the authored layer (what a decision means) was fine.

## Split

- **Generated** (deterministic fetch): status (landed / at-sea / ready / blocked), dependency edges, PR + CI state, convoy → host + staleness, labels → groups. All available from `flotilla ls --json`, `gh issue list`, and the native dependency API.
- **Authored** (human/agent): per-ticket `detail` cards and the `pulse` narrative. Not summarisable from tracker state — "docker is the new front" came from watching a dispatch fail five ways.
- **The payoff beyond less typing**: the generator can compare layers and mark a note **stale** when its ticket moved since the note was written. Today that drift is silent, which is the actual bug.

## v1 deliberately dumb

flotilla + GitHub (flotilla-org/flotilla and flotilla-org/andamento) only. **No** Forgejo, **no** convoy↔issue inference beyond exact matches. The point is to find out what a fetcher can honestly cover before investing more.

## Known-hard, do not paper over

1. **Convoy ↔ issue association is not reliably data** — convoy names are ad hoc (`tab-click-leak` for #1068); the join lives in the governor's head until #1037's association fact is stamped and readable.
2. **Cross-repo/cross-tracker state** — four trackers, different auth.
3. **The pulse cannot be generated.** Do not try.

This is also the concrete evidence-gatherer for #1090 (a portfolio artifact spanning trackers may not fit host-mastered resources at all) and a small instance of flotilla's own thesis: deterministic facts, templated presentation, authored judgment attached as annotation (#1089).

## Acceptance criteria

- [ ] `scripts/dag-fetch` writes the generated layer as JSON
- [ ] Board HTML merges generated + authored layers at load
- [ ] Notes whose ticket changed status since authoring render visibly stale
- [ ] Running the fetch twice with no fleet change produces identical output

rjwittams:
Closes #1091.

## Summary

- add `scripts/dag-fetch`, a deterministic materializer for GitHub issue/dependency/PR facts and exact `Convoy.spec.issues` associations
- add a board that merges generated facts with a separately authored pulse/detail/group/shape/edge layer and visibly marks notes when `statusAtAuthoring` drifts
- normalize fleet staleness to its semantic category and sort all output so unchanged observations serialize byte-for-byte identically
- document refresh, deployment, and safe authored-layer handling

## Status semantics

- closed issue → `landed`
- open issue carried by an Active convoy → `at-sea`
- open issue with an open native blocker or only non-Active convoy records → `blocked`
- remaining open issue → `ready`

## Legacy authored seed

The dispatch references the untracked personal file `~/dev/flotilla-tickets-dag.html`. It was not present in the vessel checkout, anywhere under the stated home path, or in reachable/dangling Git data, so this PR includes an explicit empty authored-layer template rather than inventing the operator's pulse or detail judgments. The deployment docs require migrating any populated personal authored layer before replacing the legacy HTML and explicitly warn against copying the empty template over it.

## Verification

- `python3 -m unittest scripts.tests.test_dag_fetch`
- `node scripts/tests/dag_board_test.mjs`
- live fetch twice with unchanged observations followed by byte-for-byte `cmp` (878 tickets, 42 dependency edges)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,6 +29,10 @@ jobs:
         run: |
           scripts/test-prune-target.sh
           scripts/test-cargo-sweep-mtime.sh
+      - name: Test portfolio DAG scripts
+        run: |
+          python3 -m unittest scripts.tests.test_dag_fetch
+          node scripts/tests/dag_board_test.mjs
       - name: Check Rust formatting
         run: cargo +nightly-2026-03-12 fmt --check
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,12 +29,17 @@ jobs:
         run: |
           scripts/test-prune-target.sh
           scripts/test-cargo-sweep-mtime.sh
-      - name: Test portfolio DAG scripts
-        run: |
-          python3 -m unittest scripts.tests.test_dag_fetch
-          node scripts/tests/dag_board_test.mjs
       - name: Check Rust formatting
         run: cargo +nightly-2026-03-12 fmt --check
+
+  dag-scripts:
+    name: DAG scripts
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - run: |
+          python3 -m unittest scripts.tests.test_dag_fetch
+          node scripts/tests/dag_board_test.mjs
 
   clippy:
     name: Clippy
--- a/docs/portfolio-dag.md
+++ b/docs/portfolio-dag.md
@@ -1,0 +1,71 @@
+# Portfolio DAG board
+
+The portfolio board keeps tracker and fleet facts separate from operator
+judgment:
+
+- `scripts/dag-fetch` replaces only the generated JSON layer. It reads GitHub
+  issues and pull requests for `flotilla-org/flotilla` and
+  `flotilla-org/andamento`, plus exact issue associations from Flotilla convoy
+  resources.
+- `scripts/flotilla-tickets-dag.authored.json` owns the pulse, detail notes,
+  display metadata for groups and shapes, and any edges that are judgments
+  rather than native GitHub dependencies.
+- `scripts/flotilla-tickets-dag.html` merges both layers in the browser.
+  Generated status always wins. A detail note with a different
+  `statusAtAuthoring` is rendered with a prominent stale warning.
+
+## Refresh and view
+
+The default generated output preserves the original desk-board location. Copy
+the board assets only after moving the legacy HTML's `pulse`, ticket `detail`
+cards, shapes, group presentation, and non-native edges into
+`flotilla-tickets-dag.authored.json`:
+
+```bash
+scripts/dag-fetch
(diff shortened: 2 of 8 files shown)


rjwittams:
Addressed the actionable findings in `82eb4166`:

- Fixed the replica tie-break to compare numeric `resourceVersion` values numerically, with a regression fixture covering `9` versus `10` at the same sync timestamp.
- Wired both focused Python and Node test suites into the CI Format job, so these non-Rust additions are now gated.
- Kept the default output path intentionally: the issue's dispatch note explicitly contracts that output be configurable while defaulting to the existing `~/dev/flotilla-tickets-dag` location. The script documents `--output` and does not touch the authored layer.
- Verified the requested `gh issue list` fields against the installed CLI help and through two successful live materializations across both repositories; the live repeatability check yielded byte-identical files.

The edge-source overwrite and DOM-test notes are non-blocking as observed: rendering does not consume edge provenance, while the extraction-free DOM helpers exclusively assign untrusted content through `textContent`.

rjwittams:
Applied all four non-blocking hardening notes in `98159ab9`:

- generated label groups now use a null-prototype object, with a regression assertion for a literal `__proto__` label;
- the focused tests now have their own `DAG scripts` CI job rather than reporting under Format;
- the unused `blocking` field was removed from the GitHub payload request;
- external `gh`/`flotilla` calls now fail explicitly after a five-minute bound instead of hanging indefinitely.

rjwittams:
Closed both latent edge cases in `b6cf54c7`:

- the synthetic ungrouped column now uses the internal key `__ungrouped__`, so a real `uncategorized` label remains a distinct group;
- convoy host now comes first from the resource's authoritative placement decision, while repeated `flotilla ls` rows are selected by semantic freshness (`local` > `current` > `stale`) with a deterministic fallback. The fixture covers conflicting stale/current rows and verifies that neither can override authoritative placement.

rjwittams:
Verified the non-blocking observations; no further code change is needed in this cycle:

- `groupEntries` is an array, so a real `__ungrouped__` label does not merge with or overwrite the synthetic entry; at worst the two independent sections share a cosmetic `data-group` value. This does not affect grouping or rendering correctness.
- A live all-issue query found maximum blocker and closing-PR counts of 2, with zero issues where `blockedBy.totalCount != blockedBy.nodes.length`, so the current portfolio has no nested-connection truncation.
- The leftover `blocking` keys are inert mock response data and usefully demonstrate that unrequested/unused API fields do not leak into output.

All seven checks on `b6cf54c7` are terminal with five passing and two intentionally skipped, and the PR is mergeable.

PR #1488, opened 2026-08-13
fix: make contained Claude OAuth crews headless

### LINKED ISSUE #1487: Claude OAuth consumer is not delivered to contained crews, and containerized claude is not headless

## What to build

## Acceptance criteria

rjwittams:
## Summary

- require a compatible granted Claude credential before admitting a contained `claude-code` crew
- seed Claude's non-secret onboarding and workspace-trust state before launch
- suppress the documented bypass-mode confirmation through the managed per-invocation settings overlay
- retain `CLAUDE_CODE_OAUTH_TOKEN` as process-only credential delivery and cover no-grant/default-deny behavior without printing material

## Root cause

OAuth material delivery already existed in the credential adapter, but admission allowed contained Claude sessions to start with no matching credential grant. Separately, a fresh isolated `CLAUDE_CONFIG_DIR` had no onboarding or workspace-trust state, so interactive first-run dialogs appeared before the prompt even when a token was available.

## Validation

- real-TTY Claude 2.1.223 probe with a fresh isolated config: reached the composer directly; an intentionally invalid token produced a named 401 at the prompt without onboarding, login, trust, or bypass-confirmation dialogs

Closes #1487

### DIFF
--- a/crates/flotilla-core/src/agent_adapter.rs
+++ b/crates/flotilla-core/src/agent_adapter.rs
@@ -327,6 +327,8 @@ pub struct AgentRequirement {
     pub model: Option<String>,
 }
 
+const CLAUDE_CODE_ADAPTER_ID: &str = "claude-code";
+
 pub struct CapabilityTable {
     requirements: BTreeMap<String, AgentRequirement>,
 }
@@ -371,7 +373,7 @@ impl AgentRequirement {
     /// contained Claude session must never fall through to interactive login.
     pub fn contained_credential_delivery_slot(&self) -> Option<&'static str> {
         match self.adapter.as_str() {
-            "claude-code" => Some("claude"),
+            CLAUDE_CODE_ADAPTER_ID => Some("claude"),
             _ => None,
         }
     }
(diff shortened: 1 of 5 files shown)


rjwittams:
Review-cycle update: tightened the headless-state seeding to the credential-provided `CLAUDE_CONFIG_DIR` only, so host-direct Claude sessions do not mutate a user's global `~/.claude.json`; added a regression test. I also centralized the `claude-code` adapter id used by discovery and the credential gate. The repeated spec listing is bounded admission overhead today and can be hoisted if placement candidate counts make it measurable.

rjwittams:
Addressed the host-direct side effect: headless state is now seeded only when discovery receives the isolated `CLAUDE_CONFIG_DIR` supplied by credential delivery, and a regression test proves a host-direct user's global `~/.claude.json` remains byte-for-byte unchanged. I also centralized the `claude-code` adapter id. The spec listing remains bounded admission overhead; hoisting can follow if candidate counts make it measurable.

rjwittams:
Thanks — confirmed. The only remaining observation is the explicitly non-blocking bounded listing overhead; no further code change is warranted for this bug fix.

PR #1492, opened 2026-08-14
fix: seed contained Claude from crew credentials

### LINKED ISSUE #1487: Claude OAuth consumer is not delivered to contained crews, and containerized claude is not headless

rjwittams:
## Summary

- seed Claude onboarding and workspace trust from the per-crew credential environment instead of ambient discovery state
- pass granted invocation environment into agent preparation before launch
- add a provisioning-level regression reproducing udder with no ambient `CLAUDE_CONFIG_DIR`, while asserting process-only OAuth and config delivery

## Root cause

The first fix discovered Claude state configuration before the granted crew environment existed. On udder, discovery had no ambient `CLAUDE_CONFIG_DIR`, so preparation skipped headless state seeding even though the credential adapter later selected an isolated directory for the crew process.

## Validation

Closes #1487

### DIFF
--- a/crates/flotilla-core/src/agent_adapter.rs
+++ b/crates/flotilla-core/src/agent_adapter.rs
@@ -13,7 +13,7 @@ use toml_edit::{value, DocumentMut, Item, Table};
 
 use crate::{
     path_context::ExecutionEnvironmentPath,
-    providers::{discovery::EnvironmentBag, ChannelLabel, CommandRunner},
+    providers::{discovery::EnvironmentBag, terminal::TerminalEnvVars, ChannelLabel, CommandRunner},
 };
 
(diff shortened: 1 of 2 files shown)


rjwittams:
Thanks — I checked the proposed contention case, and the current lock is already scoped to the set of paths that can contend. Each provisioned environment gets its own `ProviderRegistry` (and therefore its own Claude adapter and lock) in `probe_provisioned_environment`. Within one environment, `CredentialStore::prepare_scoped` rejects multiple credentials sharing the Claude delivery slot, so every Claude preparation in that registry resolves to the same credential-scoped `.claude.json`. Different Claude credentials necessarily belong to different provisioned environments and therefore different locks. No change needed.

PR #1494, opened 2026-08-14
fix: let contained Claude own headless config

### LINKED ISSUE #1487: Claude OAuth consumer is not delivered to contained crews, and containerized claude is not headless

rjwittams:
## Summary

- make the Claude AgentAdapter own contained config placement, seeding, launch export, and cleanup
- deliver Claude OAuth credentials as `CLAUDE_CODE_OAUTH_TOKEN` only, without treating `CLAUDE_CONFIG_DIR` as credential input
- route granted agent environments through the adapter launch plan and preserve generic delivery for tool sessions
- reproduce udder with no ambient or credential-supplied `CLAUDE_CONFIG_DIR`

## Root cause

The first two fixes still treated `CLAUDE_CONFIG_DIR` as an input. On udder, the contained environment exposed only `FLOTILLA_ENVIRONMENT_ID`, so the adapter neither seeded onboarding state nor exported a config path even though a matching OAuth grant existed.

Contained adapters now derive an isolated config directory beneath the vessel workspace, seed onboarding and workspace trust there, and force that path into the launch plan alongside the granted OAuth token. Trusted Claude sessions continue to respect ambient configuration.

## Validation

- targeted adapter, OAuth delivery, contained provisioning, and named missing-grant tests

Closes #1487

### DIFF
--- a/crates/flotilla-core/src/agent_adapter.rs
+++ b/crates/flotilla-core/src/agent_adapter.rs
@@ -312,6 +312,8 @@ pub struct AgentLaunchRequest {
     pub role: String,
     pub model: Option<String>,
     pub brief: TerminalBrief,
+    pub cwd: ExecutionEnvironmentPath,
+    pub environment: TerminalEnvVars,
 }
 
 #[derive(Debug, Clone, PartialEq, Eq)]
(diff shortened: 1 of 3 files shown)


rjwittams:
Fixed both findings in 60a8963b.

1. Contained Claude state now lives at `/run/flotilla/claude` inside the per-vessel container. Session cleanup no longer removes it, so concurrent or sequential Claude crew members share it safely for the environment lifetime; container teardown removes it once.
2. Host-direct OAuth launches retain account isolation with adapter-owned `/run/flotilla/claude-oauth/<account-key>` config directories. The account key is a truncated SHA-256 digest of the token, so credential material and reversible account identifiers do not enter paths. Explicit trusted `CLAUDE_CONFIG_DIR` still wins.

Added focused regressions for both lifecycle cases and reran format, Clippy, Claude-focused tests, and the full locked workspace suite.

rjwittams:
Both correctness findings are resolved and the refreshed checks are green. I filed the valid non-blocking rotation-cleanup observation as #1495 so reclamation can be designed without racing live host-direct Claude supervisors: https://github.com/flotilla-org/flotilla/issues/1495

PR #1502, opened 2026-08-14
fix: derive claude-oauth preflight dir from writable base

### LINKED ISSUE #1498: claude-oauth preflight stages config dir under /run/flotilla — unwritable for user-run daemons
## What happened

First contained-claude dispatch on udder (`claude-udder-smoke7`, the #1410 smoke walk) failed admission-side:

```
credential `claude-max` adapter `claude-oauth`: create writable config directory: mkdir: cannot create directory '/run/flotilla': Permission denied
```

`CredentialConsumer::ClaudeOauth` preflight (`crates/flotilla-daemon/src/credential.rs`, `claude_config_dir_fragment`) hardcodes the isolation config dir to `/run/flotilla/credentials/<name>/claude` and `mkdir -p`s it **on the host** via the runner, before any container exists. `/run` is root-owned tmpfs; every fleet daemon runs as an unprivileged user (nohup, no systemd `RuntimeDirectory=`), so this path is unwritable on every host we have. It has never been hit before because contained claude had never run anywhere — trusted-stance claude uses ambient auth with no staging, and the sibling `/run/flotilla/codex` constants are container-*internal* paths where `/run` is writable.

## What to build

Derive the host-side preflight scratch dir from a daemon-owned writable base instead of absolute `/run`:

- Prefer `XDG_RUNTIME_DIR` when present (via the injected `EnvVars`, never `std::env`), falling back to a state-dir location (e.g. under `~/.local/state/flotilla/`) the daemon already owns.
- Preserve the isolation property the hardcoded dir was buying: the preflight `claude -p` probe must not see ambient authentication (`apiKeyHelper`, stored logins) — an empty per-credential dir, wherever it lives.
- Leave container-internal `/run/flotilla/*` paths (codex home, agent-environment) untouched — they are correct in-container.
- Per #1494's banned test shape: no test may pre-set `CLAUDE_CONFIG_DIR` to make seeding work.

## Acceptance criteria

- [ ] Contained claude dispatch preflight succeeds on a host whose daemon runs as an unprivileged user process (no `/run` mkdir on the host).
- [ ] Preflight probe remains isolated from ambient claude authentication.
- [ ] Container-internal paths unchanged; codex staging unaffected.

Found on the #1410 smoke walk (gap eleven); blocks the contained-claude smoke and therefore the andamento governor (#1440).

rjwittams:
Closes #1498.

## What happened

The `CredentialConsumer::ClaudeOauth` preflight hardcoded its isolation config dir to `/run/flotilla/credentials/<name>/claude` and `mkdir -p`'d it on the host before any container exists. `/run` is root-owned tmpfs and every fleet daemon runs as an unprivileged user, so the first contained-claude dispatch on udder failed admission-side with `mkdir: cannot create directory '/run/flotilla': Permission denied`.

## What this does

- Replaces the hardcoded fragment with `CredentialStore::claude_preflight_config_dir`, which derives the scratch dir from a daemon-owned writable base: `$XDG_RUNTIME_DIR/flotilla/credentials/<name>/claude` when `XDG_RUNTIME_DIR` is present (read through the injected `EnvVars`, never `std::env`), falling back to `<state_dir>/credentials/<name>/claude` under the daemon state dir the daemon already owns.
- Preserves the isolation property the hardcoded dir was buying: the `claude -p` probe still runs under an empty per-credential config dir, so ambient authentication (`apiKeyHelper`, stored logins) cannot mask a dead token.
- Leaves container-internal `/run/flotilla/*` paths untouched: `CONTAINED_CLAUDE_CONFIG_DIR`, the codex home constants, and the agent-environment path are unchanged, and codex staging is unaffected.

## Tests

- Updated `claude_oauth_material_is_delivered_as_an_env_token_without_owning_agent_config` to assert the state-dir fallback path.
- New `claude_oauth_preflight_scratch_dir_prefers_the_user_runtime_dir_over_absolute_run` asserts the `XDG_RUNTIME_DIR` preference and that no call touches root-owned `/run/flotilla`.
- Updated the contained-claude launch test in `runtime.rs` to register the state-dir-derived mkdir; the contained launch still exports `/run/flotilla/claude` (adapter-owned, container-internal).
- No test pre-sets `CLAUDE_CONFIG_DIR` (per the #1494 banned test shape).

`cargo +nightly-2026-03-12 fmt --check`, `cargo clippy --workspace --all-targets --locked -- -D warnings`, and `cargo test --workspace --locked` all pass.

### DIFF
--- a/crates/flotilla-daemon/src/credential.rs
+++ b/crates/flotilla-daemon/src/credential.rs
@@ -555,6 +555,23 @@ impl CredentialStore {
             .unwrap_or_else(|| PathBuf::from(path))
     }
 
+    // The preflight scratch dir must keep the `claude -p` probe blind to
+    // ambient authentication (`apiKeyHelper`, stored logins) — an empty
+    // per-credential dir. The daemon runs as an unprivileged user with no
+    // systemd `RuntimeDirectory=`, so absolute `/run` is unwritable; derive
+    // the dir from a daemon-owned writable base instead: the user runtime dir
+    // when present, the daemon state dir otherwise (#1498).
+    fn claude_preflight_config_dir(&self, credential_name: &str) -> String {
+        let base = self
+            .env
+            .get("XDG_RUNTIME_DIR")
+            .map(|value| value.trim().to_string())
+            .filter(|value| !value.is_empty())
+            .map(|runtime_dir| PathBuf::from(runtime_dir).join("flotilla"))
+            .unwrap_or_else(|| self.state_dir.clone());
+        base.join("credentials").join(safe_component(credential_name)).join("claude").to_string_lossy().into_owned()
+    }
+
     async fn prepare_adapter(
         &self,
         name: &str,
@@ -686,8 +703,7 @@ impl CredentialStore {
                 if !runner.exists("claude", &["--version"]).await {
                     return Err("consumer binary is unavailable".to_string());
                 }
-                let config_dir_fragment = claude_config_dir_fragment(name);
-                let config_dir = config_dir_fragment.value;
+                let config_dir = self.claude_preflight_config_dir(name);
                 if !already_prepared {
                     runner
                         .run("mkdir", &["-p", &config_dir], Path::new("/"), &ChannelLabel::Noop)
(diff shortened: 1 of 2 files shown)


rjwittams:
Both minor notes addressed in 52581053:

- **Scratch-dir lifecycle**: the preflight now removes the scratch dir (`rm -rf` through the runner, best-effort with a warn on failure) as soon as the probe completes, on both the success and failure paths. This stops per-credential state accumulating under the persistent state-dir fallback and also strengthens the isolation property — every probe now starts from a genuinely empty directory rather than one carrying whatever a prior probe wrote. Removal is deliberately not a hard error: the probe verdict is already in hand, and failing credential delivery over scratch cleanup would be worse than a logged warning.
- **Blank `XDG_RUNTIME_DIR` coverage**: added `a_blank_user_runtime_dir_falls_back_to_the_daemon_state_dir`, exercising the whitespace-only value → state-dir fallback branch; the existing tests cover the present-and-valid and absent cases.

Issue #10, opened 2026-03-05
resolve_repo_root: handle bare repo case


rjwittams:
From PR #3 review: `git rev-parse --git-common-dir` on a bare repo returns the repo directory itself, so `parent()` gives the containing directory rather than the repo root. Currently only used for cwd auto-detect where bare repos are unlikely, but should be handled if we expand usage.

PR #1507, opened 2026-08-14
feat: compose portfolio facts into legacy DAG

### LINKED ISSUE #1504: Portfolio board: adopt the original DAG page as the renderer via a compose step; retire the kanban rewrite
## Parent

#1091 delivered the generated/authored data split (#1484) but replaced the presentation with a from-scratch label-kanban (no DAG, all landed tickets as cards). Ruled 2026-08-14: **do not rebuild the renderer** — the legacy hand-board is a proven renderer with a single data entry point (`const DATA = JSON.parse(#board-data)`), so the board becomes: original page kept byte-identical as a template, plus a compose step that recomputes its embedded `board-data` from the generated + authored layers. Prototyped and verified rendering 2026-08-14 (operator session, kiwi): the original page renders today's facts unchanged — fresh pulse date, statuses corrected by generated facts, live PR/convoy suffix lines on cards.

## What to build (mechanical repo-ification of the working prototype)

- Commit the legacy page as the board template (`board-data` blob emptied to a placeholder), preserving its rendering untouched.
- Add the compose step (script alongside `dag-fetch`) implementing the prototype's mapping: authored tickets are the curated set; column = generated status when it dominates (landed → landed, at-sea → flight) else authored stage; card title/detail from authored with a generated fact suffix (open PR + CI state, convoy name/host/phase); edges = authored judgment edges + native blockedBy edges between curated tickets; pulse/groups/columns/rollups/notes from authored.
- Add `title` per ticket to the authored layer (the migration in PR #1503 carried detail/stage/shape/dagLabel but dropped the hand-written card titles — restore from the legacy file or accept generated titles where absent).
- Retire the kanban page and its renderer module or cut them down to the merge/stale utilities the composer reuses; keep `dag-fetch` untouched; tests updated.
- Newly-landed-since-authoring tickets currently render as landed context nodes; acceptable for now — a synthetic "landed since <snapshot>" rollup is a refinement, not required here.

## Acceptance criteria

- [ ] `dag-fetch && dag-compose` (names flexible) produces a single HTML whose rendering is the legacy board's, with current facts embedded.
- [ ] Kanban presentation gone; no second board to compare against.
- [ ] Stale-note semantics preserved (authored stage/status vs generated status drives visible correction, as in the prototype's column logic).
- [ ] Board tests pass; fetch output unchanged.

## Blocked by

PR #1503 (populated authored layer).

rjwittams:
Closes #1504.

## Summary

- compose generated facts and authored judgment into the byte-preserved legacy DAG template
- preserve authored stages unless definitive landed/at-sea facts move a ticket, and append live open-PR/CI and convoy facts to cards
- combine authored judgment edges with native blockers between curated tickets
- retire the replacement kanban page/renderer and document the single-HTML refresh workflow

## Verification

- `python3 -m unittest scripts.tests.test_dag_fetch scripts.tests.test_dag_compose`
- live `scripts/dag-fetch` followed by `scripts/dag-compose` (191 curated nodes; template unchanged outside `board-data`)


rjwittams:
Fixed the cross-repository case in 3d0bd89b. Directly curated tickets outside the board repository now receive repository-qualified Mermaid node IDs, are marked external so the board-wide click handler cannot mislink them, and use the generated per-ticket URL when their title falls back to generated data. Native and authored edges use the same qualified-ID map. The regression fixture now curates `flotilla-org/flotilla#10` and `flotilla-org/andamento#10` together and verifies distinct nodes, the Andamento URL, and the cross-repo edge.

I did not escape authored `title`, `detail`, or `dagLabel`: those fields are trusted legacy presentation HTML, not tracker text. The landed authored layer deliberately contains `<a>` titles and `<br>` DAG labels, while notes contain `<b>`; the byte-preserved legacy renderer consumes all of these with `innerHTML`. Escaping them would break the required original rendering. Generated titles, URLs, PR facts, and convoy facts remain HTML-escaped. I documented this trust boundary in `docs/portfolio-dag.md`.

rjwittams:
No code change for the remaining point: the cited entries are intentionally DAG-only parked context, not silently lost cards.

The live authored contract distinguishes DAG grouping from board columns. `later` is a DAG group title ("Later horizons"), while the board columns come from `stages`, which deliberately contains `landed`, `flight`, `ready`, `grill`, `waiting`, and `design` but not `parked`. All 12 parked entries lack an authored card title; that is exactly why PR #1503 restored titles for 179 card-bearing tickets rather than inventing titles for the 12 DAG-only nodes. Eight numeric parked refs already receive generated fallback `card` objects from the current composer, but the byte-preserved renderer correctly does not place them because no parked column exists. Giving the four synthetic refs fallback cards would have the same non-effect.

Falling back to `dagLabel` would therefore neither populate a "Later horizons" column (there is no such column) nor restore legacy presentation; it would create new card data for nodes that were authored as DAG-only. Keeping them as `t.dag` nodes and omitting `t.card` preserves the original renderer/data split required by #1504. Quote-bearing `dagLabel` remains outside scope for the same reason noted in the review: current authored data has none, and the task requires leaving the proven renderer contract untouched.

PR #1509, opened 2026-08-14
fix: resolve OAuth scratch paths in runner environment

### LINKED ISSUE #1508: claude-oauth preflight trusts a stale XDG_RUNTIME_DIR; fall back when the base is missing or unwritable
## What happened

Smoke eight on udder (#1410, first dispatch on the #1502 fix) failed preflight:

```
credential `claude-max` adapter `claude-oauth`: create writable config directory: mkdir: cannot create directory '/run/user': Permission denied
```

#1502's staging-dir derivation prefers `XDG_RUNTIME_DIR`. The daemon inherits that env from whichever ssh login session launched it — but without `loginctl` linger, logind **removes `/run/user/<uid>` when the last session closes**, while the daemon's inherited env still points there. `mkdir -p` then tries to create the chain from root-owned `/run/user` and dies. The variable being *set* says nothing about the directory still *existing* for a long-lived daemon on a headless host.

## What to build

- Before choosing `XDG_RUNTIME_DIR` as the staging base, verify the directory exists and is writable (via injected collaborators, not `std::env`/direct fs assumptions in the decision layer where avoidable); if not, fall back to the state-dir location exactly as if the variable were unset.
- Evaluate at *use* time (each preflight), not once at daemon start — the directory can disappear mid-lifetime for the same reason.
- Log the fallback at debug with the rejected path so the choice is diagnosable.
- Test shape: env var set to a nonexistent path → preflight succeeds via state dir. (No test may pre-create the runtime dir to make the happy path pass — mirror of #1494's banned shape.)

Note: enabling linger on always-on hosts is the right infra hygiene independently (being requested for udder), but the adapter must not *require* it — the failure mode is silent, host-specific, and indistinguishable from #1498 to an operator.

## Acceptance criteria

- [ ] Preflight succeeds on a host whose `XDG_RUNTIME_DIR` points at a removed directory.
- [ ] Fallback is per-invocation, not cached from daemon start.
- [ ] Regression test with a dangling runtime-dir env var.

Gap twelve of the #1410 smoke walk; blocks the contained-claude smoke.

rjwittams:
## Summary

- extend the command-runner seam so each execution environment selects a writable Flotilla scratch base in its own filesystem world
- resolve Docker scratch space from the container user's writable `HOME`, falling back to `/tmp`, without passing daemon-host paths into the container
- validate host-direct `XDG_RUNTIME_DIR` at each Claude OAuth preflight and fall back to the daemon state directory when it is missing or unwritable
- cover contained preflight path selection, dangling runtime directories, and per-preflight re-evaluation

## Testing

Closes #1508

### DIFF
--- a/crates/flotilla-core/src/providers/environment/runner.rs
+++ b/crates/flotilla-core/src/providers/environment/runner.rs
@@ -1,4 +1,7 @@
-use std::{path::Path, sync::Arc};
+use std::{
+    path::{Path, PathBuf},
+    sync::Arc,
+};
 
 use async_trait::async_trait;
 use uuid::Uuid;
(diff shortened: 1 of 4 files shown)


rjwittams:
Thanks for the detailed review. No PR changes are needed for the acceptance criteria.

I filed #1510 for the valid out-of-scope SSH observation so remote direct environments can derive scratch policy from their own environment rather than a local daemon hint: https://github.com/flotilla-org/flotilla/issues/1510

The check-to-`mkdir` TOCTOU window is inherent here; a disappearance in that narrow interval still returns the existing bounded preflight error, while every later preflight re-evaluates the base.

PR #1514, opened 2026-08-14
fix: route container delivery through writable base

### LINKED ISSUE #1513: Eliminate absolute /run/flotilla container writes: all delivery paths ride the writable-base seam
## What happened (gap fourteen of the #1410 walk)

Smoke twelve on udder passed the GitHub App mint (#1511) and then failed on the next hardcoded absolute:

```
credential `github-crew-pr` adapter `github-app`: write shared Git config: mkdir: cannot create directory '/run/flotilla': Permission denied
```

`GIT_CONFIG_PATH` is the constant `/run/flotilla/credentials/gitconfig`, written through the **environment runner** (inside the container) and exported to the crew as `GIT_CONFIG_GLOBAL`. The udder crew image runs as a non-root user, so `/run` is unwritable — the same class #1509 fixed for the claude preflight scratch dir, at a different site. It passed compose-integration because those containers run as root.

## What to build

This is the **class** fix, not the site fix — one more site per cycle is not acceptable:

- Audit every absolute `/run/flotilla*` path that is *written through a runner* (as opposed to paths the image itself guarantees): known candidates are `GIT_CONFIG_PATH` (`credential.rs`), `AGENT_ENVIRONMENT_PATH` (`vessel_config.rs`), `CONTAINER_CODEX_HOME` (`agent_material.rs`), and the credential token-file paths composed near `credential.rs:100-130`. List the audit result on this issue.
- Route each write through the #1509 `writable_scratch_base` seam (or a sibling `writable_config_base` if lifetime semantics differ — the git config and agent-environment must **persist for the environment's lifetime**, unlike the preflight scratch which is removed after the probe; do not let a cleanup built for scratch delete a config the crew still needs).
- Every env var exported to the crew (`GIT_CONFIG_GLOBAL`, `CODEX_HOME`, the agent-environment path) must carry the path actually chosen by the seam, never the old constant.
- Non-root containers are the design case: the udder crew image's user is the reference environment.

## Testing (binding, per the 2026-08-14 ruling on #1512)

- **Validate in a real non-root container during implementation, before the PR.** docker is available on the crew host; run the delivery path against a real container with a non-root user and prove the writes land and the exported paths resolve. Mock-only coverage does not satisfy this contract.
- Regression tests: a runner-level test per migrated site asserting no absolute `/run/flotilla` write remains (grep-proof or call-recording), plus the non-root container behavior test.
- The PR description must state what was executed against a real container and what the observed result was.

## Acceptance criteria

- [ ] Audit list of all runner-written `/run/flotilla` absolutes posted on this issue; each migrated or justified.
- [ ] Contained delivery on a non-root container proceeds past git-config and agent-environment writes (real-container evidence in the PR).
- [ ] Crew-visible env vars carry seam-chosen paths.
- [ ] Persistence semantics: preflight scratch still cleaned; lifetime configs not deleted.

Blocks the #1410 contained-claude smoke and the #1440 governor behind it.

rjwittams:
Closes #1513

## Summary

- add a persistent `writable_config_base` runner seam, distinct from cleaned preflight scratch, with a pre-create-known Docker config base at `/tmp/flotilla-config`
- route Git config, Forgejo token/helper files, credential Codex homes, composed agent environment, and Claude OAuth mutable config through persistent seam-derived paths
- move leased Codex material to the sibling `/tmp/flotilla-codex` Docker seam path so Docker mount creation cannot make the config base root-owned; every crew-visible `CODEX_HOME`, `GIT_CONFIG_GLOBAL`, `FORGEJO_TOKEN_FILE`, and `CLAUDE_CONFIG_DIR` names the actual selected path
- preserve Codex claim conflict detection before container creation while resolving credential path values after the container runner exists
- keep Claude preflight scratch separate as `claude-preflight` and cleaned immediately; lifetime config paths are never attached to that cleanup

The full `/run/flotilla` runner-write audit is posted on #1513: https://github.com/flotilla-org/flotilla/issues/1513#issuecomment-5294215668 (with final sibling-path correction in the follow-up comment).

## Real non-root container validation

Executed before opening this PR and again after review expanded the scenario:

```text
cargo test -p flotilla-core --locked \
  providers::environment::runner::tests::persistent_delivery_paths_work_in_a_real_non_root_container \
  -- --ignored --nocapture
```

The test launched `busybox:latest` as UID/GID `65534:65534` with `HOME=/tmp`, `/run` unwritable, and a Codex material bind mount already present at `/tmp/flotilla-codex`. The Docker environment runner selected `/tmp/flotilla-config`, verified the Codex mount was writable, successfully wrote its Git config and agent-environment file as the non-root user, sourced the agent-environment file, and verified its exported `GIT_CONFIG_GLOBAL` resolved to the written file. Result: **passed**.

## Validation

- real non-root Docker test above

### DIFF
--- a/crates/flotilla-core/src/agent_adapter.rs
+++ b/crates/flotilla-core/src/agent_adapter.rs
@@ -535,8 +535,10 @@ impl AgentAdapter for CliAgentAdapter {
                 if *contained && !environment.iter().any(|(name, _)| name == "CLAUDE_CODE_OAUTH_TOKEN") {
                     return Err("contained Claude Code requires credential environment `CLAUDE_CODE_OAUTH_TOKEN`".to_string());
                 }
-                if *contained && !environment.iter().any(|(name, _)| name == "CLAUDE_CONFIG_DIR") {
-                    return Err("contained Claude Code requires seam-resolved environment `CLAUDE_CONFIG_DIR`".to_string());
+                if environment.iter().any(|(name, _)| name == "CLAUDE_CODE_OAUTH_TOKEN")
+                    && !environment.iter().any(|(name, _)| name == "CLAUDE_CONFIG_DIR")
+                {
+                    return Err("Claude Code OAuth requires seam-resolved environment `CLAUDE_CONFIG_DIR`".to_string());
                 }
                 let invocation_state = if let Some(config_dir) = self.claude_invocation_config_dir(environment) {
                     let config_dir_string = config_dir.display().to_string();
@@ -585,10 +587,11 @@ impl AgentAdapter for CliAgentAdapter {
         {
             return Err("contained Claude Code requires credential environment `CLAUDE_CODE_OAUTH_TOKEN`".to_string());
         }
-        if matches!(&self.flavor, AdapterFlavor::ClaudeCode { contained: true, .. })
+        if matches!(&self.flavor, AdapterFlavor::ClaudeCode { .. })
+            && env.iter().any(|(name, _)| name == "CLAUDE_CODE_OAUTH_TOKEN")
             && !env.iter().any(|(name, _)| name == "CLAUDE_CONFIG_DIR")
         {
-            return Err("contained Claude Code requires seam-resolved environment `CLAUDE_CONFIG_DIR`".to_string());
+            return Err("Claude Code OAuth requires seam-resolved environment `CLAUDE_CONFIG_DIR`".to_string());
         }
         if let Some(config_dir) = self.claude_invocation_config_dir(&env) {
             env.retain(|(name, _)| name != "CLAUDE_CONFIG_DIR");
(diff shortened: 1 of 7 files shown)


rjwittams:
Fixed all three findings in 2a2045f4.

The Docker contract now uses sibling paths: persistent runner writes live under `/tmp/flotilla-config`, while the pre-create Codex bind mount targets `/tmp/flotilla-codex`. Docker can therefore create the mount target without changing ownership of the config parent. The real non-root Docker regression now starts UID/GID 65534 with that Codex bind mount already present, verifies the mount is writable, then writes and resolves the Git config and agent-environment siblings successfully.

Host-direct Claude OAuth now rejects a token without a paired seam-resolved `CLAUDE_CONFIG_DIR`, with a regression test. `vessel_config_fragments_for_runner` also avoids resolving/probing the writable base when no Codex credential contributes a fragment.

rjwittams:
Confirmed: the split is intentional, and crew launch does not depend on the container's initial `docker run -e` value or on sourcing the staged file. `TerminalControllerRuntime::ensure_session` calls `CredentialStore::prepare_scoped` against the environment runner for each crew session and passes the returned seam-resolved `credential_env` into `prepare_with_environment` and `AgentLaunchRequest`; the resulting launch plan carries the real `CODEX_HOME` directly to the crew process. The staged agent-environment file is a matching persistent artifact for consumers that source it, while the container's long-lived `sleep infinity` bootstrap process has no need for credential-derived `CODEX_HOME`.

PR #1529, opened 2026-08-15
fix: readopt running environments after daemon restart

### LINKED ISSUE #1528: Daemon restart orphans running environments: no startup re-adoption into EnvironmentManager
## What happened (found by #1491's acceptance run, 2026-08-15)

Recursive attach from kiwi to the governor's session failed at the udder level with `environment not found: env-governor-govern` — while udder's own resource store holds the Environment record (with `docker_container_id`) and the container is running. The udder daemon has restarted several times since the environment was provisioned; `EnvironmentManager` is in-memory and registration happens at provision time only. Startup restores **adopted checkouts** (`reconcile_adopted_checkouts`, runtime.rs:462) but nothing re-registers running environments, so after any daemon restart every live environment becomes unresolvable — attach fails, and the TerminalSession reconciler emits the "provider registry unavailable for environment env-…" warnings seen all day for the smoke sessions.

This is the standing-convoy killer: **every fleet bump orphans the governor's environment**. It also retroactively explains at least part of the pre-#1526 bridge failures against udder ("docker environment 'env-…' did not become ready").

## What to build

- On startup (and on a periodic reconcile, mirroring the adopted-checkout pattern), re-register environments from the store: for each Environment record whose actuator is this host and whose backing (docker container, direct) is verifiably alive, rebuild its `EnvironmentManager` registration (bag, runner, provider registry).
- Dead backing (container gone) → drive the existing failure path, don't re-register.
- The reference reproduction is live on udder: the governor environment, store record present, container running, daemon freshly restarted. Validate against it (per the test-during-implementation ruling: real evidence in the PR, injected-collaborator tests for the logic).

## Acceptance criteria

- [ ] After a daemon restart, `flotilla attach` to a pre-existing environment's session succeeds without re-provisioning (the governor reproduction).
- [ ] TerminalSession reconciler stops warning "provider registry unavailable" for live environments post-restart.
- [ ] Dead-container environments fail their lifecycle properly rather than re-registering.
- [ ] Regression test: provision → simulated restart (fresh InProcessDaemon over same store/backing) → resolution works.

rjwittams:
## What changed

- reconcile durable, host-local Docker Environment records back into the in-memory EnvironmentManager during startup
- repeat environment adoption on the controller resync interval, mirroring adopted checkout restoration
- verify the durable container identity is present and running before rebuilding its bag, runner, and provider registry
- transition ready environments with missing or stopped backing to the existing Failed lifecycle state instead of registering them
- restore the active handle table so teardown continues to use the adopted handle

## Regression coverage

- provision/register an environment, construct a fresh InProcessDaemon over the same resource store and provider backing, then verify runner and registry resolution is restored
- verify a Ready environment whose recorded container is absent transitions to Failed and remains unregistered

## Live reproduction evidence

Before this fix on udder:

- the local resource store reports `env-governor-govern` as Ready with container `flotilla-env-env-governor-govern`
- `docker ps` reports that exact labeled container Up for 12 hours
- `flotilla attach governor` returns `environment not found: env-governor-govern`

This is the same durable-store/running-backing/fresh-manager split exercised by the restart regression.

## Validation

Closes #1528

### DIFF
--- a/crates/flotilla-daemon/src/runtime.rs
+++ b/crates/flotilla-daemon/src/runtime.rs
@@ -39,13 +39,13 @@ use flotilla_resources::{
     canonicalize_repo_url, clone_key, controller::ControllerLoop, descriptive_repo_slug, ChangeRequest, ChangeRequestStatus, Checkout,
     CheckoutBranchProvenance, CheckoutIntegrationStatus, Clone, CloneSpec, ConditionValue, Convoy, ConvoyReconciler, ConvoyTeardownRuntime,
     CredentialExpiry, CrewSource, CrewSpec, Demand, DemandKind, DemandSpec, DockerCheckoutStrategy, DockerPerVesselPlacementPolicySpec,
-    Environment, EnvironmentSpec, EnvironmentWaitReason, ForgeIdentity, Host, HostCondition, HostDirectEnvironmentSpec,
-    HostDirectPlacementPolicyCheckout, HostDirectPlacementPolicySpec, HostSpec, HostStatus, InputDefinition, InputMeta,
-    PlacementPolicySpec, Presentation, Project, Regard, ReplicaReadResolver, ReplicationClass, Repository, Resource, ResourceBackend,
-    ResourceError, ResourceObject, Stance, TerminalSession, TerminalSessionSource, Vessel, VesselRequirement, WorkflowTemplate,
-    WorkflowTemplateSpec, AGENT_ADAPTERS_CAPABILITY, CREDENTIAL_EXPIRY_CAPABILITY, CREDENTIAL_REFS_ENV, CREDENTIAL_REF_SESSION_TAG,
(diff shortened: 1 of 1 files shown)


rjwittams:
Fixed both findings. Each environment is now reconciled independently and concurrently under a 30-second timeout; probe, adapter-verification, registration, liveness, and timeout failures transition only that environment to `Failed` while live siblings continue adopting. Added a regression with one adapter-verification failure beside a healthy environment, and reran formatting, daemon clippy, and the full workspace suite.

rjwittams:
Fixed. Missing handles and definitive `Stopped`/`Failed` observations still drive the durable failure path, while transient `docker inspect` errors, `Building`/`Starting` states, and adoption timeouts now leave the Environment `Ready` for the periodic retry. Added a regression proving an inspect error does not irreversibly fail or register the environment. I also avoided replacing the active handle on every successful pass. Targeted tests and daemon clippy pass; the workspace rerun's unrelated temporary-fake-zellij `ETXTBSY` failure passed immediately when rerun in isolation.

PR #1531, opened 2026-08-15
fix: prefer authoritative host self-reports

### LINKED ISSUE #1524: Duplicate Host rows make admission read a stale null held_credentials row
## What happened (2026-08-15)

`convoy start` from kiwi with feta placement was refused: "workflow requires credential `claude-max`, which placement … host `feta` does not hold" — while feta's material, binary, and its *own* Host row (`held: ["claude-max","github-crew-pr"], ready: true`) were all fine. `resource list hosts` on kiwi shows **duplicate rows per host** — e.g. feta appears once with `held: null / ready: true` and once with the correct capabilities; udder and kiwi similarly doubled (kiwi three ways). Admission evidently read the stale null row. Dispatching the identical command *from feta* succeeded.

## What to establish

- Why paired/stale Host rows exist in a host's view (local vs replica generation mix after daemon restarts?) and which one admission's placement check consults.
- Placement must consult the authoritative row (freshest generation for the target host), or the duplicates must be impossible.
- The visible symptom class (dispatch refused with a claim the target host's own record contradicts) should name the record generation it used.

## Acceptance criteria

- [ ] One authoritative Host row per host in every host's view, or admission provably selects the authoritative one.
- [ ] Regression test: stale null-capability replica + fresh capability row → placement uses the fresh row.
- [ ] Cross-host dispatch (kiwi → feta placement) succeeds under the reproduction conditions.

rjwittams:
Closes #1524.

## What changed

- Select the authoritative Host self-report for placement admission across local and replicated same-name rows.
- Prefer generation-bearing reports, then the newest daemon start and heartbeat.
- Include the selected daemon generation in Host readiness and credential refusal messages.
- Cover both the focused stale-null/fresh-capability regression and a kiwi-to-feta `ConvoyStart` over the client protocol.

## Root cause

The transitional peer-summary path materializes a local Host row with adapter and terminal-pool capabilities, while resource replication separately delivers the target daemon's full Host self-report. The two rows share a Host identity. Credential admission used a local-only lookup, so it could read the summary row's null `held_credentials` instead of the placement host's generation-bearing self-report.

## Validation

### DIFF
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -1180,7 +1180,7 @@ fn canonical_placement_host_ref_from_sources(
     hosts: &[ReadResourceObject<ResourceHost>],
     host_ref: &str,
 ) -> Result<Option<PlacementTargetHost>, String> {
-    let exact = hosts.iter().find(|host| host.object.metadata.name == host_ref);
+    let exact = authoritative_host_source(hosts.iter().filter(|host| host.object.metadata.name == host_ref));
     let resolved = if let Some(host) = exact {
         host
     } else {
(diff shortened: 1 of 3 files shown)


rjwittams:
Verified. The three notes are non-blocking and I am keeping this bugfix focused on authoritative Host selection and the contradictory readiness/held-capability refusal class from #1524. Capacity already skips summary-only rows structurally, and malformed expiry metadata remains a distinct validation path. The two selection reads are admission-time and preserve the existing target-resolution interface; a broader read-path consolidation can be considered separately if profiling shows it is hot.

PR #1533, opened 2026-08-15
fix: correct CLI resource and convoy output

### LINKED ISSUE #1237: convoy start reports 'session already has a foreground client' instead of started-confirmation
Observed on `24e09f29` (2026-07-29): `flotilla convoy start --project flotilla --issue N --name X` from a non-interactive shell prints only

```
session terminal-X-work-coder already has a foreground client
```

where it used to print `convoy started: X`. The convoy starts fine (phase Active, crew Working) — the start flow's attach step finds the crew runner holding the session foreground and its report *replaces* the start confirmation, leaving the operator unsure the dispatch worked.

Expected: always confirm the start (`convoy started: X`); the attach outcome is secondary ("attach skipped: crew holds the session" or, post cleat-rollout, fall back to watch per cleat#178 semantics). Non-interactive invocations arguably shouldn't attempt attach at all.

### LINKED ISSUE #1497: resource list workflowtemplates errors: 'not enabled for overlay replication'
On main (8e45dd62), `flotilla resource list workflowtemplates` fails:

```
Error:
   0: invalid resource: WorkflowTemplate is not enabled for overlay replication
Location: src/main.rs:1184
```

Every other kind lists fine. WorkflowTemplate presumably being outside the overlay replication set shouldn't make *local listing* impossible — the daemon seeds these templates itself (`single-agent`, `single-agent-trusted`, `single-agent-contained`, …) and being unable to enumerate them from the CLI means an operator cannot discover which `--workflow` values exist (hit while debugging the contained-claude smoke on udder, where the workflow choice was exactly the question).

Expected: `resource list workflowtemplates` lists local records regardless of the kind's replication class.

rjwittams:
## Summary

- treat replica-inclusive list/get as local reads for resource kinds that do not replicate
- skip default convoy auto-attach when stdin is non-interactive while preserving interactive and explicit `--attach` behavior
- add regressions for non-replicated workflow reads and terminal-aware convoy auto-attach

## Verification

Closes #1497
Closes #1237

### DIFF
--- a/crates/flotilla-core/tests/in_process_daemon.rs
+++ b/crates/flotilla-core/tests/in_process_daemon.rs
@@ -793,6 +793,63 @@ async fn resource_list_and_get_queries_return_wire_json() {
 }
 
 #[tokio::test]
+async fn resource_list_and_get_queries_return_local_non_replicated_resources() {
+    let temp = tempfile::tempdir().expect("create tempdir");
+    let daemon =
+        InProcessDaemon::new(vec![], test_config_store(temp.path().join("config")), fake_discovery(false), HostName::local()).await;
+    daemon
+        .resource_backend()
+        .using::<WorkflowTemplate>("flotilla")
+        .create(
+            &InputMeta::builder().name("local-workflow".to_string()).build(),
+            &flotilla_resources::WorkflowTemplateSpec::builder().vessels(Vec::new()).build(),
+        )
+        .await
+        .expect("create workflow template");
+
+    let listed = daemon
+        .execute_query(
+            Command {
+                node_id: None,
+                provisioning_target: None,
+                context_repo: None,
+                action: CommandAction::QueryResourceList {
+                    namespace: "flotilla".to_string(),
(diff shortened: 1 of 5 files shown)


rjwittams:
Good catch on the causal gap. I replaced the flush-only change with the behavior suggested by #1237: a default convoy start now executes its returned attach plan only when stdin is a terminal. Non-interactive starts keep the successful `ConvoyStarted` result as their final output; interactive starts still auto-attach, and explicit `--attach` remains an override even without terminal stdin. This decision stays in the CLI after admission, so it does not change dispatch-regard semantics. A regression test covers all four cases (default interactive/non-interactive, explicit attach, and JSON), and the exact format, Clippy, and workspace test gates pass locally.

I left replica-inclusive watch behavior unchanged because `resource watch` defaults to local reads and the linked contract covers list/get.

rjwittams:
Updated the PR description to describe the shipped terminal-aware auto-attach behavior and its regression coverage. Thanks for catching the stale summary.

PR #1535, opened 2026-08-15
fix: resolve remote credential paths natively

### LINKED ISSUE #1495: Reclaim rotated host-direct Claude OAuth config directories
From PR #1494 review by @claude: https://github.com/flotilla-org/flotilla/pull/1494#issuecomment-5288525262

Host-direct Claude OAuth config isolation keys ephemeral directories beneath `/run/flotilla/claude-oauth/` by a non-reversible digest of the token. Rotating a named credential therefore selects a new directory while the previous directory remains until host reboot.

Define and implement a safe reclamation policy for superseded host-direct Claude OAuth config directories. Cleanup must not race live Claude supervisors or sessions and must not require credential material to appear in paths, logs, or resource-store records.

Context: PR #1494 deliberately leaves contained config lifecycle to per-vessel container teardown; this follow-up concerns only long-lived host-direct runtime state across credential rotations.

### LINKED ISSUE #1510: Resolve credential scratch paths from remote direct environments
From PR #1509 review by claude-review bot: https://github.com/flotilla-org/flotilla/pull/1509#issuecomment-5293550639

`SshCommandRunner` inherits the direct-runner scratch-base policy, but the preferred `XDG_RUNTIME_DIR` hint supplied by `CredentialStore` comes from the local daemon environment. The validation probe runs remotely and safely rejects missing paths, yet a coincidentally matching path can be accepted without being derived from the remote environment's own variables.

Thread the remote direct environment's scratch policy or environment values through the runner/environment seam so scratch selection is fully native to the remote filesystem world. Keep this separate from #1508, which fixes Docker placement and stale host-direct XDG paths and does not regress SSH behavior.

### LINKED ISSUE #1522: Cover the forward-provisioned ambient-only credential fallback
From PR #1521 review by @claude[bot]: https://github.com/flotilla-org/flotilla/pull/1521#issuecomment-5298057362

`ambient_credential_dependent_vessels` is currently forward-provisioned for a future adapter that has an ambient credential scope but no deliverable material form. Claude now has both an ambient scope and a delivery slot, so current admission always selects delivered material before this fallback can be reached.

Add a clarifying comment at the fallback seam and synthetic coverage for an ambient-only adapter/requirement so the expiry refusal path cannot silently bit-rot when the capability table evolves.

Context: PR #1521 intentionally retires ambient authentication for trusted Claude crews while retaining the generic ambient-only model for consumers without deliverable material.

rjwittams:
## Summary

- make `SshCommandRunner` resolve scratch and persistent config bases from the remote shell's `XDG_RUNTIME_DIR`, `XDG_STATE_HOME`, or `HOME`, without accepting daemon-host path hints
- lock in the post-#1514 Claude OAuth rotation policy: stable credential-name config paths are reused across refreshable material rotations, avoiding both orphan creation and cleanup races with live sessions
- model credential behavior on resolved agent requirements and cover the forward-provisioned ambient-only expiry refusal path with a synthetic capability

## Real SSH environment validation

Executed against the machine's real batch-mode SSH environment:

```text
cargo test -p flotilla-core --locked \
  providers::ssh_runner::tests::writable_bases_resolve_in_a_real_ssh_environment \
  -- --ignored --nocapture
```

The runner crossed an actual `ssh localhost` boundary, ignored deliberately invalid daemon-host hints under `/daemon-only`, selected the remote environment's `/run/user/1000/flotilla`, and verified that directory exists and is writable from the remote execution world. Result: **passed**.

## Validation

- real SSH environment test above

Closes #1510
Closes #1495
Closes #1522

### DIFF
--- a/crates/flotilla-core/src/agent_adapter.rs
+++ b/crates/flotilla-core/src/agent_adapter.rs
@@ -326,9 +326,23 @@ pub struct AgentLaunchPlan {
 pub struct AgentRequirement {
     pub adapter: String,
     pub model: Option<String>,
+    credential_policy: AgentCredentialPolicy,
 }
 
 const CLAUDE_CODE_ADAPTER_ID: &str = "claude-code";
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum AgentCredentialPolicy {
+    None,
+    DeliveredWithAmbientFallback {
+        slot: &'static str,
+        scope: &'static str,
+    },
+    #[cfg(test)]
+    AmbientOnly {
+        scope: &'static str,
+    },
+}
 
 pub struct CapabilityTable {
     requirements: BTreeMap<String, AgentRequirement>,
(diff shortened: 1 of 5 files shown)


rjwittams:
Fixed in f15b4144. Each remote candidate now proves or creates its own writable Flotilla directory before selection, so an unusable `XDG_STATE_HOME` falls through to `$HOME/.local/state/flotilla`. The new real-shell regression forces `XDG_STATE_HOME=/proc/flotilla-unwritable` and verifies the HOME fallback; the ignored real-SSH test still passes and selects `/run/user/1000/flotilla` on this host. Formatting, clippy, and the full workspace suite pass after the change.

rjwittams:
Confirmed. Sharing the selected base is the existing direct-runner contract (`writable_config_base` delegates to `writable_scratch_base` by default); callers own distinct descendants, and the trait docs explicitly forbid attaching scratch cleanup to a config path. Credential cleanup targets only the `claude-preflight` leaf, never the selected base. No code change needed for the non-blocking observation.

PR #1536, opened 2026-08-15
fix: protect standing convoy resurrection

### LINKED ISSUE #1523: Operator reap of an ensured convoy burns restart budget and stalls resurrection
## What happened (governor enactment, 2026-08-14)

The designed flow for redelivering credentials to a standing convoy is reap-and-let-the-ensure-resurrect (record-is-memory, ADR 0030). In practice, `convoy delete governor --force` was recorded by the ensure as `last_failure: "ensured convoy was reaped"`, incremented `restart_count` (to 7), and the ensure then **stopped recreating** — apparently over a restart budget — leaving the standing convoy absent until the operator deleted and re-applied the ConvoyEnsure resource itself. This happened twice in one enactment (once after a credential-grant fix, once after a git-auth fix).

## Expected

- An operator-initiated reap of an ensured convoy is a *request for resurrection*, not a failure: it should not count against the crash-restart budget, and recreation should be prompt (next reconcile, not a backoff tail).
- Budget exhaustion should be visible in `flotilla ls`/TUI attention (the ensure sat silently stalled with a stale `retry_at`).
- Bonus: a first-class verb (`flotilla convoy bounce <name>` or `ensure kick`) expressing "recreate with current grants/config", since credential changes only take effect at provisioning time.

## Acceptance criteria

- [ ] Reap of an ensured convoy leads to recreation without manual ensure surgery; crash-restart budget unaffected.
- [ ] Stalled/budget-exhausted ensures surface as attention.
- [ ] Regression test covering reap → resurrect.

### LINKED ISSUE #1530: Forced ensure teardown overrode 33 safety refusals and destroyed a healthy standing convoy
## Post-mortem (governor kill, 2026-08-15, from udder's flotillad.jsonl)

- ~11:0x: fleet bump (protocol 19, pre-#1528). Orphaned environment ⇒ `stale resourceVersion` conflicts + `provider registry unavailable` ⇒ convoy controller marks the **healthy, running** governor Failed.
- Ensure enters its forced-restart path (reap-then-start).
- `automatic convoy reclaim refused` ×33 from 10:54: "not safe to delete: Landed evidence is missing" — the delete gate demands settlement evidence **a standing convoy can never possess**.
- 11:22:06: `automatic convoy reclaim recovered after repeated refusal (refused_attempts: 33)` — the forced reap overrides the gate and **destroys the live container** with a working governor inside.
- The restart half then fails (workflow template transiently unresolvable in the same conflict window), stranding the ensure at restart_count 53.

## Defects to fix (distinct from #1523 and #1528, which are causes upstream)

1. **Refusal-override is the wrong recovery for standing convoys.** Repeated safety-gate refusals of an ensured/standing convoy must escalate (attention surface, hold) — never eventually force through. "An incomplete restart is acceptable; destroying a live governor is not."
2. **The delete gate's evidence model doesn't fit standing convoys** — no exit table means Landed evidence never exists; the gate needs a standing-convoy-aware criterion (e.g. "backing dead" or operator-forced), not a merge-evidence check that always fails then gets overridden.
3. **Failure diagnosis before teardown**: the ensure reaped based on a Failed phase caused by daemon-side resolution loss while the backing was verifiably alive (`docker ps` would have said so). Before forced teardown of a standing convoy, verify the backing is actually dead.

## Acceptance criteria

- [ ] A standing convoy whose phase goes Failed due to daemon-side resolution loss, with live backing, is never torn down — recovery restores resolution (per #1528) or escalates.
- [ ] Repeated reclaim refusals surface as attention and hold; no force-through path remains for ensured convoys.
- [ ] Regression test reproducing the sequence (restart-orphaned env → Failed phase → ensure restart) asserting the container survives.

rjwittams:
## Summary

- require provider-backed proof that a standing convoy's backing is dead before ensure teardown, and hold live/unknown backing behind persistent attention
- remove the ensure force-delete path and accept verified backing death as the standing-convoy reclaim criterion
- treat operator reap as prompt resurrection without incrementing crash restart count, with durable timed retries surviving daemon restart
- restore provisioned environment resolution before starting the ensure loop

## Tests

Closes #1530
Closes #1523

### DIFF
--- a/crates/flotilla-core/src/in_process.rs
+++ b/crates/flotilla-core/src/in_process.rs
@@ -46,17 +46,17 @@ use flotilla_resources::{
     CheckoutPhase as ResourceCheckoutPhase, CheckoutSpec as ResourceCheckoutSpec, CheckoutStatus as ResourceCheckoutStatus, Clock,
     ConditionValue, Convoy as ResourceConvoy, ConvoyEnsure, ConvoyEnsureSpec, ConvoyEnsureStatusPatch, ConvoyIssue, ConvoyPhase,
     ConvoyRepositorySpec, ConvoySpec, ConvoyStatusPatch, CredentialConsumer, CredentialGrant, CredentialSpec, CrewCompletionPending,
-    CrewSource, CrewWorkPhase, Demand as ResourceDemand, Environment as ResourceEnvironment, HoldAct, Host as ResourceHost,
-    HostDirectPlacementPolicyCheckout, HostDirectPlacementPolicySpec, HostStatus as ResourceHostStatus, InMemoryBackend, InputMeta,
-    InputValue, IntegrationCondition, IssueSnapshot, IssueSourceResolution, IssueSourceUnavailable, LifecycleAuthority,
-    ObservedCheckoutSpec as ResourceObservedCheckoutSpec, PlacementPolicy, PlacementPolicySpec, Presentation as ResourcePresentation,
-    Project, ProjectRepositoryRole, ProjectRepositorySpec, ProjectSpec, ReadResourceObject, Repository, RepositoryKey, RepositorySpec,
-    Resource, ResourceBackend, ResourceError, ResourceObject, ResourceProvenance, SettlementMode, SystemClock, TerminalAttentionState,
-    TerminalBrief, TerminalCrewContext, TerminalCrewMessage, TerminalSession as ResourceTerminalSession, TerminalSessionIdentity,
-    TerminalSessionPhase as ResourceTerminalSessionPhase, TerminalSessionSource, TerminalSessionStatus, TerminalSessionStatusPatch,
-    TurnDeliveryRung, UnmetSettlementExpectation, Vessel, WatchEvent, WatchStart, WorkCompletionAuthority, WorkPhase as ResourceWorkPhase,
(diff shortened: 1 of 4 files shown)


rjwittams:
Fixed the fail-closed gap: a backing Environment without a Docker container identity now explicitly holds teardown, and the provider-backed regression covers it. I also clear stale reclaim-refusal attention before any absent-convoy restart attempt, including failed attempts.

I retained the successful-inventory/not-listed case as verified dead: once a durable container identity is known, a successful provider list is the `docker ps` evidence requested by #1530, and absence from that inventory is positive evidence that the named backing is no longer running. Live, uncertain-status, missing-identity, remote, and provider-error cases all hold.

rjwittams:
Confirmed intentional. `restart_count` now counts terminal/crash episodes, not provisioning attempts within one episode: only `BackingOff` increments it. `Retrying` schedules durable retry using that episode count without turning workflow/provider resolution failures into additional crashes. This is also what keeps operator reap budget-free as required by #1523.

The resulting cadence is bounded rather than unbacked: a first automatic crash retries at 60s, later crash episodes retain exponential history up to 15 minutes, and operator-requested resurrection begins at 30s when its provisioning dependency remains unavailable. Reintroducing a per-attempt increment would recreate the reported backoff-tail/budget-burning behavior. The durable retry regression pins the intended operator case. The remaining test ideas are useful but non-blocking and don't indicate a current defect.

PR #1548, opened 2026-08-15
Add graceful daemon shutdown

### LINKED ISSUE #1546: Graceful flotillad shutdown verb: the deploy-time stop/restart seam
## Why (from the fleet pull/install review, 2026-08-15)

The fleet install proposal (project-map briefs/fleet-pull-install-handoff-2026-08-15.md) correctly refuses to kill an arbitrary daemon process from the installer and asks for a safe stop seam. Today every restart is `pkill -x flotillad` + nohup — survivable only because #1529 re-adopts environments, #1536 protects standing convoys, and cleat sessions outlive the daemon by design.

## What to build

A graceful shutdown verb (`flotilla daemon stop` or `flotillad --shutdown` via socket): finish in-flight resource writes, close peer connections with a Goodbye, drop the client socket cleanly, exit 0. Optionally `daemon restart` composing stop + spawn. The installer (and my own bump ritual) then has one safe command instead of pkill.

## Acceptance criteria

- [ ] Stop verb exits cleanly with no resource-store write truncation; restart re-adopts environments per #1529 (test at InProcessDaemon level plus one real restart).
- [ ] Standing convoys and cleat sessions demonstrably unaffected across stop/start.
- [ ] pkill no longer required anywhere in documented deploy flows.

rjwittams:
## Summary

- add a typed shutdown RPC and `flotilla daemon stop` command
- acknowledge shutdown before draining client, peer, HTTP, and networking tasks
- send peers an explicit shutdown goodbye and preserve daemon-backed state across restart
- switch the real-daemon integration restart seam and deployment documentation to graceful shutdown

## Validation

- real-binary smoke test confirming the stop acknowledgement, zero exit status, socket removal, and clean lifecycle marker

Closes #1546

### DIFF
--- a/crates/flotilla-client/src/lib.rs
+++ b/crates/flotilla-client/src/lib.rs
@@ -126,6 +126,18 @@ impl SocketDaemon {
         from_session_stateful_bounded(socket_path, session, None).await
     }
 
+    /// Ask the connected daemon to shut down gracefully.
+    ///
+    /// Success means the daemon accepted the request. The server sends this
+    /// acknowledgement before it closes listeners and drains active work.
+    pub async fn shutdown(&self) -> Result<(), String> {
+        let result = send_request(&self.session, &self.pending, &self.next_id, Request::Shutdown).await?;
+        match into_success_response(result)? {
+            Response::Shutdown => Ok(()),
+            other => Err(format!("unexpected shutdown response: {other:?}")),
+        }
+    }
+
     pub async fn connect_with_surface(socket_path: &Path, surface: SurfaceDeclaration) -> Result<Arc<Self>, String> {
         let session = connect_unix_message_session(socket_path).await?;
         from_session_stateful_bounded(socket_path, session, Some(&surface)).await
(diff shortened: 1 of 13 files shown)


rjwittams:
Thanks for checking the surrounding shutdown paths. I’m keeping both points as non-blocking tradeoffs in this change:

1. `Shutdown` deliberately uses the existing 15-second suppression window. The daemon has exited and may remain unavailable while a deployment replaces it, so immediate retries would recreate the reconnect storm the explicit goodbye is intended to avoid. A distinct duration can be tuned separately if deployment measurements show 15 seconds is too conservative.
2. The accept loop has stopped before the peer snapshot, so the remaining gap is limited to a connection already accepted and still completing its handshake. Closing those handshakes before sending goodbyes would require broadcasting the same shutdown signal that also closes established peer handlers, which would invert the required goodbye-before-close ordering. Making that guarantee airtight calls for a separate handshake-registration barrier and is outside this shutdown seam; established peers at the snapshot receive the explicit reason, while an in-flight handshake falls back to the existing disconnect/backoff behavior.

PR #1554, opened 2026-08-15
feat: stamp fleet candidate protocol version

### LINKED ISSUE #1551: Fleet candidate manifest carries the wire/protocol version
## Parent

Split from #1547 (item 1 of 3).

## What to build

The coordinated fleet candidate build (#1485) records source SHAs in each bundle's `manifest.json` but not the peer wire/protocol version the bundled binaries speak. Per-host explicit pulls guarantee mixed-generation windows; the 2026-08-15 protocol 19→20 window broke peering. Promotion and `fleet-install` need the manifest to say what protocol a generation speaks so wire-bumping generations can be flagged "update all hosts promptly".

Stamp the protocol version into the candidate manifest during the build. Derive it from the built artifact or its exact source (the binaries already report it — the client handshake error prints "speaks proto N") — not from a hand-maintained duplicate; a test must fail if the manifest stamp can drift from what the binaries actually speak.

Consumption (promotion tool comparing consecutive generations and flagging bumps, `fleet-install` display) lives in the promotion/install work and is out of scope here — this ticket only makes the field exist and be trustworthy.

## Acceptance criteria

- [ ] Candidate `manifest.json` carries the peer protocol version of the bundled binaries.
- [ ] The value is derived from the built source/artifact, with a drift-preventing test.
- [ ] Existing manifest consumers (`install-candidate.sh` validation) still pass.

## Blocked by

None - can start immediately

rjwittams:
## Summary

- report the peer protocol version from both Flotilla binaries' `--version` output
- derive `peer_protocol_version` in candidate manifests from the bundled executables
- reject candidate builds when the CLI and daemon disagree or omit the protocol stamp
- keep the installer contract covered with the extended manifest

## Testing

- `bash ci/fleet-candidates/test-workflow.sh`

Closes #1551

### DIFF
--- a/ci/fleet-candidates/build-candidate.sh
+++ b/ci/fleet-candidates/build-candidate.sh
@@ -37,6 +37,18 @@ sha256_file() {
   fi
 }
 
+read_protocol_version() {
+  local binary="$1"
+  local version
+  version="$("$binary" --version)"
+  if [[ "$version" =~ proto=([0-9]+) ]]; then
+    printf '%s' "${BASH_REMATCH[1]}"
+    return
+  fi
+  printf '%s --version did not report a peer protocol version: %s\n' "$binary" "$version" >&2
+  return 1
+}
+
 main() {
 flotilla_sha="$(require_sha FLEET_FLOTILLA_SHA "${FLEET_FLOTILLA_SHA:-}")"
 cleat_sha="$(require_sha FLEET_CLEAT_SHA "${FLEET_CLEAT_SHA:-}")"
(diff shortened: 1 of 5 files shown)


rjwittams:
No change needed for the non-blocking fixture note. The synthetic value is deliberately just an unknown manifest field for the current installer compatibility contract; the fixture contains no Flotilla binaries whose real protocol it could derive. Drift against the actual protocol is covered separately by `binaries_report_their_wire_generation_and_protocol_version`, while the production builder derives the manifest value by parsing both built artifacts. If the installer begins consuming this field, its contract fixture will need Flotilla artifact doubles and value-validation cases as part of that consumption work.

PR #1560, opened 2026-08-15
fix: refresh GitHub App token files

### LINKED ISSUE #1556: Standing vessels outlive their GitHub App token: file-based delivery with host-side refresh
## Observed (andamento governor, 2026-08-15)

The governor vessel got `GH_TOKEN` as a GitHub App installation token minted once at vessel start (16:20Z). Installation tokens live one hour; it expired at 17:20Z and nothing refreshes it — the git credential helper is `gh auth git-credential`, which reads the same dead env var. All GitHub access (API and git) 401'd from 17:20 until restart. The CredentialSpec lifecycle is `refreshable` (#1353 requires it for GithubApp), but refresh only happens at delivery time — a standing vessel never sees a re-mint.

A second consequence: permission changes to the App only appear on newly minted tokens, so a standing vessel can't pick up granted permissions within its lifetime either.

## What to build

Deliver refreshable material to contained vessels as a file, not a baked env value — the pattern the Forgejo path already uses. The vessel environment gets a token *file* path (e.g. `GITHUB_TOKEN_FILE`-style) mounted from a host-side location; the credential delivery machinery re-mints before expiry and rewrites the file; the in-vessel credential helper reads the file on every use. Env-var single-value convention holds: one well-known var naming the file.

## Acceptance criteria

- [ ] A standing vessel's GitHub access survives past the 1-hour installation-token lifetime without vessel restart.
- [ ] Refresh happens host-side; the vessel never holds App private-key material.
- [ ] `gh` and git both pick up the rotated token on next use (credential helper reads the file per invocation).
- [ ] Covered at the credential-delivery seam with an injected clock/minter — no live GitHub in the loop for the behavior tests; the real mint path stays covered by the existing recorded fixtures.

Related: #1353 (refreshable lifecycle requirement), #1505 (expiry conditions). Distinct from #1442 (message delivery to busy crews).

rjwittams:
## Summary

- deliver GitHub App installation tokens through a protected `GITHUB_TOKEN_FILE` instead of a baked `GH_TOKEN`
- make both `gh` and the Git credential helper read the file on every invocation
- refresh registered standing-vessel deliveries host-side five minutes before token expiry
- inject the clock and token minter for deterministic rotation coverage while retaining recorded coverage of the real mint path

## Testing

Closes #1556

### DIFF
--- a/crates/flotilla-daemon/src/credential.rs
+++ b/crates/flotilla-daemon/src/credential.rs
@@ -5,14 +5,15 @@ use std::{
     sync::Arc,
 };
 
-use chrono::{DateTime, Utc};
+use async_trait::async_trait;
+use chrono::{DateTime, Duration, Utc};
 use flotilla_core::providers::{
     discovery::{EnvVars, EnvironmentBag},
     ChannelLabel, CommandRunner, HttpClient, ReqwestHttpClient,
 };
 use flotilla_resources::{
-    CredentialConsumer, CredentialExpiry, CredentialLifecycle, CredentialSource, CredentialSpec, CredentialSpecSpec, Repository,
-    RepositoryKey, ResourceBackend, ResourceError, AMBIENT_CLAUDE_CREDENTIAL_SCOPE,
+    Clock, CredentialConsumer, CredentialExpiry, CredentialLifecycle, CredentialSource, CredentialSpec, CredentialSpecSpec, Repository,
+    RepositoryKey, ResourceBackend, ResourceError, SystemClock, AMBIENT_CLAUDE_CREDENTIAL_SCOPE,
 };
 use jsonwebtoken::{Algorithm, EncodingKey, Header};
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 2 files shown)


rjwittams:
Addressed all three points:

- Fixed the in-flight refresh race with a per-registration generation. Initial delivery writes and registration are serialized with refresh commits; after minting, a refresh rechecks the current generation while holding that lock before writing. The new `stale_in_flight_refresh_cannot_overwrite_a_reprepared_delivery` test controls the mint interleaving and proves stale material is never written.
- Replaced the eight-argument constructor and clippy suppression with a `GithubAppMinting` collaborators value.
- Filed #1563 to make final file permissions part of atomic secure writes for both GitHub App and Forgejo delivery.

The post-fix pinned format, workspace clippy, and full locked workspace test gates all pass locally.

rjwittams:
Verified. The mint itself remains outside the mutex, but initial preparation deliberately holds the global delivery lock across file/helper setup and the `gh api` preflight so the file write and registration are ordered against refresh commits. That is bounded but can serialize unrelated initial preparations as noted.

Filed #1564 to move this coordination to per-registration/per-key locking while retaining the same-key generation invariant and concurrency coverage. Keeping that scalability refinement separate leaves this PR focused on the standing-token correctness bug.

PR #1561, opened 2026-08-15
fix: detect hosts without sleep capability

### LINKED ISSUE #1549: Sleep inhibition wrongly required on container hosts, permanent DEGRADED
## Observed

udder (Proxmox CT 200) reports permanently:

```
⚠ DEGRADED: SleepInhibition: sleep inhibition required but not held (all system sleep inhibitor strategies failed:
systemd-inhibit: ... Failed to inhibit: Access denied; kde-inhibit: No such file or directory; ...)
```

The host is an LXC container — it cannot sleep, and `systemd-inhibit` inside the CT is denied by design. Meanwhile feta's row shows `not required`, so a requiredness determination exists; udder is misclassified as sleep-capable.

## Expected

The requiredness probe should detect environments where host sleep is not a thing (container/CT/VM without sleep capability — e.g. no `/sys/power/state` writable targets, systemd `SleepOperation` empty, or virtualization detection) and mark inhibition `not required` rather than attempting strategies and pinning the host DEGRADED forever. A host that genuinely should inhibit but can't should still degrade.

## Acceptance

- udder-class CT hosts report `not required`, no DEGRADED.
- A sleep-capable host with all strategies failing still reports DEGRADED.
- Determination covered by behavior tests with injected probes.

Related: #1425 (quiescence keying — orthogonal; that governs *when* to inhibit, this governs *whether the host can sleep at all*).

rjwittams:
## Summary

- probe whether the daemon host can sleep before applying workload-driven inhibition
- classify Linux containers and hosts without kernel sleep states as `not required`
- preserve conservative inhibition when the capability probe fails
- cover both incapable-host and capable-but-failing behavior with injected probes

## Root cause

The controller derived requiredness only from active convoys and vessels. It never gated that demand on whether the host had a usable system sleep capability, so container hosts repeatedly attempted inhibitor strategies and became permanently degraded.

## Validation

Closes #1549

### DIFF
--- a/crates/flotilla-daemon/src/sleep_inhibitor.rs
+++ b/crates/flotilla-daemon/src/sleep_inhibitor.rs
@@ -23,8 +23,27 @@ const FAILED_RETRY_INTERVAL: Duration = Duration::from_secs(300);
 const ACQUISITION_CONFIRMATION: Duration = Duration::from_millis(250);
 const KDE_INHIBITION_ENFORCEMENT_DELAY: Duration = Duration::from_millis(5_250);
 const KDE_VERIFICATION_TIMEOUT: Duration = Duration::from_secs(3);
+#[cfg(target_os = "linux")]
+const CAPABILITY_PROBE_TIMEOUT: Duration = Duration::from_secs(2);
 const FAILURE_THRESHOLD: u32 = 3;
 const INHIBITOR_REASON: &str = "Flotilla vessel crew is active";
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+enum SleepCapability {
+    Available,
+    Unavailable,
+}
+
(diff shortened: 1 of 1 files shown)


rjwittams:
Addressed both coverage gaps in 6f0af48c. The controller behavior suite now injects a failing capability probe and verifies the conservative `Held` fallback. Linux classification is also separated into a pure decision function with cases for detected containers, absent and empty kernel sleep states, and available sleep states. The focused daemon suite, workspace clippy, and sandbox-safe workspace tests pass; one unrelated temporary-executable `Text file busy` flake passed on focused rerun and the complete workspace rerun then passed.

rjwittams:
No further changes needed from this re-review. The updated commit is green across all required checks and the working tree matches the pushed branch.

PR #1562, opened 2026-08-15
fix: recover contained CLI after daemon upgrade

### LINKED ISSUE #1557: Contained CLI cannot recover from daemon upgrade: file bind mount pins deleted inode, re-exec ENOENTs
## Observed (andamento governor, 2026-08-15, proto 19→20 fleet bump)

After the host daemon upgraded, every daemon-backed verb in the vessel failed with `No such file or directory` at the re-exec seam (src/main.rs:609 area). Diagnosis: the contained CLI (old build, proto 19) connects fine, the proto-20 daemon refuses the handshake, and the CLI's incompatible-daemon recovery re-execs `current_exe` — but the flotilla binary is a read-only single-file bind mount whose host inode was deleted when the upgrade `mv`'d the new binary into place. `/proc/self/exe` points at a deleted inode → exec ENOENT. Even if the exec succeeded, the matching build doesn't exist in the container: a file bind mount pins the old inode forever.

## What to build

1. Mount the CLI into vessels via a directory mount (or equivalent) so a host-side atomic replace is visible in the container, and the re-exec recovery actually lands on the new build.
2. Surface the failure legibly: when re-exec fails, report "daemon speaks proto N, this CLI speaks M, and re-exec could not reach a matching build" instead of a raw ENOENT.

## Secondary (verify while in there)

The daemon rewrote `/run/flotilla-daemon/socket-path` to its host-local path (`/home/robert/.config/flotilla/run/flotilla.sock`) — meaningless inside the container. Contained clients should not be handed host-local canonical paths (they connect via `FLOTILLA_DAEMON_SOCKET`, which was correct throughout); verify nothing consumes the advertised path in-container and fix the advertisement for contained consumers.

## Acceptance criteria

- [ ] A daemon upgrade + restart leaves a standing vessel's CLI able to reach wire-lockstep without vessel recreation.
- [ ] Version-mismatch-with-failed-recovery produces a message naming both builds/protocols.
- [ ] Behavior test at the environment-actuation seam covering binary replacement visibility (injected runner; a real-container test may confirm).

rjwittams:
Closes #1557

## What changed

- deliver the contained Flotilla CLI through a directory bind mount, with a stable launcher at `/usr/local/bin/flotilla`, so host-side atomic binary replacement is visible to standing vessels
- preserve the daemon/client build IDs and protocol versions when incompatible-daemon re-exec fails, and explain that no matching build could be reached
- publish daemon socket discovery paths relative to the shared runtime directory when possible, while resolving them to absolute paths for SSH peer dialers
- add behavior coverage at the injected Docker runner seam for replacement-visible mounts

## Root cause

A single-file bind mount pinned the old CLI inode inside a vessel. Replacing the host binary left `/proc/self/exe` referring to a deleted inode, so protocol-mismatch recovery failed with raw `ENOENT`. The shared socket discovery file also advertised a host-only absolute path inside the mounted runtime directory.

## Validation

- `cargo test -p flotilla-daemon --lib --locked`

### DIFF
--- a/crates/flotilla-core/src/providers/environment/tests.rs
+++ b/crates/flotilla-core/src/providers/environment/tests.rs
@@ -570,6 +570,47 @@ async fn create_delivers_tool_assets_and_applies_tool_environment() {
 }
 
 #[tokio::test]
+async fn create_mounts_the_flotilla_binary_directory_so_atomic_replacements_stay_visible() {
+    use flotilla_protocol::ImageId;
+
+    let runner = Arc::new(RecordingRunner::new_ok("container-id-123"));
+    let provider = DockerEnvironmentProvider::new(runner.clone());
+    let image = ImageId::new("ubuntu:22.04");
+    let tool = EnvironmentTool::new("flotilla", "/usr/local/bin/flotilla")
+        .with_asset(EnvironmentToolAsset::new(
+            "/host/flotilla/bin",
+            "/opt/flotilla/bin",
+            EnvironmentToolAssetKind::Directory,
+            EnvironmentToolAssetAccess::ReadOnly,
+            "the flotilla CLI",
+        ))
+        .with_asset(EnvironmentToolAsset::new(
+            "/host/flotilla/launcher",
+            "/usr/local/bin/flotilla",
+            EnvironmentToolAssetKind::File,
+            EnvironmentToolAssetAccess::ReadOnly,
+            "the flotilla CLI launcher",
+        ));
+    let opts = CreateOpts {
+        tokens: Vec::new(),
+        working_directory: None,
+        provisioned_mounts: Vec::new(),
+        tools: vec![tool],
+        image_pull_policy: ImagePullPolicy::IfNotPresent,
+        docker_config_dir: None,
+    };
+
+    provider.create(EnvironmentId::new("upgrade-visible"), &image, opts).await.expect("create environment");
+
+    let args = &runner.calls()[0].1;
+    assert!(args.contains(&"/host/flotilla/bin:/opt/flotilla/bin:ro".to_string()), "binary parent must be the bind source: {args:?}");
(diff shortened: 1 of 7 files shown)


rjwittams:
Addressed both points in `44bcb3b8`.

The daemon now copies only the validated adjacent `flotilla` binary into a dedicated state directory and atomically renames it into place at startup. Standing vessels mount that dedicated directory, so upgrades remain visible without exposing unrelated siblings from a general-purpose install or `target/debug` directory. A regression test verifies that unrelated source-directory files are not staged and that a subsequent staged replacement becomes visible at the stable path.

Launcher publication now also uses a uniquely named temporary file, sets its executable mode before publication, and atomically renames it into place. The launcher path and content remain stable for existing file mounts.

rjwittams:
No code change here because startup-only staging is the intended lockstep boundary for this issue.

The acceptance scenario is explicitly “a daemon upgrade + restart,” and the observed mismatch arose after the proto-20 daemon replaced/restarted the proto-19 daemon. `EnvironmentToolProvisioner::for_local_host` runs for that new daemon runtime and atomically stages the CLI adjacent to the newly running daemon, so every standing directory mount sees the new CLI inode after the rename.

Refreshing from the live host path during each later `prepare()` would weaken that invariant: if only the CLI were replaced while the old daemon remained running, newly provisioned vessels could receive a newer CLI that does not speak the running daemon’s protocol. Keeping the staged CLI generation tied to the daemon runtime means contained clients and that daemon remain wire-compatible for the daemon’s lifetime; the next daemon restart advances the staged generation for standing and new vessels together.

The re-exec path does not imply CLI-only upgrades are supported. It is recovery for clients surviving the daemon upgrade/restart described by #1557. A periodic watcher would add a second upgrade lifecycle outside this issue and could deliberately break daemon/CLI lockstep.

Issue #1259, opened 2026-07-29
Fleet releases slice 2: release pin file + host updater pull loop


rjwittams:
## Parent
#917 (ruled design). 

## What to build
- A **release pin file** schema in project-map (`fleet-release.toml` or similar): per-repo artifact versions + `store_epoch` + release id. Assembled by hand or a latest-green helper for now.
- A **host updater**: watches the pin (level-triggered, manifest-loop pattern), fetches artifacts for its platform from the forge, verifies checksums, rename-dances binaries into place, restarts the daemon into the new binary, and reports convergence state. Handles the mac codesign-preservation constraint (artifacts arrive signed from comte; updater must not strip attributes).
- `store_epoch` handling: on epoch change, snapshot the resource store to a dated `pre-*` dir before daemon start (the fleet's existing manual ritual, mechanized).
- Pre/postflight gates per the #917 ruling: preflight (checksums, binary executes and self-reports the pin's wire generation, image present per policy); postflight (daemon heartbeating on new generation, epoch matches, capabilities not regressed vs previous generation, replication cursors advancing). Fail → stay on previous release with a legible reason.
- Fleet table: `Version` column becomes release id + converged/converging/failed state.

## Acceptance criteria
- [ ] Editing the pin in project-map converges all three hosts with zero ssh from the operator
- [ ] A deliberately-corrupt artifact fails preflight; host stays on previous release and says why
- [ ] An epoch bump produces snapshot+fresh store; managed objects reappear via manifests
- [ ] Capability regression or zero-cursor replication fails postflight visibly
- [ ] `flotilla fleet` shows per-host release id and convergence state

## Blocked by
- #1258

rjwittams:
Re-scoped by the fleet pull/install review (2026-08-15): the first slice is **explicit pull only** — the project-map handoff (briefs/fleet-pull-install-handoff-2026-08-15.md) proposes `fleet-install status|latest|<generation>|rollback` with immutable Forgejo generations, atomic current/previous switching, and no timer/reconciler/auto-canary. This ticket's 'host updater pull loop' becomes a **later slice on top of that proven command**, not the first cut. Prerequisites filed: #1546 (graceful daemon stop seam), #1547 (manifest wire version, skew visibility, Darwin install-time signing). First adopter should be udder (always-on, no dev churn); Kiwi last (dev builds legitimately shadow generations there).

PR #1567, opened 2026-08-15
feat: add promoted fleet installer

### LINKED ISSUE #1565: fleet-install skeleton: explicit pull/install of promoted generations (Linux first)
## Context

First usable slice of the fleet pull/install path, per project-map `briefs/fleet-pull-install-handoff-2026-08-15.md` (endorsed in `briefs/fleet-pull-install-review-2026-08-15.md`). Prerequisites landed: graceful daemon stop (#1546) and `peer_protocol_version` in the candidate manifest (#1551). Work from a clean worktree off current main.

## What to build

`scripts/fleet-install`, bootstrapped to `~/.local/bin/fleet-install`, consuming immutable promoted generations from the Forgejo Generic Package `lab/flotilla-fleet` (endpoint `/api/packages/lab/generic/flotilla-fleet/<generation>/<file>`).

Interface: `fleet-install status | latest | <generation> | rollback`.

- `status` is read-only: local current generation vs promoted latest, including the manifest's `peer_protocol_version` and a wire-bump warning when current and target differ.
- `latest` resolves once to an exact generation id, prints the current→target transition, then uses only the immutable id.
- Layout: `~/.local/opt/flotilla-fleet/releases/<generation>/{bin,lib,manifest.json}`, `current`/`previous` symlinks, stable launchers in `~/.local/bin` exec'ing `current/bin/<name>`. Download to a temp dir on the same filesystem, verify outer digest + full inner manifest, stage completely, switch only the `current` symlink atomically. Never write into the selected generation.
- Daemon preflight: detect a running flotillad and stop it via `flotilla daemon stop` (the #1546 verb) before switching; fail with a clear instruction if stop fails. Rollback = same preflight, then switch `current` back to `previous`.
- Credential: package-read-only token read from a `0600` file under `~/.config/flotilla/`; never on the command line or in the repo.
- Darwin: **fail closed** — refuse to install with a message pointing at #1553. Linux (`linux-x86_64-gnu2.36`) is the deliverable.

## Contract tests (in-repo, no live Forgejo)

Exercise the HTTP/package contract against a fake Forgejo/package server: generation selection, platform selection, authenticated download, manifest/digest rejection, interrupted download, atomic switch, exact-generation reinstall, PATH shadow detection, rollback, and daemon-preflight refusal. Producer and consumer share the manifest contract (#1485/#1554 stamping).

## Out of scope

Fleet-spread reporting (#1552), Darwin signing (#1553), automatic convergence (#1259), R2 replication, promotion tooling (`lab-fleet-promote` lives in project-map).

The standing-vessel CLI replacement problem was tracked separately in #1557. Its mount/re-exec fix is not part of this installer; rollout validation should nevertheless prove that switching a host generation does not strand a standing vessel.

## Acceptance criteria

- [ ] All four commands implemented with the layout and atomicity rules above; contract tests pass in CI.
- [ ] `status` surfaces wire-version difference between current and target.
- [ ] Darwin fails closed with a pointer to #1553.
- [ ] End-to-end validation against a real promoted generation is a separate step (disposable Debian, then udder) and not claimed by this issue.

rjwittams:
## Summary

- add the Linux fleet installer for status, latest, exact-generation install, and rollback
- verify the immutable generation digest and complete candidate manifest before atomically switching the current release
- stop a running daemon before switches, detect PATH shadowing, and fail closed on Darwin
- document the promotion manifest consumed by the installer and run its offline contract suite in CI

## Validation

- `ci/fleet-candidates/test-workflow.sh`
- `scripts/test-fleet-install.sh`

Real promoted-generation validation remains the separate Debian/udder deployment step and is not claimed here.

Closes #1565

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,7 +33,9 @@ jobs:
           scripts/test-prune-target.sh
           scripts/test-cargo-sweep-mtime.sh
       - name: Test fleet candidate workflow contract
-        run: ci/fleet-candidates/test-workflow.sh
+        run: |
+          ci/fleet-candidates/test-workflow.sh
+          scripts/test-fleet-install.sh
       - name: Check Rust formatting
         run: cargo +nightly-2026-03-12 fmt --check
 
--- a/ci/fleet-candidates/README.md
+++ b/ci/fleet-candidates/README.md
@@ -29,3 +29,39 @@ The later trusted half must consume these exact bytes, sign the Darwin
 derivatives, run pristine-runtime proofs, publish the complete cohort, copy it
 offsite, and only then write an immutable completed generation. It must not
 rebuild either project.
+
+## Promoted generation consumer contract
+
+`scripts/fleet-install` consumes immutable versions of the Forgejo Generic
+Package `lab/flotilla-fleet`. Each version must contain the unchanged candidate
+archive and a `manifest.json` with this shape:
+
+```json
+{
+  "schema_version": 1,
+  "kind": "flotilla-fleet-generation",
+  "generation": "20260815T220000Z-f0123456789ab-cabcdef012345",
+  "peer_protocol_version": 20,
+  "platforms": {
+    "linux-x86_64-gnu2.36": {
+      "artifact": "fleet-candidate-linux-x86_64-gnu2.36.tar.gz",
+      "sha256": "<64 lowercase hexadecimal characters>"
+    }
+  }
+}
+```
+
+The protocol value must be copied from the candidate manifest produced above;
+the consumer rejects disagreement between the generation and candidate
+manifests. On Linux, bootstrap the reviewed script to
+`~/.local/bin/fleet-install`, store the package-read token at
+`~/.config/flotilla/fleet-reader-token` with mode `0600`, and run one of:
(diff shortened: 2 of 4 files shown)


rjwittams:
Fixed in `fabcc5c6`. The restrictive umask now runs only in a subshell around the authorization-header write, and the contract test asserts that a freshly-created launcher directory retains mode `0755`.

rjwittams:
Addressed both observations in `fec11ea1`. The contract now exercises a running daemon whose `flotilla daemon stop` succeeds and proves the switch proceeds. Mutating installs and rollbacks acquire a non-blocking `flock` on the install root, serializing both release publication and the `current`/`previous` pair; final release publication also uses `mv -T` so an existing target cannot become a nesting destination.

rjwittams:
Added the direct lock regression in `124cc5de`: the public contract holds `.install.lock`, verifies an exact-generation mutation is refused with the documented concurrent-mutation error, then releases the lock. The installed-release re-hash remains intentional integrity verification for this explicit operator command; automatic convergence is out of scope under #1259.

rjwittams:
Confirmed. This pass raises no new findings; no further code changes are needed.

PR #1568, opened 2026-08-15
fix: admit daemon shutdown across same-protocol builds

### LINKED ISSUE #1566: daemon stop is unusable for upgrades: handshake demands identical builds even at matching protocol
## Observed (2026-08-15, first post-#1548 fleet roll)

The graceful stop verb (#1546/#1548) exists precisely to stop an *older* daemon before switching binaries. But on its first real deployment use it refused itself:

```
wire generation mismatch: client built 63871b20… speaks proto 20; daemon built bfc40125… speaks proto 20 — rebuild/reinstall the CLI
```

Same `PROTOCOL_VERSION`, different build IDs → connection refused. Since staging replaces the CLI on disk before the restart, the only binary that could pass the identity check no longer exists; the deploy fell back to pkill — the exact thing #1546 was built to retire. (The lifecycle lock correctly prevented a second daemon from starting; no harm done.)

## The tension

The build-identity check guards against serde/wire drift that `PROTOCOL_VERSION` doesn't capture. That's a legitimate default for rich verbs, but it makes the protocol constant meaningless for cross-build communication, and shutdown *must* work across builds to be a deployment seam.

## What to build

Make the shutdown request reachable by any client whose `PROTOCOL_VERSION` matches (or define a minimal always-stable pre-handshake subset that shutdown belongs to). Rich verbs may keep the strict build-identity gate. Decide and document which stance the handshake takes; a `--force`-style escape that bypasses the version check entirely is not acceptable — the RPC must stay typed and acknowledged.

## Acceptance criteria

- [ ] `flotilla daemon stop` from a newer same-protocol build stops an older daemon with acknowledgement and zero exit.
- [ ] Cross-*protocol* stop behavior is explicitly defined (refuse with a clear message is fine).
- [ ] Behavior test covering new-client/old-daemon stop at the handshake seam.
- [ ] docs/development.md deploy flow verified against the actual behavior.

rjwittams:
Closes #1566.

## Summary

- add a shutdown-specific client handshake policy that accepts different build IDs only when the protocol version matches
- restrict same-protocol/different-build server sessions to the typed `Shutdown` RPC and its acknowledgement; rich RPCs retain the build-identity gate
- keep cross-protocol shutdown refused with the existing explicit protocol-version mismatch
- document the verified deployment sequence where the CLI is replaced before stopping the older daemon

The root cause was independent strict build-identity checks on both sides of the normal client handshake. The server closed before reading `Shutdown`, while the client rejected the server Hello before it could send the request.

## Verification

### DIFF
--- a/crates/flotilla-client/src/lib.rs
+++ b/crates/flotilla-client/src/lib.rs
@@ -53,14 +53,24 @@ impl SpawnLockGuard {
 /// Sends a `Hello` with `ConnectionRole::Client` and a fresh `session_id`,
 /// then waits for the server's Hello reply.
 async fn do_client_hello(session: &MessageSession) -> Result<Option<String>, String> {
-    do_client_hello_with_surface(session, None).await
+    do_client_hello_with_surface(session, None, WireGenerationPolicy::RequireMatch).await
 }
 
 async fn do_client_hello_with_declared_surface(session: &MessageSession, surface: SurfaceDeclaration) -> Result<Option<String>, String> {
-    do_client_hello_with_surface(session, Some(surface)).await
+    do_client_hello_with_surface(session, Some(surface), WireGenerationPolicy::RequireMatch).await
 }
 
-async fn do_client_hello_with_surface(session: &MessageSession, surface: Option<SurfaceDeclaration>) -> Result<Option<String>, String> {
+#[derive(Clone, Copy)]
+enum WireGenerationPolicy {
+    RequireMatch,
+    AllowMismatchForShutdown,
+}
+
+async fn do_client_hello_with_surface(
+    session: &MessageSession,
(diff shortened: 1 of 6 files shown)


rjwittams:
Addressed the non-blocking cleanup. `SocketDaemon::shutdown` had no remaining call sites after the CLI switched to the one-shot `shutdown_existing` deployment seam, so I removed the obsolete public method. The dedicated path is now the sole client shutdown API and retains its handshake and acknowledgement coverage.

rjwittams:
Verified. The current head preserves the protocol-version refusal, restricts cross-build sessions to one typed shutdown request, and has all refreshed CI checks passing. No further code changes are needed from this review.

Issue #1553, opened 2026-08-15
Central Darwin signing: fixed-function Comte signer emits attested derivatives


rjwittams:
## Parent

Split from #1547 (item 3 of 3). This supersedes the original local install-time signing proposal.

## Why

The credential-free Darwin build runner correctly emits an unsigned/ad-hoc candidate, but Kiwi needs identity-signed Flotilla binaries for the #689 syspolicyd mitigation. Distributing Apple signing credentials to every consumer Mac, or making `fleet-install` a signer, is not a durable trust boundary.

Signing belongs on Comte behind a separate fixed-function account/service. It must remain separate from the normal `forgejorunner` account and from repository-code execution. Consumer Macs verify and install centrally signed bytes and hold no signing credential.

## What to build

Create a fixed-function Comte signing path that:

1. accepts an exact Darwin candidate archive plus its already-verified input digest and generation metadata;
2. never checks out source, rebuilds, or executes repository/artifact scripts;
3. stages the payload safely and signs every Mach-O payload, including `flotilla`, `flotillad`, `cleat`, and bundled dynamic libraries, with the configured central identity and required entitlements;
4. verifies the signed payload and emits a derivative archive, input/output digests, signing-identity metadata, entitlements metadata, and an attestation;
5. allows Raclette's fixed-function promoter to verify that attestation before publishing a completed Darwin generation.

The first internal implementation may use the suitable centrally held Apple Development identity. Developer ID signing and notarisation can later replace or extend that identity without changing the `fleet-install` contract.

## Acceptance criteria

- [ ] Signing runs under a dedicated non-administrator Comte account/service with no Forgejo runner registration, source checkout, human ambient credential, or general repository execution path.
- [ ] The signer consumes only an exact digest-bound Darwin candidate and fails closed on any input mismatch.
- [ ] Every Mach-O payload is identity-signed with the intended entitlements and passes strict post-sign verification; any failure publishes no derivative.
- [ ] The derivative records the unsigned input digest, signed output digest, signing identity/team metadata, entitlements, and verification result.
- [ ] Raclette verifies the derivative and attestation before marking the Darwin member of a generation complete.
- [ ] `fleet-install` verifies and installs the central derivative without access to a signing credential.
- [ ] Mechanics are proved in a disposable macOS VM and the #689 syspolicyd mitigation is proved on Kiwi before Kiwi migrates.

## Sequencing

[...]

PR #1569, opened 2026-08-16
feat: install promoted fleet generations on macOS

rjwittams:
## Summary

- consume the promoter's real `generation.json` contract instead of the provisional manifest shape
- install immutable Linux candidates and centrally signed Darwin derivatives
- verify outer and inner manifests, sizes, digests, platform/source linkage, and executable payloads before selection
- require the trusted Apple team, recorded authority, strict designated requirement, and empty entitlements for every Darwin Mach-O payload
- keep installs operator-pulled with status, exact-generation, latest, and verified rollback commands
- document bootstrap and the promoted-generation consumer contract

## Root cause

The installer merged in #1567 expected `manifest.json` with kind `flotilla-fleet-generation`, while the promoter publishes `generation.json` with kind `internal-promoted-fleet-generation`. It also relied on GNU-only utilities, so it could neither consume the live package nor run on macOS.

## Validation

- `scripts/test-fleet-install.sh`
- `ci/fleet-candidates/test-workflow.sh`
- `bash -n scripts/fleet-install scripts/test-fleet-install.sh`
- `shellcheck --severity=warning scripts/fleet-install scripts/test-fleet-install.sh`
- `git diff --check`

The completed generation `20260815T233635Z-r84-f63871b20ec6f-c4e78c7c83873` was also installed into isolated roots on Kiwi and Udder. Flotilla, flotillad, and Cleat ran on both hosts, and Cleat resolved its bundled Ghostty dynamic library. The proof did not replace either host's active binaries or move its live `current` link.

Relates to #1553.

### DIFF
--- a/ci/fleet-candidates/README.md
+++ b/ci/fleet-candidates/README.md
@@ -45,29 +45,60 @@ central-signing linkage.
   "schema_version": 1,
   "kind": "internal-promoted-fleet-generation",
   "generation": "20260815T233635Z-r84-f63871b20ec6f-c4e78c7c83873",
+  "source_generation": "20260815T223755Z-r84-f63871b20ec6f-c4e78c7c83873",
   "sources": {
     "flotilla": "63871b20ec6f5a9e41308b5d56a90a4fefd41f8e",
     "cleat": "4e78c7c83873916dcb2342a51001bdfed3d63eda"
   },
   "peer_protocol_version": 20,
+  "central_signing": {
+    "derivative_package": "lab-signing/flotilla-fleet-darwin-signed",
+    "derivative_version": "20260815T223755Z-r84-f63871b20ec6f-c4e78c7c83873",
+    "attestation": "darwin-signing-attestation.json",
+    "attestation_sha256": "<64 lowercase hexadecimal characters>",
+    "cms": "darwin-signing-attestation.cms",
+    "cms_sha256": "<64 lowercase hexadecimal characters>",
+    "certificate": "darwin-signing-certificate.pem",
+    "certificate_sha256": "dffbf762cdba4dab884d89df2350a50f324daada53d86d55068c311fbbf59c4e",
+    "signing": {
+      "identity": "Apple Development: Robert Wittams (DYYMCPD885)",
+      "team_id": "973L4GV58R",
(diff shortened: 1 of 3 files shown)


rjwittams:
Addressed review comment 5304953657 in `47473564`.

1. Replaced the crash-persistent lock directory with an atomically created, uniquely owned symlink lock. A contender refuses a live PID, reclaims a dead owner only after comparing the exact unique owner token, and cleanup removes only its own lock. Tests cover both live-lock refusal and stale-lock recovery.
2. `latest` now skips only HTTP 404 or a manifest explicitly incomplete for the host platform, and reports each skip. Transport/auth/other HTTP errors and malformed advertised manifests stop the operation instead of silently selecting an older generation. Regression tests cover incomplete fallback, exact incomplete requests, and a newer-generation HTTP 503.
3. Confirmed that omitting `--location` is deliberate: package credentials are never forwarded through redirects, and unexpected redirect responses fail closed. The live Forgejo endpoints used by Kiwi and Udder do not redirect.
4. The automated `codesign` DSL limitation remains inherent to non-Darwin CI. The exact requirement was exercised successfully with real `/usr/bin/codesign` against every signed payload on Kiwi; the PR body records that live proof.

rjwittams:
Addressed follow-up review comment 5304991026 in `355964bd`.

The README example now includes the exact source-generation linkage, complete `central_signing` record, per-platform source artifact linkage, and both identical signing objects required by the installer. It uses the proven live generation's concrete source IDs, artifact sizes/digests, identity, team, certificate digest, and entitlements digest; the two unavailable long digests remain explicitly labeled metavariables. A note states the required equality relationships.

The JSON example parses successfully and the fleet candidate workflow contract remains green. The earlier operability findings from comment 5304953657 remain addressed by `47473564`.

rjwittams:
Handled in `47473564`: crash-stale locks now self-heal safely, fatal manifest fetch/validation failures no longer cause silent downgrade, redirect refusal is documented as deliberate credential hardening, and the real Kiwi `codesign` proof remains recorded.

rjwittams:
Handled in `355964bd`: the documented Darwin generation now includes the complete source and central-signing linkage required by the consumer, including the equality constraints between signing records.

rjwittams:
Verified. This pass reports no new blocking or actionable issue; no further code change is needed.

